PR #3, opened 2016-12-16
Stop forcing Odamex to 35FPS on very few conditions.

Ch0wW:


AlexMax:
Thanks for the contribution!

PR #4, opened 2016-12-17
Fixes flying spectators being able to play locally the grunt1 sound. 

Ch0wW:
You can reproduce it easily on moo2d while flying on some sectors. (like the window near the Rocket Launcher)


Ch0wW:
I decide to close this pull request, because this shouldn't be the proper way to fix this. Actually, it's because of a slight oversight related to hitting a sector when flying.

PR #5, opened 2016-12-27
SPECTATOR: Fixes bouncing as a flying spectator. (+1 fix)

Ch0wW:
This fix also prevents grunt1 sound from playing.


AlexMax:
I was a bit worried about checking `slidemo->player->spectator` without a separate check for `NULL` in `slidemo->player`, but I checked with @Ch0wW and he was unable to reach this function outside the context of a player, so I'm cool with merging it.

Thank you for your contribution.

PR #7, opened 2016-12-29
Safeprinting the version of Odamex in the console

Ch0wW:
The reason is a few commits can strangely extend 7 characters, making the console crash because of an unexpected overflow.

Also, I limited the git describing to 6 characters, to better respect regular git standards, when it comes to show them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/modules)
 
 # git describe
 include(GetGitRevisionDescription)
-git_describe(GIT_DESCRIBE --tags --always)
+git_describe(GIT_DESCRIBE --tags --always --abbrev=6)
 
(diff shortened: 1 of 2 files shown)


AlexMax:
Nice catch.  However, I took a look at the manpage of `git describe` and found something worrying:

> --abbrev=<n>
> Instead of using the default 7 hexadecimal digits as the abbreviated object name, use <n> digits, **or as many digits as needed to form a unique object name**.

Emphasis mine.  It seems that we cannot rely on `git describe` to return us an **n** length object name.  In the case of ambiguous objects, `git describe` will add characters until it returns a unique object.

- Could you instead perhaps rewrite this code so it will accept arbitrary length object names?
- Could you remove the `--abbrev` parameter?  If we're not actually gaining anything from using a shorter object name, it'd be nice to use whatever the git standard is.

AlexMax:
Merged, with some additional changes.

Thank you for your contribution.

PR #12, opened 2017-01-18
"Drasticaly" lower FPS on new SDL2 builds

capodecima:
Tested some builds from october (still SDL1). And some new SDL2 builds. I got way lower FPS in SDL2. I tested in lower resolutions. On SDL1 i got 450 fps now around 300 in SDL2. Just wanted to you guys know. Probably needs some more optimalizition. I was testing 32/64 bit builds, its same...


AlexMax:
Pull requests are not issues.

Please do not open another one unless you have code to contribute.

PR #28, opened 2019-01-07
Add keys to the Warmup message

Ch0wW:
This PR creates a new function in C_BIND.cpp (C_GetKeyStringsFromCommand) that tells you what buttons you use from a bind.

That created, I used it on the "Warmup: you are not ready" so that it says "Warmup: press <KEYBINDS> to ready up".


Ch0wW:
Here's an example for the "Ready" message, if there's a key found :
image

If no key was found : 
image

PR #31, opened 2019-01-15
Language fixes

Ch0wW:
- Forces ENGLISH as the language of the server.
- Fixes `language` CVAR being seen as a boolean, while technically accepting up to 256 languages. (yet it doesn't though)


Ch0wW:
Nope, you're right. We could safely remove that.

PR #32, opened 2019-01-15
Adding a demofilter

Ch0wW:
Requested by the community, to be able not to record everything from every gamemode.

I also added these options under the "Network options" submenu.


Ch0wW:
Also, here's a screenshot of how it looks ingame.
image

Ch0wW:
The problem, is that I first planned a byte system, but it'd be absolutely messy for the player if modifying from the .cfg . Individual CVARs were far easier to manage in case of a problem though.

client/src/m_options.cpp
```diff
@@ -216,6 +226,11 @@ value_t OnOffAuto[3] = {
 	{ 2.0, "Auto" }
 };
 
+value_t DemoRestrictions[2] = {
```
hobomaster22: Whys is 0 allow and 1 restrict? This is OK as far as setting in the application. This is confusing in odamex.cfg because having cl_autorecord_filter_duel set to 1 would imply that it would record but it's the opposite.

To me it seems we should flip the DemoRestrictions values and default the cvar to 1.

PR #33, opened 2019-01-16
HUD additions/fixes

Ch0wW:
This PR contains several fixes about the HUD, and common additions that the community requested, such as:
- Being able not to display the scoreboard on death ;
- Being able to hide the Demo Bar/timer during a netdemo replay ;
- Being able to choose the FullHUD Style in the options ;
- Being able to simplify the border and to disable flashes when picking up a flag in CTF ;
- Fixing a 8bpp palette refreshing error after ending a map (screen was slightly red/yellow in the intermission)
- Fixing a case of "Warmup: you are not ready" in Singleplayer
- Fixing cheats not being reseted on some special occasions (this will be more in effect with a future rewrite of the cheats support)


Ch0wW:
Options #1 :
image
 
Options #2 : 
image

Simplier hud border in effect (flashes disabled) : 
image

client/src/cl_ctf.cpp
```diff
@@ -378,16 +383,23 @@ void CTF_DrawHud (void)
...
+			else if (tintglow > 45 && tintglow < 60)
+				tintglowtype = tintglow - 45;
+			else if (tintglow >= 60 && tintglow < 75)
+				tintglowtype = 75 - tintglow;
+			else
+				tintglowtype = 0;
+		}
 		else
+		{
 			tintglowtype = 0;
```
hobomaster22: I think it's better to just initialize tintglowtype = 0 where it's declared at the top and then remove this else

PR #36, opened 2019-01-25
Several various minor to major fixes 

Ch0wW:
> Removes the double localized "You need a XXX key" from the server.

This is a fix that disables the message spread from the server to the client. Besides, I don't get why that part of code was created. So, I just assume it was for testing purposes, but is troublesome otherwise, if the PWAD uses custom messages.

For some reason, that said message is also spread from the server's locale (check PR #31 to also remove that). So for instance, it'll be displayed in French even if the user's language is in english.

### DIFF
--- a/common/p_pspr.cpp
+++ b/common/p_pspr.cpp
@@ -199,7 +199,7 @@ void P_SetPsprite(player_t* player, int position, statenum_t stnum)
 		// Modified handling.
 		if (psp->state->action)
 		{
-			if (!player->spectator)
+			if (!player->spectator && player->mo != NULL)
 				psp->state->action(player->mo);
 
 			if (!psp->state)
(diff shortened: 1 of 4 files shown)


Ch0wW:
About commit 2577123, it doesn't fix another major problem when using addmap on Linux.

When using it with a -waddir parameter, it seems it looks the **whole folder as the wadname** instead of simply the WAD, saying

>could not find resource file: /SERVER/DOOM/WADS/FREEDOOM1.WAD

When the correct name should be FREEDOOM1.WAD 
>[01:53:38] *1. /SERVER/DOOM/WADS/FREEDOOM1.WAD E1M1
[01:53:38] +2. /SERVER/DOOM/WADS/FREEDOOM1.WAD E1M2
[01:53:38]  3. /SERVER/DOOM/WADS/FREEDOOM1.WAD E1M3

The result looks like that, and E1M1 won't stop looping over and over after an exit.

seanmleonard:
Great set of fixes @Ch0wW !

PR #40, opened 2019-02-17
Fix for bug 1067

hobomaster22:
Clear the moving sectors when going from in game back to spectator so that moving sectors do not appear stuck.


seanmleonard:
Thanks Hobo!

PR #62, opened 2019-07-13
Instant Switching weapons (Ticket #1119)

Ch0wW:
This commit allows weapons to be instantly switched, with the help of sv_instantswitch .

BUG from ► https://odamex.net/bugs/show_bug.cgi?id=1119


seanmleonard (changes requested):
You'll need to ensure this is disabled when recording or playing a vanilla demo

PR #67, opened 2019-08-26
Fixing the Wii compilations

Ch0wW:
After a long time, I've fixed the Wii builds. It works now on Windows MSYS2, but it should also work on other platforms if you follow the guidelines included in Wii/README.md


Ch0wW:
Will close it, since I've rewrote it using CMake.

PR #68, opened 2019-09-08
Use DeuTex to build the wad

vilhelmgray:


Ch0wW:
It looks solid to me !

PR #74, opened 2019-12-10
Added Freedoom updated to 0.12.1 + Fix for SP losing the inventory after each map.

Ch0wW:
Also fixed a messy condition in G_CleanupDemo().


Ch0wW:
I've updated the PR so that it now includes the following: 
- Simple check for depreciated IWADs (especially Freedoom)
- Fix an error which made the player lose inventory after each map.

PR #77, opened 2019-12-11
Synchronize secrets online across clients

Ch0wW:
Odamex doesn't make clients aware if a secret has been discovered on the map or not.

This PR fixes that.


seanmleonard:
Note: this appears to change C/S message protocol and will need to be deferred until 0.9.0

AlexMax:
LGTM

PR #94, opened 2019-12-31
Add Discord-RPC support

Ch0wW:
Odamex could beneficiate from Discord's RPC feature, being able to get some gaming infos (is he playing on DM ? SP ?). I added it on CMake, and made it as an optional, static library.

Also, the feature can be disabled on the client.


Ch0wW:
I'm self-closing this PR. The idea was nice but the way I incorporated it is absolutely biased.

Will probably reopen it once it'll be in better quality.

PR #96, opened 2020-01-04
Don't draw the scoreboard when chatting.

Ch0wW:
It'll make the game think the button is still pressed while it's definitely not the case.

This PR fixes that so if you press chat while drawing the scoreboard, it disables it for you.


Ch0wW:
This doesn't fix entirely the problem ; TAB can still be pressed and be bound to a button in Chatmode.

Ch0wW:
Issue found.

Since "KEY_TAB" is considered as KEY #9, it is used in C_DoKey().

HOWEVER, since HU_Responder goes BEFORE C_DoKey, and TAB didn't have any handler, it made an action.

PR #98, opened 2020-01-13
Change source path of odamex.wad for macOS client

Terr:
When compiling for macOS, CMake was looking for `odamex.wad` in the project root directory but it looks like it should be `wad/odamex.wad` instead.

I'm not sure how it is for other OS'es. I think I'm seeing another reference to a root-located `odamex.wad` in the Windows installer wizard script, but I cannot test it.


seanmleonard:
It looks like both you and JamesDunne added PRs for this today. I merged his PR since it was already based against the "development" branch, rather than "master".

PR #99, opened 2020-01-13
Boom fixes and Coop E*M8's intermission restoration

Ch0wW:
This commit does the following :

- Fix the scrolling sectors so that it goes at normal speed
- Added back a few ZDoom sectors and custom actions
- Fix ExM8 intermission screen, so that it doesn't go immediately to the next episode without breaking classic demo compatibility.


common/p_map.cpp
```diff
@@ -3116,6 +3116,9 @@ msecnode_t *P_AddSecnode (sector_t *s, AActor *thing, msecnode_t *nextnode)
 
 	msecnode_t *node;
 
+	if (s == 0)
```
seanmleonard: Can you change to "s == NULL" to more clearly convey this is a check for a null pointer?

common/p_setup.cpp
```diff
@@ -1819,7 +1847,13 @@ CVAR_FUNC_IMPL(sv_timelimit)
 
 CVAR_FUNC_IMPL(sv_intermissionlimit)
 {
-	level.inttimeleft = (var < 1 ? DEFINTSECS : var);
+	if (sv_gametype == GM_COOP && var < 10) {
```
seanmleonard: Any particular reason for enforcing a 10 second minimum for co-op?

seanmleonard: Nevermind, I see this code was part of the previous implementation:
`int intlimit = (sv_intermissionlimit < 1 || sv_gametype == GM_COOP ? DEFINTSECS : sv_intermissionlimit);`

common/p_map.cpp
Ch0wW: I just made the change with my latest commit.

PR #100, opened 2020-01-14
Fix various crashes in client

JamesDunne:


client/src/r_things.cpp
```diff
@@ -670,7 +670,13 @@ void R_ProjectSprite(AActor *thing, int fakeside)
 	unsigned int		rot;
 	bool 				flip;
 
-	if (!thing || !thing->subsector || !thing->subsector->sector)
```
seanmleonard: C/C++'s short-circuit evaluation means that the removed line equivalent to the added lines so we're not really adding more protection against accessing null pointers here

JamesDunne: Yup they should be logically equivalent but there are two reasons why I broke it apart:
1. You get separate line numbers for debugging in the crash report to better tell which null pointer condition actually triggered the crash. When it's all on one line, you can only guess.
2. It avoids the potential for compiler bugs and human error in interpreting long chains of boolean logic with short-circuit evaluation.

seanmleonard:
I like the idea of having a separate line for crash reporting purposes!

JamesDunne:
What's weird is ... and this is totally anecdotal ... but after I made that change and released new builds, I did not see further crashes around there. I can't make that claim with 100% certainty though; sometimes you don't get the crash dump back and I haven't looked at all the files uploaded either.

PR #103, opened 2020-01-14
Server: performance fixes for O(n^2) loops

JamesDunne:
SV_WriteCommands() was calling SV_UpdateHiddenMobj() from a foreach player loop; SV_UpdateHiddenMobj() has its own foreach player loop; moved the SV_UpdateHiddenMobj() function out of the foreach player loop in SV_WriteCommands() which drastically reduces server CPU utilization according to playtesting.

Replaced I_Yield() call in D_RunTics() for SERVER_APP with explicit I_Sleep() and calculated sleep duration. This reduces unnecessary CPU utilization because I_Yield() simply loops and sleeps for 1ms each loop iteration.

optimized poor performance of SV_ClientFullUpdate() updating switch/button state: introduced new P_UpdateButtons() function to iterate efficiently through all switch objects.

More efficient std::queue clearing by swapping with empty instance instead of while(!empty){pop()} loop.

### DIFF
--- a/common/d_main.cpp
+++ b/common/d_main.cpp
@@ -1130,10 +1130,11 @@ void D_RunTics(void (*sim_func)(), void(*display_func)())
 	dtime_t simulation_wake_time = simulation_scheduler->getNextTime();
 	dtime_t display_wake_time = display_scheduler->getNextTime();
 
-	do
-	{
-		I_Yield();
-	} while (I_GetTime() < MIN(simulation_wake_time, display_wake_time));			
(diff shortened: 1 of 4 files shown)


JamesDunne:
The D_RunTics() code change also affects client performance in a positive way: +100fps on stctfmp.wad map01.

JamesDunne:
That was the very first change I made after decino's first stress test. Looked like it was accidentally wrapped up in the foreach-players loop a long time ago.

PR #104, opened 2020-01-14
Vibranium doors

JamesDunne:
doors: added sv_activation_rate default to 35 tics to rate limit linedef activations for untagged doors and lifts.


JamesDunne:
I just realized this may break compatibility on the client side so I need to make sure it's only enableable for server side.

PR #106, opened 2020-01-14
Server coop-completionist mode

JamesDunne:
defined new sv_coop_completionist cvar to prevent exit unless all monsters killed and all secrets found.

count only original map spawned killable monsters as completionist kills.
move sv_coop_completionist cvar to common/c_cvarlist.cpp but left unsynced between server and client. switch Printf to C_MidPrint so player sees unfinished states for completionist.
singularize and pluralize parts of message for monster(s) and secret(s).
fix bug where crossing special exit linedefs did not reattempt to exit after first failure; same with shootable linedefs or pushable linedefs.
print server message as well as send client message when exit attempt fails.
send message to all players when each secret is revealed.
allow clients to vote on killable monster count and findable secret count. future enhancement should load these counts per wad per map from a json file ideally loaded from odamex.net.
callvotes are named "secrets" and "kills", and enable callvotes when sv_coop_completionist is enabled.

### DIFF
--- a/common/p_setup.cpp
+++ b/common/p_setup.cpp
@@ -48,6 +48,11 @@
 
 #include "p_setup.h"
 
+EXTERN_CVAR(sv_coop_completionist_kills)
+EXTERN_CVAR(sv_coop_completionist_secrets)
+EXTERN_CVAR(sv_coop_completionist_killed)
+EXTERN_CVAR(sv_coop_completionist_found)
+
 void SV_PreservePlayer(player_t &player);
 void P_SpawnMapThing (mapthing2_t *mthing, int position);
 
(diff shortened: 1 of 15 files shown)


common/p_setup.cpp
```diff
@@ -1659,6 +1664,13 @@ void P_SetupLevel (char *lumpname, int position)
 		wminfo.maxfrags = 0;
 	wminfo.partime = 180;
 
+	// reset completionist counts to -1 which is translated to level.total_monsters and level.total_secrets in p_lnspec.cpp:
+	sv_coop_completionist_kills.ForceSet(-1.0f);
+	sv_coop_completionist_secrets.ForceSet(-1.0f);
```
JamesDunne: These `-1` sentinel values are convenient for the server to know to look up the value for the current map but for rcon users reading a `-1` values tells them nothing useful and is confusing. It'd be better to reset the cvar counts on map load to their actual values and avoid `-1`.

PR #107, opened 2020-01-14
Client: spyname command

JamesDunne:
Added `spyname` command for convenience of spying players by name instead of id.

try to actively spy on named player when they reconnect, e.g. level changes.

spyname with no args unfollows player and will disable spy mode

### DIFF
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -1078,27 +1078,6 @@ END_COMMAND (spyprev)
 
 BEGIN_COMMAND (spy)
 {
-	byte id = consoleplayer_id;
-
-	if (argc > 1)
-		id = atoi(argv[1]);
-
-	if (id == 0)
-	{
-		Printf(PRINT_HIGH, "Expecting player ID.  Try 'players' to list all of the player IDs.\n");
-		return;
-	}
-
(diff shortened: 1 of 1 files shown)


seanmleonard:
Can we combine the `spy` and `spyname` commands to a single command `spy`?

Some pseudocode:
if not is_integer(argv[1])
    for i = 0; i < numclients; i++
        if clients[i].name == argv[1]
            displayplayer_id = i
else
    displayplayer_id = int(argv[i])

JamesDunne:
I'd probably go further and just replace `spy <number>` with `spy <name>`. Using player numbers for `spy` doesn't make a ton of sense in the first place. The player numbers always change depending on the random order of players (re)joining the game and they also change across maps because a map change disconnects all clients and instructs them to reconnect so they're pretty unstable as an identifier to use in a long-running game.

seanmleonard:
Can you check to see that the server prevents more than one player from using the same name? If so, I agree, let's have it entirely name-based.

JamesDunne:
Sure I'll double check that. That was my assumption too.

JamesDunne:
Oh yeah, it just adds a numeric suffix to your name "Player2" "Player3" if your name is not already unique.

PR #108, opened 2020-01-14
Socket buffer size cvars

JamesDunne:
netcode: make cvars for send/receive buffer size: net_rcvbuf, net_sndbuf both default to 128KiB.


seanmleonard:
Thanks @JamesDunne!

PR #109, opened 2020-01-14
Fix Icon of Sin save game crash

Terr:
Duplicates how vanilla Doom II did it: just remove the spawning cube when it has an invalid target (which it has after loading)

Fixes https://odamex.net/bugs/show_bug.cgi?id=1302


Ch0wW:
Looks like a solid commit, since it adapts well to the Doom 2's saved data.

PR #114, opened 2020-01-14
Server: performance fix for ambush monsters

JamesDunne:
server: dont send monster updates for monsters that were awoken by sound but have MF_AMBUSH flag.

### DIFF
--- a/common/p_enemy.cpp
+++ b/common/p_enemy.cpp
@@ -736,11 +736,12 @@ void A_Look (AActor *actor)
 		{
 			if (P_CheckSight(actor, actor->target))
 				goto seeyou;
+			else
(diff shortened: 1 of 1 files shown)


seanmleonard (commented):
We will need to verify that this does not break LMP demo compatibility. If it does break compatibility, we'll need to add a clause like:

`if (!demoplaying && !demorecording)`

JamesDunne:
Right! I only did a cursory test with `demo1` in doom2. I'll do some more in-depth testing. If you have a pointer to some LMPs to test that'd be greatly appreciated. doom2 map02 is what inspired this fix in the first place, so LMP demos on that vanilla map would be best to start with.

PR #115, opened 2020-01-14
Server: fix double-callback on cvar Set

JamesDunne:
cvars: do not call Set and Callback twice serverside when unlatching cvars.


seanmleonard (commented):
If I'm reading the code correctly and the server admin changes the CVAR after a client has connected, the server will never transmit that change to the clients (via D_SendServerInfoChange).

JamesDunne:
I'm unable to get to my laptop this weekend but I do remember debugging callbacks for latched cvars and they were being doubly called on map start when unlatching. There's another code path that manually calls D_SendServerInfoChange on updating the cvars which made this one redundant.

PR #117, opened 2020-01-14
Server: prevent DoS during WAD downloading

JamesDunne:
server: short-cut clc_wantwad to resume download without expensive MD5 checks.

Without this short-cut, the server will constantly keep recomputing the MD5 hash of the requested WAD whenever the client misses a packet. This causes the CPU utilization to climb to 100% easily when packet loss is high.

### DIFF
--- a/common/d_player.h
+++ b/common/d_player.h
@@ -298,11 +298,12 @@ public:
 		{
 		public:
 			std::string name;
+			std::string md5;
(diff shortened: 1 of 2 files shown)


seanmleonard:
Looks good. I've got some changes in the `resource_files` branch that caches the MD5 that I'll adapt to work with your changes.

PR #121, opened 2020-01-15
Survival game mode

JamesDunne:
survival: sv_survival enables/disables survival mode. all players get 1 life on map start and once all players are dead the map restarts.

disallow rejoining after spectating.
disallow joining after join_timer expires (15 seconds).
override sv_keepkeys in survival mode
GM_COOP is normal survival mode; GM_DM is last man standing mode; GM_TEAMDM and GM_CTF are TODO.
proper join timer reset for LMS mode.
disallow map exits when not in COOP.
add cvar controls for sv_maxlives and sv_survival_join_timer.
added server command sv_survival_reset_timer to be able to reset join timer at any point if one wants players to join mid-round.
broadcast messages about remaining player counts.

new cvars:
sv_survival             - enables or disables survival mode
sv_survival_join_timer  - allowed time for players to join at start of level in seconds (default 15)
sv_maxlives             - number of initial lives given to each player when joining (default 1)

commands:
sv_survival_reset_timer - reset the join timer

### DIFF
--- a/common/c_cvarlist.cpp
+++ b/common/c_cvarlist.cpp
@@ -35,6 +35,15 @@ CVAR_RANGE(			sv_gametype, "0", "Sets the game mode, values are:\n" \
 					CVARTYPE_BYTE, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_LATCH | CVAR_NOENABLEDISABLE,
 					0.0f, 3.0f)
 
+CVAR(				sv_survival, "0", "Survival mode",
+					CVARTYPE_BOOL, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_LATCH)
+
+CVAR(				sv_survival_join_timer, "15", "Survival mode join timer",
+					CVARTYPE_FLOAT, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_NOENABLEDISABLE)
+
+CVAR(				sv_maxlives, "1", "Survival mode number of lives",
+					CVARTYPE_INT, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_NOENABLEDISABLE)
+
 CVAR(				sv_friendlyfire, "1", "When set, players can injure others on the same team, " \
 					"it is ignored in deathmatch",
 					CVARTYPE_BOOL, CVAR_SERVERARCHIVE | CVAR_SERVERINFO)
@@ -178,7 +187,7 @@ CVAR(				sv_coopunassignedvoodoodolls, "1", "",
 					
 CVAR(				sv_coopunassignedvoodoodollsfornplayers, "255", "", 
 					CVARTYPE_WORD, CVAR_SERVERINFO | CVAR_LATCH)
-	
+
 
 // Compatibility options
 // ---------------------------------
--- a/client/src/cl_stubs.cpp
+++ b/client/src/cl_stubs.cpp
@@ -62,5 +62,8 @@ void CTF_SpawnFlag(flag_t f) {}
 bool SV_AwarenessUpdate(player_t &pl, AActor* mo) { return true; }
 void SV_SendPackets(void) {}
 
+void SV_SurvivalRestartLevel(void) {}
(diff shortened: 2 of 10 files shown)


common/c_cvarlist.cpp
```diff
@@ -35,6 +35,15 @@ CVAR_RANGE(			sv_gametype, "0", "Sets the game mode, values are:\n" \
 					CVARTYPE_BYTE, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_LATCH | CVAR_NOENABLEDISABLE,
 					0.0f, 3.0f)
 
+CVAR(				sv_survival, "0", "Survival mode",
+					CVARTYPE_BOOL, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_LATCH)
+
+CVAR(				sv_survival_join_timer, "15", "Survival mode join timer",
+					CVARTYPE_FLOAT, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_NOENABLEDISABLE)
+
+CVAR(				sv_maxlives, "1", "Survival mode number of lives",
```
seanmleonard, line 44: Can we change the CVAR name to `sv_survival_maxlives` to keep it in the same "namespace" as the other survival CVARs?

seanmleonard:
I think this looks good to go into 0.9.0. 

I'm leaning towards not storing game-mode specific data in the `player_t` class and instead having game-mode classes that store per-client or per-player data. It's probably not that important for your survival PR since you're just adding `survival_lives`, but I may enlist some help from you with designing game-mode abstractions in the future.

common/c_cvarlist.cpp
Ch0wW: I'd disagree, since LMS/TLMS will also be included shortly into that.

Ch0wW:
Serverside, the code seems alright. Clientside, we need to make them aware of the lives available and more elements behind it ; which would require some protocol changes.

AlexMax:
What if the struct member was renamed to just `lives` instead?  That way, it doesn't imply a specific game mode, and could be repurposed for something like LMS/TLMS in the future.

seanmleonard:
`lives` works for me!

tm512:
Back in 2014 and 2017, I worked on implementing survival and (T)LMS myself. It seems like this patch is missing several features that I had implemented in my own. The goal with mine was to create a similar user experience to zandronum's survival mode.

My survival patch is built on top of my join queue patch. This provides a queue (available in all game modes) that spectators can join when a game is full or when a round is in progress. Open slots in the game are filled in the order that people joined the queue, helping ensure that people who have been waiting to play won't have their spot in the game sniped from someone who hasn't been waiting as long.

My patch also specially handles players who are forced to spectate after running out of lives, by flagging them as a "deadspectator". deadspectators are given priority over the queue when the map resets or when moving on to a new map. It's also possible for a deadspectator to use the "spectate" console command to clear the deadspectator flag, so the game won't automatically respawn them when a new round starts.

I also implemented a lives display for the HUD. My initial work can be found in [my fork](https://github.com/tm512/odamex/tree/survival), but it looks like Sleep has done some work bringing the patch up to date, which can be found [here](https://github.com/so-sleepy/odamex-tm512/tree/survival). I haven't tested the update myself though.

If JamesDunne's patch is the route you guys want to take, perhaps implementing these additional features is something worth considering.

JamesDunne:
Just for clarity, I implemented this under the constraints of not breaking network protocol so a lot of features were not able to be implemented. It sounds like tm512 has a more comprehensive solution so it may be a better starting point.

so-sleepy:
> I also implemented a lives display for the HUD. My initial work can be found in [my fork](https://github.com/tm512/odamex/tree/survival), but it looks like Sleep has done some work bringing the patch up to date, which can be found [here](https://github.com/so-sleepy/odamex-tm512/tree/survival). I haven't tested the update myself though.

This one is integrated into the development branch https://github.com/so-sleepy/odamex/tree/development

AlexMax:
A more complete survival mode, and much more besides, is merged for 0.9.  Appreciate you taking the time regardless.

PR #131, opened 2020-04-04
Compiler doesn't issue any warnings

so-sleepy:
All warning messages have been removed.

### DIFF
--- a/client/src/c_bind.cpp
+++ b/client/src/c_bind.cpp
@@ -737,7 +737,7 @@ std::string C_GetKeyStringsFromCommand(const char *cmd, bool bTwoEntries)
 	C_GetKeysForCommand(cmd, &first, &second);
 
 	if (!first && !second)
-		return "< ??? >";
+		return "<??\?>";
 
 	if (bTwoEntries)
 		return C_NameKeys(first, second);
@@ -748,7 +748,7 @@ std::string C_GetKeyStringsFromCommand(const char *cmd, bool bTwoEntries)
 		else
 			return KeyName(first);
 	}
-	return "< ??? >";
+	return "<??\?>";
 }
 
 
--- a/client/src/c_bind.cpp
+++ b/client/src/c_bind.cpp
@@ -36,7 +36,6 @@
 #include "hu_stuff.h"
 #include "cl_demo.h"
 #include "d_player.h"
-//#include "stringutils.h"
 
 extern NetDemo netdemo;
 
--- a/client/src/c_bind.cpp
+++ b/client/src/c_bind.cpp
@@ -36,6 +36,7 @@
 #include "hu_stuff.h"
 #include "cl_demo.h"
 #include "d_player.h"
+//#include "stringutils.h"
 
 extern NetDemo netdemo;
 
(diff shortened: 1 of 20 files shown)


so-sleepy:
GCC 9.2.1 issues no warnings.

client/src/c_bind.cpp
```diff
@@ -729,15 +730,15 @@ C_GetKeyStringsFromCommand
...
+std::string C_GetKeyStringsFromCommand(const char *cmd, bool bTwoEntries)
 {
 	int first = -1;
 	int second = -1;
 
 	C_GetKeysForCommand(cmd, &first, &second);
 
 	if (!first && !second)
-		return "<???>";
+		return "< ??? >";
```
Ch0wW, line 740: What happens if you use <\?\?\?> instead, to avoid sending spaces to the HUD ? (unbind your `ready` key, and join a warmup server to see the string in action)

client/src/hu_elements.cpp
```diff
@@ -451,7 +451,7 @@ std::string NetdemoElapsed() {
 	timeelapsed -= minutes * 60;
 	int seconds = timeelapsed;
 
-	char str[9];
+	char str[21];
```
Ch0wW, line 454: Any reason for 21 instead of 25 like the other lines above ?

client/src/c_bind.cpp
so-sleepy: The compiler warns about ignored trigraphs, specifically ??>. I found no way to keep it as is so I added spaces on both sides. https://stackoverflow.com/questions/1234582/purpose-of-trigraph-sequences-in-c

client/src/hu_elements.cpp
so-sleepy: The compiler warns about the possibility of lacking space, there is a risk for segfault in the sprintf() functions right below. 9 bytes are sufficient, but gcc insists it isn't. 20 bytes still results in warning, 21 doesn't.

client/src/c_bind.cpp
so-sleepy: Actually I could just use \?.

Ch0wW: Long after the original commit, I modified it with \?\?\? to avoid that warning without pushing it, but if ??\? also works, then it's good to me.

so-sleepy: Commit added addressing this.

client/src/hu_elements.cpp
so-sleepy: For any other occurrences, I followed the same principle, that is, to raise the number of bytes for each array until gcc stopped issuing warnings, all of them were warnings around the usage of sprintf() in different places.

client/src/st_stuff.cpp
```diff
@@ -701,7 +701,7 @@ bool ST_Responder (event_t *ev)
 			if (gamemode == retail_chex)
 				sprintf(buf,"1%c",buf[1]);
 
-            sprintf (buf + 3, "map %s\n", buf);
+            sprintf (buf + 3, "map %.7s\n", buf);
```
seanmleonard, line 704: I think `%.7s` is a bit misleading to the reader as they may initially think it refers to how many characters will be written to the string. Since it should always be writing 2 characters, would `%.2s` work (eg, properly terminate the string)?

so-sleepy: That does work, and commit sent.

seanmleonard:
Is this ready to be merged?

so-sleepy:
In my opinion, yes, I haven't noticed any issue.

PR #133, opened 2020-04-05
Bug #1236, plus additional stubs

AlexMax:
This is an attempt to fix the custom pinkie crash in Valiant, but the scope got a little wider past that.

Essentially, the custom pinkie uses a state index that Odamex simply does not have.  In theory, this change could be fixed by adding a bunch of blank states, like what happened in Woof and Graf's fork of prBoom+.  However, I decided to go a little bit above and beyond and stub out the missing MBF codepointers and mobj types as well.  This ensures that the final boss of Valiant, which uses the helper dog mobj, works as well.

Most of the other MBF code pointers are stubbed out, because the important ones that ZDoom gives first class support for already exist.  I also did not add any default graphics or sounds for the mobjs to the WAD either, since that might give the impression that these features are fully supported and not just a shim (although if such resources were desired, Woof has some under a usable license).  I also imagine that the Beta BFG would be a gigantic drain of network resources anyway.

Note that in order to get this stuff working, I moved the position of MBF mobj definitions and state frames to after Vanilla but before ZDoom/Odamex.  The reason I did this is because I wanted the mobj definition array and state table to line up with MBF's state table, just in case there were ever any compatibility subtleties that could arise from the different orders.  However, I suspect doing this also implies a protocol breakage, since mobj ids and frame indexes wouldn't be compatible (but that's going to be the case no matter where I put the mobjs and state frames).

This pull request also comes with some extra code that prevents a crash when trying to go to a state index outside the state table.  To implement this, I also added an extra `COUNTOF` macro that counts the size of a static C array, which should be useful elsewhere.

### DIFF
--- a/common/d_dehacked.cpp
+++ b/common/d_dehacked.cpp
@@ -348,9 +348,16 @@ void A_SpawnSound(AActor*);
 void A_SpawnFly(AActor*);
 void A_BrainExplode(AActor*);
 void A_MonsterRail(AActor*);
-void A_Die(AActor*);
 void A_Detonate(AActor*);
 void A_Mushroom(AActor*);
+void A_Die(AActor*);
+void A_Spawn(AActor*);
+void A_Turn(AActor*);
+void A_Face(AActor*);
+void A_Scratch(AActor*);
+void A_PlaySound(AActor*);
+void A_RandomJump(AActor*);
+void A_LineEffect(AActor*);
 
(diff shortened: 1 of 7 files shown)


AlexMax:
Since this bugfix branch was mostly targeted at Valiant, I decided to throw in a bugfix for `A_Detonate` doing no damage, which Valiant uses for one of the monsters.

PR #135, opened 2020-04-06
co_zdoomphys 1 includes ZDoom's RNG

so-sleepy:
I believe there is a cleaner way to do this, but that requires finding the exact function that sets and unsets CVars.

### DIFF
--- a/common/m_random.cpp
+++ b/common/m_random.cpp
@@ -79,6 +79,13 @@ static const unsigned char rndtable[256] = {
 
 unsigned char rndindex, prndindex;
 
+struct zdoomrng_t {
+	bool is_enabled = 0;
+	unsigned long int table[NUMPRCLASS];
+} zdoomrng;
+
+EXTERN_CVAR(co_zdoomphys)
+
 extern bool step_mode;
 
 //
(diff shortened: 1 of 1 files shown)


so-sleepy:
Differences to consider:

https://github.com/so-sleepy/zdoom-1.22/blob/aa410c460b2126d295f851f3ee8a0ad1fd1b682b/src/m_random.cpp#L75
https://github.com/so-sleepy/zdoom-1.23b33/blob/86a486978a3cf1f24d104f11b3477059ced538bf/src/m_random.cpp#L75
https://github.com/so-sleepy/odamex/blob/4d291d5ee6962a5280a6e56fc0122c187697c9a9/common/m_random.cpp#L134

AlexMax:
I don't think it's wise to tie this to ZDoom physics.

On Zandronum, it was incredibly common to run servers with Doom's original PRNG and ZDoom physics - both 1.23 emulated physics and whatever modern ones happened to be in that version of Zandronum.  Doom's original PRNG was more favourable to the SSG damage rolls, and players really tend to prefer that over true randomness.  In fact, I can't recall a single moment where a player ever asked me to turn Doom's PRNG off.

However, this PR did spark an idea.  I do think that it might be a good idea to have an alternative random number generator that might be used for things that players are not necessarily aware of.  But I think that it should exist in addition to Doom's PRNG, not as a complete substitution.  This would mean keeping two seeds instead of one, and having a separate set of functions that would allow for getting a "better" random value if the right compatibility value was set (or unset).

And in any case, I don't think we're not bound to ZDoom's PRNG for having a better PRNG.  [Xorshift](https://en.wikipedia.org/wiki/Xorshift) can get much better random values in three lines of code and 32 bits of state.

so-sleepy:
> I don't think it's wise to tie this to ZDoom physics.
> 
> On Zandronum, it was incredibly common to run servers with Doom's original PRNG and ZDoom physics - both 1.23 emulated physics and whatever modern ones happened to be in that version of Zandronum. Doom's original PRNG was more favourable to the SSG damage rolls, and players really tend to prefer that over true randomness. In fact, I can't recall a single moment where a player ever asked me to turn Doom's PRNG off, although I'm sure there are some.

https://odamex.net/bugs/show_bug.cgi?id=1304

AlexMax:
I was talking more about my experience running Zandronum servers.  That said, I saw the author of the bug and recognized the name almost immediately, and his advocacy doesn't surprise me.

My point was that I am very much averse to tying the ZDoom PRNG to the ZDoom physics console variable.  I think they should be separate.

so-sleepy:
I will see if I can create a new CVar then, I am closing this PR.

PR #136, opened 2020-04-07
[PROTOBREAK] PRINT_ | Console fixes and additions

Ch0wW:
This PR does the following :
- Transforms '\c' colorcode to '\034' to avoid undesired coloring in Printf messages.
- Fixed PRINTCOLOR_NORMAL to use selected PRINT_HIGH color instead of forcing to CR_GRAY.
- Adds PRINT_WARNING and PRINT_ERROR colorcodes.
- Adds PRINT_NORCON flag (which transforms later into PRINT_HIGH).
- Removes all SV_BroadcastPrintf messages to RCON'd clients to avoid duplicates.
- Stop sending chat/teamchat/serverchat to rcon clients to avoid info-cheating.
- Simplified Printf and SV_BroadcastPrintf so that by default, it uses PRINT_HIGH.
- Transformed some PRINT_HIGH warnings to PRINT_WARNING.
- Colorizes Team-game messages (such as the flag, or the team color) that can be disabled through a new CVAR, `con_coloredmessages`.
- Renames PRINT_LOW to PRINT_PICKUP, PRINT_MEDIUM to PRINT_OBITUARY.
- Added a way to filter either Obituaries or pickup messages instead of giving a messagelevel.


client/src/c_console.cpp
```diff
@@ -852,10 +868,11 @@ static int C_PrintString(int printlevel, const char* color_code, const char* out
...
 	// [SL] the user's message color preference overrides the given color_code
 	// ...unless it's supposed to be formatted bold.
-	if (color_code && color_code[2] != '+' && printlevel >= 0 && printlevel < PRINTLEVELS)
+	static char printlevel_color_code[2];
+
+	if (color_code && color_code[1] != '+' && printlevel >= 0 && printlevel < PRINTLEVELS)
 	{
-		static char printlevel_color_code[4];
-		sprintf(printlevel_color_code, "\\c%c", 'a' + PrintColors[printlevel]);
+		sprintf(printlevel_color_code, "\034%c", 'a' + PrintColors[printlevel]);
```
AlexMax, line 875: Here you are asking `sprintf` to write 3 characters (the two characters above plus the null terminator) into a 2 byte buffer.  In general, you should avoid the use of `sprintf` and prefer the use of `snprintf`.

client/src/c_console.cpp
```diff
@@ -105,8 +109,19 @@ static struct NotifyText
 	byte text[256];
 } NotifyStrings[NUMNOTIFIES];
 
-#define PRINTLEVELS 5
-int PrintColors[PRINTLEVELS+1] = { CR_RED, CR_GOLD, CR_GRAY, CR_GREEN, CR_GREEN, CR_GOLD };
+// Default Printlevel
+#define PRINTLEVELS 8 //(5 + 3)
+int PrintColors[PRINTLEVELS] =
```
AlexMax, line 114: I see that you removed the `+1` here.  I think this causes a problem elsewhere in the program.  Apparently, the implementation of the cvar `msgmidcolor` uses `PRINTLEVELS` as its default value, which means that the function `setmsgcolor` tries to pull an out-of-bounds index into the `PrintColors` array.  I'm not really sure that having `PRINTLEVELS` act as a "zero value" was the best design idea, so see if you can figure out a more elegant way of solving this issue.

Ch0wW:
Closing, as https://github.com/odamex/odamex/pull/187 the new one.

PR #142, opened 2020-04-08
COUNTOF and STACKARRAY_LENGTH to ARRAY_LENGTH

AlexMax:
This PR changes the name of my `COUNTOF` Array Length macro (that I contributed with my first MBF fix patch) to `ARRAY_LENGTH` to be more clear and gets rid of the old `STACKARRAY_LENGTH ` macro that was vulnerable to misuse.


seanmleonard:
I'm a fan of this -- let me know when you think it's done and I'll merge it.

AlexMax:
I think that about covers it.

PR #143, opened 2020-04-10
[BUG #1248] sv_sharekeys CVAR

Ch0wW:
When a key has been found by a player, share that key to all players.

I'ved also removed the CVAR_LATCH property in sv_keepkeys, it has no reason to be there.

Original bug thread : https://odamex.net/bugs/show_bug.cgi?id=1248


AlexMax:
LGTM.

PR #147, opened 2020-04-14
Slightly modify the HUD message while waiting in the queue

tm512:
This is a small cosmetic modification, it adds an ordinal suffix to the HUD message displayed when a client is waiting in the join queue, changing "Position in line to play: N" into "Waiting to play - Nth in line".

### DIFF
--- a/client/src/hu_elements.cpp
+++ b/client/src/hu_elements.cpp
@@ -184,6 +184,23 @@ int teamTextColor(byte team) {
 	return color;
 }
 
+// generate an ordinal suffix from an integer (1 -> "st", 2 -> "nd", etc)
+// used in HelpText() to display position in queue
+static std::string ordinal(int n)
+{
+	n %= 100;
+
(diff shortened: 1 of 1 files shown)


client/src/hu_elements.cpp
```diff
@@ -184,6 +184,23 @@ int teamTextColor(byte team) {
 	return color;
 }
 
+// generate an ordinal suffix from an integer (1 -> "st", 2 -> "nd", etc)
+// used in HelpText() to display position in queue
+static std::string ordinal(int n)
```
nstlaurent, line 189: This is nit-picky but can we just return a const char* here instead to avoid extra object construction.

tm512: Yeah, that's fine. I think std::string was just a carry-over from my old queue code, where I was concatenating the return value with another string, but it looks like std::ostringstream eliminates the need for that.

PR #152, opened 2020-04-18
Introduce StrFormat and VStrFormat

AlexMax:
These functions act like `asprintf`, except instead of formatting to an allocated `char*` it allocates to an existing passed `std::string`.  The reason I chose this approach was that I wanted to leave open the possibility of reusing the same `std::string` multiple times and only incurring the cost of a single unavoidable internal allocation per use.

This commit also includes a sample use of the function elsewhere in the code.


AlexMax:
I found and fixed an issue related to misuse of `va_list`.

PR #155, opened 2020-04-25
Add cvar for resetting the PRNG on map reload

AlexMax:
Turning this off makes initial spawns less predictable.

Pretty much what it says on the tin.  Just thought i'd squeeze this in for upcoming tournaments.


AlexMax:
I made the discovery that predictable first spawns aren't even vanilla-accurate.  Thus, I withdraw this PR and will submit another one that simply disables the feature.

PR #167, opened 2020-05-10
Bindings refactoring & Automap bindings

Ch0wW:
Refactored the bindings, so that all new binding commands won't need to be copy/pasted.
Also added independant automap bindings, and added a menu control dedicated to it.


AlexMax:
Nice work.  My only concern is that if you upgrade from an older version of Odamex and have an existing config, none of the automap keys will be bound by default.

Ch0wW:
The last commit should rectify the issue you mentionned.

client/src/c_bind.cpp
```diff
@@ -570,4 +549,94 @@ std::string C_GetKeyStringsFromCommand(const char *cmd, bool bTwoEntries)
...
+}
+END_COMMAND(bind)
+
+BEGIN_COMMAND(unbind)
+{
+	if (argc > 1) {
+		Bindings.UnbindKey(argv[1]);
+	}
+}
+END_COMMAND(unbind)
```
AlexMax, line 580: Could you add a short little help if argc < 2 instead of printing nothing?

client/src/c_bind.cpp
```diff
@@ -570,4 +549,94 @@ std::string C_GetKeyStringsFromCommand(const char *cmd, bool bTwoEntries)
...
+}
+END_COMMAND(doublebind)
+
+BEGIN_COMMAND(undoublebind)
+{
+	if (argc > 1) {
+		DoubleBindings.UnbindKey(argv[1]);
+	}
+}
+END_COMMAND(undoublebind)
```
AlexMax, line 596: Could you add a short little help if argc < 2 instead of printing nothing?

client/src/c_bind.cpp
```diff
@@ -570,4 +549,94 @@ std::string C_GetKeyStringsFromCommand(const char *cmd, bool bTwoEntries)
...
+}
+END_COMMAND(ambind)
+
+BEGIN_COMMAND(amunbind)
+{
+	if (argc > 1) {
+		AutomapBindings.UnbindKey(argv[1]);
+	}
+}
+END_COMMAND(amunbind)
```
AlexMax, line 611: Could you add a short little help if argc < 2 instead of printing nothing?

leximayfield (changes requested):
Mostly looks good, just have a few small things to take care of and I'll merge ASAP.

leximayfield (approved):

AlexMax:
LGTM

PR #171, opened 2020-05-16
3 way ctf - support for green team

nstlaurent:
Includes a lot of refactoring for teams. Started a crude TeamInfo and refactored team related code to fetch the TeamInfo. Refactored team related functions to iterate through NUMTEAMS or sv_teamsinplay where appropriate. This should drastically reduce the changes required for adding a new team or making any team code changes.

For more than two teams CTF messages/sounds will not be possessive if not your team events.

HUD is now dynamic for teams checking sv_teamsinplay for both the hi-resolution and low-resolution HUDs. They will both now grow as large as possible, and clamp to the top left of the screen when too big, and just keep drawing off the screen if there is not enough room.


AlexMax:
Should the green flag be spawning when `sv_teamsinplay` is 2?  That doesn't seem right to me...

PR #175, opened 2020-07-05
Merge in WDL stats

AlexMax:
This PR just merges in WDL statistics from `master`, along with a bugfix or two...


client/src/v_video.cpp
```diff
@@ -245,10 +245,27 @@ CVAR_FUNC_IMPL(vid_pillarbox)
 		V_ForceVideoModeAdjustment();
 }
 
+//
```
fuzztooth, line 248: This doesn't seem related and should probably be in a separate PR, but I realize how annoying that bug is.

PR #176, opened 2020-07-05
Fix some mingw building weirdness

AlexMax:
Other pull requests were failing because CMake was finding zlib in the PostgreSQL directory of the CI machine, leading to all sorts of trouble, as implied by the name of the branch.

This fix introduces some new dependent options that force CMake to always use the internal versions of zlib and libpng.  These options are set to true on Windows and false everywhere else.

I also decided to get rid of the ambiguity between [LoadMenu](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-loadmenua) and our internal menu definition also named `LoadMenu`.  This change was my first attempt at a fix before I realized what the actual problem was, but honestly I think it's probably a better idea to leave it like that.


CMakeLists.txt
```diff
@@ -13,11 +13,19 @@ include(GNUInstallDirs OPTIONAL)
...
+else()
+  set(USE_INTERNAL_LIBS 0)
+endif()
+
 # options
 option(BUILD_CLIENT "Build client target" 1)
 option(BUILD_SERVER "Build server target" 1)
 option(BUILD_MASTER "Build master server target" 1)
 option(BUILD_ODALAUNCH "Build odalaunch target" 1)
+cmake_dependent_option( USE_INTERNAL_ZLIB "Use internal zlib" ${USE_INTERNAL_LIBS} BUILD_CLIENT 0 )
```
HyperEye, line 27: Based on my reading of cmake_dependent_option the USE_INTERNAL_LIBS variable can be overridden by the user if they configure with USE_INTERNAL_ZLIB and USE_INTERNAL_PNG set to true. Assuming that is correct I think this is fine for WIN32 build.

PR #179, opened 2020-07-30
Add GCC attributes to functions

AlexMax:
This patch introduces wrapper defines for two GCC "attributes".  One of them checks the formatting of printf-style functions, and the other annotates a function as not returning, which should squelch warnings that arise from the compiler incorrectly assuming it can advance past a noreturn function call.

`FORMAT_PRINTF` was applied to the three print functions and `StrFormat`.  `NORETURN` was applied to `I_FatalError`.  MSVC has an equivalent for noreturn, but not for format, and the define accounts for this.

### DIFF
--- a/common/c_cvars.cpp
+++ b/common/c_cvars.cpp
@@ -782,6 +782,13 @@ BEGIN_COMMAND (help)
 END_COMMAND (help)
 
 // [AM] Crash Odamex on purpose - with no survivors.  Used for testing crash handlers.
+BEGIN_COMMAND(errorout)
+{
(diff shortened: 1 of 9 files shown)


common/c_cvars.cpp
```diff
@@ -781,6 +781,13 @@ BEGIN_COMMAND (help)
 }
 END_COMMAND (help)
 
+// [AM] Crash Odamex on purpose - with no survivors.  Used for testing crash handlers.
```
fuzztooth, line 784: We're ok with leaving this in?  So long as it can't be executed remotely...

PR #195, opened 2020-10-24
Get Odalaunch built with MSVC

AlexMax:
I basically fixed a bunch of the build weirdness preventing Odalaunch from being built with Visual Studio.  I also went through and fixed two crash bugs that I saw and got it integrated into the CI and demoted the MinGW CI build to build-only, instead of being needed for the artifact.  I also disabled the mac CI completely because homebrew is broken on the github runners and I am unable to figure out how to fix it - no big loss, the artifact didn't work anyway.


AlexMax:
Merging.  This is a fix for the build system, which is covered by CI.

PR #197, opened 2020-11-06
Fix 32-bit build, add to CI

AlexMax:
The 32-bit build wasn't working because the build system wasn't forcing 32-bit flags on the curl CMake invocation.

Also added to CI so we can have 32-bit builds ready to go for testing.


AlexMax:
Merged, pretty much a CI addition.

PR #200, opened 2020-11-11
Output debug string

AlexMax:
Debug string output to the Visual Studio console


AlexMax:
Trivial addition, and it would be really nice to have in other branches.

PR #202, opened 2020-11-11
Add functions to print packet header strings

AlexMax:
Pretty simple - just adds a function for dumping a string for a specific packet type.  Was used to help debug demos in another PR.


AlexMax:
Closed. I had overlooked a similar feature elsewhere.

PR #203, opened 2020-11-11
Add back some ACS functions

Ch0wW:
This PR adds a few functions that are used on ACS, and fixes possible crashs on some cases.


AlexMax:
LGTM

AlexMax (approved):

PR #206, opened 2020-11-25
Flexibilize video emulation

pmjdebruijn:
Currently vid_320x200 and vid_640x400 emulate very specific situation, where
lower resolution emulation is unnecessarily limited and unnecessarily entangled
with nonwidescreen.

This introduces vid_emulines, which only sets the height of the emulated
resolution and the width is automatically derived based on the relevant
aspect ratio, which means it's also disentangled from widescreen.

### DIFF
--- a/client/sdl/i_video.cpp
+++ b/client/sdl/i_video.cpp
@@ -79,8 +79,7 @@ EXTERN_CVAR(vid_fullscreen)
 EXTERN_CVAR(vid_vsync)
 EXTERN_CVAR(vid_filter)
 EXTERN_CVAR(vid_overscan)
-EXTERN_CVAR(vid_320x200)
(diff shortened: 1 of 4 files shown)


Ch0wW:
Looks solid! However, I've noticed a vertical line using `vid_emuline` set to `200`, only appearing on widescreen. It doesn't seem to happen on the previous release.

image
image

pmjdebruijn:
I do not get that, what resolution are you using?

Ch0wW:
I tried it using 1920x1080, but it happens even on 720p. Coincidentally, the stripes happen exactly on the edge of the screen.
image
image

pmjdebruijn:
I can't reproduce, maybe try moving your odamex config out of the way, start with a fresh config, and then compare which settings might be causing the issue?

PR #207, opened 2020-11-25
Remove poor client CXX_FLAGS

pmjdebruijn:
-march=native that cannot be easily overridden causes problems for doing
binary package builds, where a single binary needs to run a wide variety
of machines.

The APPLE conditional actually means OSX, and OSX hasn't supported PowerPC
for quite some time now, and in the foreseeable future might mean either
AMD64 or ARM64. Therefore CPU architecture specific flags should not
hinge on operating system related conditionals.


AlexMax:
Hey there.

I appreciate you bringing this to our attention, and yeah, that part of the CMake file did jump out as me as potentially problematic.  However, I would like to make sure that omitting `-march=native` does not also disable any of the SSE2 drawers in-tree for people who simply install the app from source, and I have not gotten around to figuring that out.  If that's something you have time to investigate, that would be a load off of my mind, but if not, I plan on taking a look at it myself eventually.

AlexMax:
I have some changes in the `misc-fixes` branch that hopefully should satisfy.  I think this can be safely closed now.

PR #208, opened 2020-11-25
Restore compatibility with the Shareware IWAD

pmjdebruijn:
Odamex' new statusbar/hud depends on CELLA0, CELPA0 and PSTRA0 being present,
which is not the case with the Shareware IWAD

As opposed to introducing all kinds of special case handling in the codebase,
I've opted for a more elegant approach of introducing transparent dummy
replacement sprites into odamex.wad, which won't negatively affect
the proper Registered IWADs as odamex.wad is loaded before any IWAD, so
any IWAD content will supercede odamex.wad


AlexMax:
Thanks for pointing this out.  Internally, we were discussing how big of a pain in the neck it would be to detect when these lumps are missing and simply refuse to draw them, as opposed to having stand-ins.  However, I'm not going to close this because your approach is indeed valid and if nothing comes of the latter solution, I suspect we'll simply accept this one.

PR #211, opened 2020-12-27
Levelstate, Survival, LMS, etc.

AlexMax:
The fundamental shift that this branch makes is that it vastly extends the old "warmup" system to instead function as a system for managing the state of a game no matter what game mode you are using.  Essentially, it provides states like "Warmup", "Ingame" "Preround", "Postround" and such that can be used and transitioned between with any game mode.

On top of this foundation, I have implemented a rounds and lives system.  These are two separate systems, you can enable one or both, and work with all 4 existing game modes.  When rounds are enabled, anything that would normally end the game ends the round instead, and there are separate win condition cvars for round wins.  New HUD elements have been introduced to show lives.

However, this branch also signals my intent to not introduce a ton of new `sv_gamemode` cvars.  I am of the opinion that there should be fewer game modes, but with more configuration available.  Less Quake 3 Arena, more Halo.  You might notice a few flags in here towards that ideal - for example I threw in a "one-side CTF" and "no touch flag return" while I was at it to drive the point home that I want this to be a sandbox.

The complete list of new cvars is as follows:

**g_ctf_notouchreturn** (Boolean, default: 0)
Prevents touch-return of the flag, forcing the player to wait for it to timeout.

**g_rounds** (Boolean, default: 0)
Turns on rounds - if enabled, reaching the gametype's win condition only wins a single round, and a player or team must win a certain number of rounds to win the game.

**g_sides** (Boolean, default: 0) *[AM] I don't like the name "sides" but I couldn't think of anything better.*
Turns on offensive vs defensive sides for team-based game modes - teams on offense don't have to defend any team objective, teams on defense win if time expires, and if rounds are enabled the team on defense changes every round

**g_roundlimit** (Integer, default: 0)
Number of total rounds the game can go on for before the game ends

**g_lives** (Integer, default: 0)
Number of lives a player has before they can no longer respawn - this can result in a game loss depending on the gametype.

**g_lives_jointimer** (Integer, default: 0)
When players have limited lives, this is the number of seconds after the round or game started that new players can join - if set to zero, players need to queue and wait until the next round to join.

**g_winlimit** (Integer, default: 0)
Number of times a round must be won before a player or team wins the game.

**g_speclosers** (Boolean, default: 0)
After a game is won, force losing players or teams to spectate.

**g_preroundtime** (Integer, default: 5)
Amount of time before a round where you can't shoot.

**g_postroundtime** (Integer, default: 3)
Amount of time after a round before the next round/endgame.

In addition, I have a couple of built-in console commands that aim to set up some common game modes; **survival**, **lms** and **tlms**.  These commands are self-documenting:

```
tlms
[19:49:33] tlms - Configures some settings for a basic game of Team Last Marine Standing

Flags:
  lives <LIVES>
    Set number of player lives per round to LIVES.  Defaults to 1.
  teams <TEAMS>
    Set number of teams in play to TEAMS.  Defaults to 2.
  wins <WINS>
    Set number of round wins needed to win the game to WINS.  Defaults to 5.
tlms wins 3
[19:49:43] Configuring Team Last Marine Standing...
sv_gametype 2; sv_skill 5; sv_nomonsters 1; sv_forcerespawn 1; g_lives_jointimer 0; g_rounds 1; sv_friendlyfire 0; g_winlimit 3; g_lives 1; sv_teamsinplay 2
[19:49:43] sv_gametype will be changed for next game.
[19:49:43] sv_skill will be changed for next game.
[19:49:43] sv_nomonsters will be changed for next game.
[19:49:43] g_rounds will be changed for next game.
[19:49:43] g_lives will be changed for next game.
```

Most of the other changes in this branch, like the rework of the join queue, were either done out of necessity or to leave code that I had to change anyway in a better state than I found it.

Probably the most notable change on the protocol level is the introduction of a "varint" type - the idea taken from Google's Protocol Buffers - in order to encode int32's and int64's in a way that allows for the full width of the type (as opposed to a wing and a prayer that a short is enough) but also doesn't waste bandwidth for small numbers.  A more technical breakdown of how it works can be found [here](https://developers.google.com/protocol-buffers/docs/encoding) and I have also licensed my particular implementation of varints permissively, just in case we ever need it for the launcher.  I used varints when reworking network messages I had to touch anyway, and in some cases the space savings were substantial.


AlexMax:
So the big question is - why should this be merged?  It's huge, after all.

Several testing sessions have been done with this branch.  I'm not sure as I would go so far as to say it's 100% done, but all of the major bugs have been taken care of, and it's certainly done enough to merge into mainline protobreak in preparation for 0.9.  There will be plenty of testing of protobreak and opportunities for "one last tweak" in the lead-up to 0.9's release.

There are some CI problems, however the problems are with the CI itself.  I have fixes in another branch.

AlexMax:
I took care of a few things I had left undone.

1. I tried to resolve a merge with mainline and in the process discovered the secret sound we committed in `misc-fixes` turned out to be way too long and way too big.  I trimmed the sound, downgraded the quality and converted it to mono.
2. The levelstate debug overlay is gone now.  Whoops.
3. Earlier I made a change to where you can bring up the scoreboard in single-player - this is mostly for debugging purposes and "why shouldn't this work", but seeing it automatically on death was kind of jarring, so I prevented that from happening in single-player.
4. Previously I introduced many "gametype commands" for the new lives-based modes, but realized that the non-lives-based modes lacked them.  To prevent myself from further copy-pasting a bunch of code, I cleaned up the existing code and introduced commands for `coop` `dm` `tdm` and `ctf`.

PR #214, opened 2021-01-02
Fix looking around with keyboard/joystick

AlexMax:
The improved mouse code I believe exposed an issue where turning with the keyboard and joystick stopped working, and looking up and down was incredibly jittery.

I believe this PR fixes both issues.  Now, we interpolate if there is no accumulated yaw or pitch values, and we also fix a bug where accumulated yaw was blasting away any other types of yaw.

### DIFF
--- a/client/src/g_game.cpp
+++ b/client/src/g_game.cpp
@@ -552,10 +552,16 @@ void G_BuildTiccmd(ticcmd_t *cmd)
 	}
 	else
 	{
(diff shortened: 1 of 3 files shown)


Ch0wW:
Looks good to me, tried it, and fixed the joystick/+left/+right with no issues.

nstlaurent (approved):
Looks like a good fix to me

PR #215, opened 2021-01-11
Miscellanous fixes

Ch0wW:
The following PR fixes a few bugs discovered: 

- Fix the statbar to not render tens of lives
- Improved the hud_revealsecrets behaviour online
- Fix Teamkill messages not being displayed
- Display a message in case of an invalid client
- Fix attempt to crash when a player tries to join an unknown team
- Fix serverside CVARs being forced-reset by the client when resetting to default

### DIFF
--- a/client/src/cl_cvarlist.cpp
+++ b/client/src/cl_cvarlist.cpp
@@ -455,8 +455,8 @@ CVAR_RANGE(		hud_fullhudtype, "1","Fullscreen HUD to display:\n// 0: ZDoom HUD\n
 CVAR_RANGE(		hud_gamemsgtype, "2", "Game message type",
 				CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 1.0f, 2.0f)
 
-CVAR(			hud_revealsecrets, "1", "Print HUD message when secrets are discovered",
-				CVARTYPE_BOOL, CVAR_CLIENTARCHIVE)
+CVAR(			hud_revealsecrets, "1", "Print a HUD message and|or play a sound when a secret is discovered.",
+				CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 0.0f, 3.0f)
 
 CVAR(			hud_scale, "1", "HUD scaling",
 				CVARTYPE_BOOL, CVAR_CLIENTARCHIVE)
--- a/client/src/cl_cvarlist.cpp
+++ b/client/src/cl_cvarlist.cpp
@@ -455,7 +455,7 @@ CVAR_RANGE(		hud_fullhudtype, "1","Fullscreen HUD to display:\n// 0: ZDoom HUD\n
 CVAR_RANGE(		hud_gamemsgtype, "2", "Game message type",
 				CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 1.0f, 2.0f)
 
-CVAR(			hud_revealsecrets, "1", "Print a HUD message and|or play a sound when a secret is discovered.",
+CVAR_RANGE(		hud_revealsecrets, "1", "Get a notification if you or another player finds a secret.",
 				CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 0.0f, 3.0f)
 
 CVAR(			hud_scale, "1", "HUD scaling",
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -158,6 +158,9 @@ EXTERN_CVAR (cl_forcedownload)
 // [SL] Force enemies to have the specified color
 EXTERN_CVAR (r_forceenemycolor)
 EXTERN_CVAR (r_forceteamcolor)
+
+EXTERN_CVAR (hud_revealsecrets)
+
 static argb_t enemycolor, teamcolor;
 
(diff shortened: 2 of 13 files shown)


client/src/cl_cvarlist.cpp
```diff
@@ -455,8 +455,8 @@ CVAR_RANGE(		hud_fullhudtype, "1","Fullscreen HUD to display:\n// 0: ZDoom HUD\n
 CVAR_RANGE(		hud_gamemsgtype, "2", "Game message type",
 				CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 1.0f, 2.0f)
 
-CVAR(			hud_revealsecrets, "1", "Print HUD message when secrets are discovered",
-				CVARTYPE_BOOL, CVAR_CLIENTARCHIVE)
+CVAR(			hud_revealsecrets, "1", "Print a HUD message and|or play a sound when a secret is discovered.",
+				CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 0.0f, 3.0f)
```
AlexMax, line 459: This is failing the build on Linux - I think this needs to be a `CVAR_RANGE`.

client/src/d_netinfo.cpp
```diff
@@ -110,6 +110,13 @@ team_t D_TeamByName (const char *team)
 			return (team_t)i;
 	}
 
+	if (stricmp(team, "0") == 0)
+		return TEAM_BLUE;
+	else if (stricmp(team, "1") == 0)
+		return TEAM_RED;
+	else if (stricmp(team, "2") == 0)
+		return TEAM_GREEN;
+
```
AlexMax, line 119: You can get away with a plain old `strcmp` here - case-sensitivity doesn't matter for numbers.

common/g_gametypecmd.cpp
```diff
@@ -140,7 +140,7 @@ static StringList GametypeArgs(const GametypeParam (&params)[SIZE], size_t argc,
  */
 
 static GametypeParam coopParams[] = {
-    {"g_skill", 4, "skill", "SKILL",
+    {"sv_skill", 4, "skill", "SKILL",
```
AlexMax, line 143: Nice catch!

common/g_gametype.cpp
```diff
@@ -659,10 +653,7 @@ void G_LivesCheckEndGame()
 			if (tv.size() == 1)
 			{
 				GiveTeamWins(tv.front()->Team, 1);
-				SV_BroadcastPrintf(
-				    PRINT_HIGH,
-				    "%s team wins for having the highest score with %s left!\n",
-				    tv.front()->ColorString.c_str(), teams);
+				SV_BroadcastPrintf("%s team wins for having the highest score with %s left!\n", tv.front()->ColorString.c_str(), teams);
```
AlexMax, line 656: Please run `clang-format` on this document.

`g_gametype` is a file I created recently, and if I was the one that created the file, I run `clang-format` on it using the settings at the root of the Odamex tree.  In other files, I try to selectively `clang-format` just the parts of the code I commit.

server/src/sv_main.cpp
```diff
@@ -3287,10 +3292,10 @@ void SV_UpdateSecretCount(player_t& player)
...
 	        continue;
 
-	    std::ostringstream buf;
-		
-	    buf << player.userinfo.netname << " found a secret!";
-	    SV_MidPrint(buf.str().c_str(), &(*it), 0);
+		client_t* cl = &(it->client);
+
+		MSG_WriteMarker(&cl->reliablebuf, svc_secretevent);
+		MSG_WriteByte(&cl->reliablebuf, it->id);
```
AlexMax, line 3298: Could you create a function in `msg_server.cpp` to write this out?  It will make it much easier to replace when we upgrade the protocol.

AlexMax (changes requested):
Have left some comments.

client/src/cl_cvarlist.cpp
Ch0wW: Oops, that was something I failed to see! Thanks for noticing it!

client/src/cl_main.cpp
```diff
@@ -2959,6 +2962,26 @@ void CL_ClearSectorSnapshots()
...
+void CL_SecretEvent()
+{
+	player_t& player = idplayer(MSG_ReadByte());
+
+	// Don't show other secrets if requested
+	if (!hud_revealsecrets || hud_revealsecrets > 2)
+		return;
+
+	std::string buf;
+	StrFormat(buf, "%s%s %sfound a secret!\n", TEXTCOLOR_YELLOW, player.userinfo.netname, TEXTCOLOR_NORMAL);
```
AlexMax, line 2978: You forgot a c_str

common/msg_server.cpp
```diff
@@ -267,4 +267,13 @@ void SVC_LevelState(buf_t& b, const SerializedLevelState& sls)
 	MSG_WriteVarint(&b, sls.last_wininfo_id);
 }
 
+/**
+ * @brief Send information about a player who discovered a secret.
+ */
+void SVC_BroadcastSecretFound(buf_t& b, int playerid)
```
AlexMax, line 273: Please call this `SVC_SecretEvent` - named after the marker.

AlexMax (changes requested):
The mac build failed pointing out something potentially serious, and I'd rather there be a better name for the message function.  Other than that, it looks good, just those two things.

AlexMax (approved):

PR #216, opened 2021-01-14
More Teams fixes

Ch0wW:
This PR does the following:
- changeteams' maxvalue is synced with the server
- Disallow serverside green spawning on classic CTF
- The server doesn't forceteam spectators anymore.


AlexMax (commented):
Thanks for putting in the legwork on that bug.

PR #217, opened 2021-01-18
Fix a String attack vulnerability

Ch0wW:
This PR fixes a string vulnerability from `SV_BroadcastPrintf(const char *msg)` that was able to escape % parameters.


AlexMax (approved):
Nice catch!

PR #218, opened 2021-01-25
More Miscellanous fixes

Ch0wW:
This PR Fixes several missing colored strings sent by the server.

Also corrects RCON messages being printed in red rather that yellow, colored keys when sharekeys is enabled and make sv_teamsinplay a CVAR_RANGE.


AlexMax (approved):
LGTM

PR #219, opened 2021-01-26
More Bugfixes client and serverside

Ch0wW:
This PR does the following:
- Fix the incorrect `level.killed_monsters` value when connecting to a server.
- Fix a broadcasted message on CTF with an incorrect parameter.
- Do not interpolate the camera during the teleporting freeze.
- Correct `sv_monstersrespawn` not working midgame.
- Correct a rare occurence of monsters disappearing randomly after changing the difficulty from nightmare to a lower one.
- Fix powerups being carried after callvoting a map.
- (Serverside only) Respawned monsters on Nightmare or from an archvile properly increases the `level.total_monsters` value.
- Remove duplicated messages when a client has RCON access.
- Forbid clients from changing compatibility values clientside in the compability menu.
- Removes unused CVARs clientside and serverside.


common/p_user.cpp
```diff
@@ -117,6 +117,17 @@ void P_ClearPlayerCards(player_t& p)
...
 
+/**
+ * @brief Clear all powerups from a player.
+ *
+ * @param p Player to clear.
+ */
+void P_ClearPlayerPowerups(player_t& p)
+{
+	for (size_t i = 0; i < NUMPOWERS; i++)
+		p.powers[i] = false;
```
AlexMax, line 128: `powers` is not a boolean, you should probably prefer 0.

client/src/m_options.cpp
```diff
@@ -563,23 +567,23 @@ menu_t SoundMenu = {
...
+	{svdiscrete, "Items can drop off ledges",      {&co_allowdropoff},      {2.0}, {0.0}, {0.0}, {OnOff}},
+	{redtext,   " ",								{NULL},                  {0.0}, {0.0}, {0.0}, {NULL}},
+	{bricktext, "Engine Compatibility",				{NULL},                  {0.0}, {0.0}, {0.0}, {NULL}},
+	{svdiscrete, "BOOM actor/sector/line checks",  {&co_boomphys},			 {2.0}, {0.0}, {0.0}, {OnOff}},
+	{svdiscrete, "ZDOOM 1.23 physics",             {&co_zdoomphys},         {2.0}, {0.0}, {0.0}, {OnOff}},
+	{redtext,   " ",								{NULL},                  {0.0}, {0.0}, {0.0}, {NULL}},
+	{bricktext, "Sound",							{NULL},                  {0.0}, {0.0}, {0.0}, {NULL}},
+	{svdiscrete, "Fix silent west spawns",         {&co_nosilentspawns},    {2.0}, {0.0}, {0.0}, {OnOff}},
+	{svdiscrete, "ZDoom Sound Response",			{&co_zdoomsound},		 {2.0}, {0.0}, {0.0}, {OnOff}},
+    {svdiscrete, "Global Pickup Sounds",			{&co_globalsound},		 {2.0}, {0.0}, {0.0}, {OnOff}},
```
AlexMax, line 586: This section has a lot of spacing changes - any reason why?

AlexMax (changes requested):
Just a few changes and questions.

common/p_user.cpp
Ch0wW: I adapted with what existed already - will change it now.

client/src/m_options.cpp
Ch0wW: I'd believe it was due to a cause of bad formatting. I'll try to restore it to a normal state.

AlexMax (approved):
LGTM

PR #212, opened 2020-12-30
Add better support for console ports

Ch0wW:
This PR improves support for console porting on Odamex, by doing the following: 

- Add support for the Nintendo Switch port (CMake) ;
- Convert the compilation for the Nintendo Wii (vWii) to CMake ;
- Adds gameplatform_t enum to track easily the changes per platform ;
- Adds joy_deadzone CVAR to control the deadzone of the joysticks ;
- Adds +fastturn CMD & joy_fastsensitivity CVAR to change the sensitivity on the fly ;
- Changed the way responder keys are triggered (see /client/src/cl_responderkeys.cpp), fixes key issues and improve clarity for several consoles
- Wrote README.md documentation files for the Switch/vWii ;
- Joystick support is changed to Game Controller (SDL2.0 only) ; 

**PLEASE NOTE THAT XBOX KEYS in cl_responderkeys.cpp weren't done since I don't have the proper keynames!**

PR #220, opened 2021-01-27
More misc fixes for 0.9

AlexMax:
For context, `misc-fixes` began life as a branch where small little changes I was making could be amalgamated and Pull-Requested all at once, instead of opening a PR for each individual issue.  However, it has turned out that the faster pace of development is attracting an audience, and this PR actually contains code from PR's #212 #216 #217 #218 .  I also merged in #208 .  I think it would be prudent to merge in this PR before things get any more messy.

Other than those PR's mentioned above, here's what this branch contains

- Fix accidental instances of `g_skill` in the gamemode console commands - it should be `sv_skill`.
- Fix more corner-cases found in the new mouse interpolation branch.
- Add a new `-crashdir` feature that you can use to redirect crash dumps to a specific output directory.
- Address a shortcoming in the protocol where a state number was getting truncated and resulting in garbage psprites in some cases.
- Address an incorrect warmup conditional.
- Fix win condition detection when we hit roundlimit in a teamgame.
- Don't display the join timer in non-survival coop.
- Get rid of the "smart" auto-game-finish behavior in non-lives based gamemodes.
- Don't kill players when they switch teams during warmup.
- New default to a widescreen 1280x720p resolution.  It's 2021, I think it's time.
- Get the intermission timer to show again.
- Add a "FIGHT" message after the round starts.
- Download and use DeuTex to build the WAD file all the time and get rid of our in-repo WAD file - every time we update it it blows up the repository by a few megabytes, and it's a ticking time bomb.
- Don't set crash handlers during debug builds.
- Try to bind to 32 ports instead of just 16.
- Many, many memory handling bugs have been addressed.
- Add support for `$ifgame` to LANGUAGE - this was done to try and get D4V.wad to work, but there's other work to be done.
- Skip `skill` and `clearskills` in MAPINFO, same as above.
- Properly try to reload the original set of WAD files if the new set of WADs fails to load - as opposed to leaving the game in a strange indeterminate state that crashes if you sneeze at it.

As for testing and stability, `misc-fixes` seems to be the branch that the WDL servers test with now.


AlexMax:
#219 I had sent back for changes before creating this PR, the changes were made, I LGTM-ed them.

AlexMax:
And #221 .

AlexMax:
And now #222 .

AlexMax:
Okay I had to merge before the diff got any more out of control.

Last minute polishing and bugfixing for 0.9 just be that way I suppose.

PR #221, opened 2021-01-27
More... More Fixes for Misc-Fixes

Ch0wW:
This PR does the following:
- Fix an OoB found with ASAN and msgmid.
- Properly synchronize total monsters online and respawned monsters.
- Fix the Archvile resurrecting sound for clients. (it isn't played at all)
- Add Lives counter to ZDoom HUD.
- Changed slightly the Match-Broadcasted message in case of a Coop game.

(it's the last one, I swear! 👀)


AlexMax (approved):
LGTM

PR #222, opened 2021-01-28
More and more bugfixes, part Deux

Ch0wW:
This PR does the following:
- Freeze the timer during a warmup, if sv_timelimit is set.
- Give 1 round point for the defending team if `g_sides` is enabled and the timelimit is hit.
- Fixes the new cameraangle interpolation in case of demoplayback.
- Fix softlocks when 1 player on a team has no more lives, and another from the same team leaves/is kicked/spectates.
- Clamp `cl_team` CVAR if the value is higher than sv_teamsinplay.
- Fix alive players to be able to spectate players who don't have lives anymore.

### DIFF
--- a/client/src/am_map.cpp
+++ b/client/src/am_map.cpp
@@ -37,6 +37,7 @@
 #include "m_cheat.h"
 #include "c_dispatch.h"
 #include "cl_demo.h"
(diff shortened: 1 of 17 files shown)


AlexMax (approved):
LGTM

PR #223, opened 2021-01-30
Resolved File/Wanted File abstraction

AlexMax:
This PR basically guts our entire WAD resolution machinery and replaces it with two new types:

**OResFile** - This type represents a "resolved" file that 100% exists in a specific spot on a hard drive with a known hash.
**OWantFile** - This type represents a "wanted" file that may or may not exist and may or may not have an associated hash.

When a WAD file is accepted over the wire from the server or typed in as a command line parameter or entered as a console command parameter, it is a WantFile.  In order to be loaded by Odamex, it must be turned into a ResFile, and there is currently only a single function that does that, and in the process that one function checks all of the user's configured WAD file locations.  Constructing the OResFile also means that the file is hashed.

There are many wins here:
- A file is no longer just a string, but many things tied together.
    - Wanted files have associated wanted type and hash.
    - Resolved Files have a full path to the file and a canonical hash and short name.
- A resolved file and an unknown file are no longer are the same type.
  - Any wanted file is now known to be untrustworthy and should not be taken as gospel.
  - Any resolved file is now known to be a pointer to a valid file with a known hash, short of somebody moving the file or disk failure.  More importantly, it does not have to be checked or further resolved needlessly.
- You cannot modify a wanted or resolved file after you create it.
  - This prevents a resolved file from suddenly becoming invalid after creation. 
- There are now very few places in the code where a wanted file is turned into a resolved file
  - If the logic is defective there's only a few places in the code to check where things could possibly be going wrong.


AlexMax:
I looked over this top to bottom with a fresh pair of eyes, I think it's good to go in.

PR #224, opened 2021-01-30
Flag waypoints for g_sides CTF

AlexMax:
Pretty simple.  When playing g_sides CTF, there is now a hovering flag waypoint pointing at the flag socket of the attacking team.  The color of the flag waypoint is the color of the defending team.

image


AlexMax:
Looked over this with a fresh pair of eyes - it looks good to go.

PR #225, opened 2021-02-02
Added no stop teleport line special 154.

Ru5tK1ng:
Pull request needs review.


AlexMax (approved):
This looks pretty good to me, the only thing I would comment on is the strange-looking formatting.  We have a `.clang-format` that is used when you use Visual Studio's "Format Selection" on code.  I recommend using that on any code you touch for consistency - though not the entire file, that tends to mess up other branches.  Also, our project officially uses 4-width tabs and brace on new line (Allman-style).

PR #226, opened 2021-02-06
Add support 16 rotational frame sprites

DeathEgg:
Support for an extra 8 frames to look at an object from in between the existing frames. Tested with teapot16.wad, code reworked using Doom Retro as a guide.


AlexMax (approved):
LGTM.  I also did some local testing and 16 rotations work as expected, and 8 and single rotation sprites also continue to work.

PR #227, opened 2021-02-07
Correcting several mistakes

Ch0wW:
This PR does the following:
- Fix secrets broadcasting moving items vertically if an item was near another sector (DOOM2 MAP03 for instance)
- Print WARNING & ERROR messages to RCON players
- Warn clients if a server changed a CVAR.
- Do not clear Frags & Deaths after a round of LMS/TLMS/LMSCTF
- Add some MAPINFO declarations & added SETCOMPATFLAG 
- Fix the vertical mouse movement from working.
- Fix netdemo bindings not working, and make them bindable.


common/p_user.cpp
```diff
@@ -136,16 +136,22 @@ void P_ClearPlayerPowerups(player_t& p)
...
 
 	p.lives = g_lives.asInt();
-	p.fragcount = 0;
+
+	// Reset frags/death if not on rounds
+	if (force || G_IsDuelGame() || !G_IsRoundsGame() || !G_IsLivesGame())
+	{
+		p.fragcount = 0;
+		p.deathcount = 0; // [Toke - Scores - deaths]
+	}
```
AlexMax, line 151: This is troublesome.

I created the round system with the intention that each round was like a "mini-game", so any win condition that you could use for a game without rounds would be applied to every round instead.  Thus if you set `sv_fraglimit`, that would be a fraglimit for just that round.

Not resetting the fraglimit every round basically throws this conceit out of the window, making `sv_fraglimit` useless as a way of ending a round.  We could work around that problem, but there's probably two more problems waiting in the shadows if we did.

If we eventually want to add a round-agnostic `totalfrags` or something that doesn't reset between rounds, kind of like how `roundwins` is like now, I can totally understand that.  But not resetting frags between rounds I fear is opening up a can of worms that I would rather keep shut.

common/p_user.cpp
```diff
@@ -136,16 +136,22 @@ void P_ClearPlayerPowerups(player_t& p)
  * @param wins True if a player's wins should be cleared as well - should
  *             usually be True unless it's a reset across rounds.
  */
-void P_ClearPlayerScores(player_t& p, bool wins)
+void P_ClearPlayerScores(player_t& p, bool wins, bool force)
```
AlexMax, line 139: For future reference, I much prefer bitfields to booleans.

AlexMax (changes requested):

AlexMax:
Thanks for addressing the issue I brought up.

PR #232, opened 2021-02-21
MAPINFO overhaul

DeathEgg:
* Added support for UMAPINFO
* Converted Z/MAPINFO parsing to use OScanner
* Added custom episode support


AlexMax:
Thank you for this contribution.  I intend to take a closer look at this after 0.9 is released, and fully intend to merge it assuming there's nothing catastrophically wrong with it, however I at this particular moment I only have bandwidth for 0.9 fixes and a particular piece of 0.10 work.

PR #234, opened 2021-03-04
Reset found keys on survival after a game lost.

Ch0wW:
On coop/survival mode, found keys are added to an array `keysfound`. However, whenever the game must restart on survival (because everyone lost), the array isn't cleared.

As a result, if `sv_sharekeys` is enabled, everyone will get the keys on the newest round, breaking the purpose of the gamemode.

This PR fixes this issue.


AlexMax (approved):
LGTM

PR #235, opened 2021-03-07
Restore FLY ability to spectators after a game starts.

Ch0wW:
Since G_PlayerReborn is called whenever a game is restarted, all spectators have their CF_FLY flag removed, forcing them to the ground. Of course, this is absolutely an unintended behaviour.

This PR **only** disable cheats for non-spectators.


AlexMax (approved):
LGTM

PR #238, opened 2021-03-18
Fix a tile of Doomguy's face being overwritten by the lives counter.

Ch0wW:
Originally, when playing any gamemode without lives, the lives statinfo is still called, overriding the HUD and hiding a part of the mugface when hit to the right.

This commit fixes it by not calling STlib_ClearRect() through STlib_DrawNum() and STlib_updateNum.


leximayfield (approved):
LGTM

PR #241, opened 2021-03-18
Fix +/-lookdown command not being found.

Ch0wW:
The error was caused on a slight error when sorting actionbytes.

This PR corrects it.


leximayfield (approved):
LGTM.

One day it might be a good idea to put a torch to that table and use a real hashtable in the future.

PR #250, opened 2021-03-21
Bump minimum CMake version, add CI for Ubuntu 18.04

leximayfield:
We were running into some issues where folks were using a version of CMake that was too old but was not explicitly disallowed by the cmake version.

I rectified this, while also adding a CI buildbot that uses 18.04 LTS (the distro the users were having trouble with).  This still seems to be a relatively popular distro, and I'd hate for the build to break on it without us knowing it.  Between it and CentOS 7, I think we have the two major LTS distros covered.


fuzztooth (approved):

fuzztooth:
Should we be concerned here?

leximayfield:
I think we're good.  The most recent commit is green, and I think we were just a victim to temporary CI weirdness.

PR #262, opened 2021-03-29
Set IMPORTED_LOCATION for jsoncpp target

ghost:
IMPORTED_LOCATION should be set in order to provide a default location
for build types other than Release and Debug. This patch sets the
default location as that set by JSONCPP_LIBRARY_RELEASE.


fuzztooth (approved):

leximayfield:
There was a momentary glitch in Docker that prevented green lights, but I just reran the thing and everything worked out.

PR #264, opened 2021-04-02
Check for STDISK in an unknown IWAD

leximayfield:
The current list of lumps had too many false positives and was leading to WADs incorrectly being detected as an IWAD when it wasn't.

STDISK seems to be a lump that is the most common sticky wicket that Odamex complains about missing, so I added that lump to the list of lumps to check for.

### DIFF
--- a/common/w_ident.cpp
+++ b/common/w_ident.cpp
@@ -197,13 +197,14 @@ public:
 
 		// [SL] not an offical IWAD.
 		// Check for lumps that are required by vanilla Doom.
-		static const int NUM_CHECKLUMPS = 5;
+		static const int NUM_CHECKLUMPS = 6;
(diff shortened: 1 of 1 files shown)


leximayfield:
Please hold off on this pull request for now, I just realized that I'm not sure if IWAD detection is ever done by checking the header.  If not, that seems like an important omission.

EDIT: [Nevemrind, apparently this cannot be relied upon.](https://doomwiki.org/wiki/WAD)  PR should be good to go.

> The IWAD or PWAD signature is meant to define whether the file is an IWAD or a PWAD, however this is not actually checked by the engine. Loading an IWAD as a PWAD is possible, and inversely loading a PWAD as an IWAD is also possible. For example, CHEX.WAD and TNT.WAD have a PWAD signature despite serving as IWAD.

fuzztooth (approved):

PR #319, opened 2021-04-26
Allow multiple DeHackEd patches, fix language strings

leximayfield:
This patch largely aims to solve problems encountered when trying to layer multiple DEH patches at once.

1. In some instances, only one DeHackEd patch was recognized.  Now, multiples of normal patches, WAD patches, or both are applied.
1. String replacement was inconsistent due to previously only being applied from WAD files when the `language` cvar was changed, as well as only being applied properly when english strings were in use (DEH strings are supposed to override all language-specific strings).

### DIFF
--- a/common/stringtable.cpp
+++ b/common/stringtable.cpp
@@ -225,7 +225,7 @@ void StringTable::loadLanguage(const char* code, bool exactMatch, int pass, char
 	}
 }
 
-void StringTable::loadStringsLump(int lump, const char* lumpname)
+void StringTable::loadStringsLump(const int lump, const char* lumpname, const bool engOnly)
 {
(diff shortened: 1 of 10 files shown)


common/stringtable.cpp
```diff
@@ -338,7 +341,7 @@ bool StringTable::hasString(const OString& name) const
 //
 // Load strings from all LANGUAGE lumps in all loaded WAD files.
 //
-void StringTable::loadStrings()
+void StringTable::loadStrings(const bool engOnly)
```
fuzztooth, line 344: Just a question here - what's the purpose of having a boolean for english only?  That is, what would pass in a true to engOnly?

fuzztooth (approved):

PR #326, opened 2021-04-28
Enable vileghosts behavior by default

DCGRetrowave:
This change enables the vanilla engine's Arch-vile ghost monster resurrect quirk by default on Odamex, allowing this behavior for both single player and multiplayer games.  "co_vileghosts" can be set to "1" to disable this as Boom compatibility fixes the bug.  Several vanilla/limit-removing maps abuse this bug as a feature so this will allow those maps to run as intended (and it's always fun to witness this obscure glitch occur in maps where there is no intentional ghost trap).


common/c_cvarlist.cpp
```diff
@@ -277,6 +277,9 @@ CVAR_RANGE(sv_countdown, "5",
 	CVAR(			co_allowdropoff, "0", "Allow monsters can get pushed or thrusted off of ledges",
 					CVARTYPE_BOOL, CVAR_ARCHIVE | CVAR_SERVERINFO | CVAR_LATCH)
 
+	CVAR(			co_vileghosts, "0", "Disables vanilla's ghost monster quirk that lets Arch-viles resurrect crushed monsters as unshootable ghosts",
```
fuzztooth: This should be called co_novileghosts since setting this to 1 _disables_ thereby changing vanilla behavior.

fuzztooth (changes requested):

leximayfield:
Something has gone incredibly wrong with your most recent commit.  You seem to have accidentally deleted a good number of files from your repository.

In the interest of getting your changes in, I am going to see if I can construct a PR from you commits.

leximayfield:
I fixed things up for you.  Replaced with #359

PR #330, opened 2021-04-29
Weapon drop feature.

ghost:
Ammo count is yet to be implemented. Disabled by default.

### DIFF
--- a/common/p_interaction.cpp
+++ b/common/p_interaction.cpp
@@ -59,6 +59,7 @@ EXTERN_CVAR(cl_predictpickup)
 EXTERN_CVAR(co_zdoomsound)
 EXTERN_CVAR(co_globalsound)
 EXTERN_CVAR(g_lives)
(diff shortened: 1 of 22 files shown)


common/p_interaction.cpp
```diff
@@ -1031,6 +1034,55 @@ void SexMessage (const char *from, char *to, int gender, const char *victim, con
 	} while (*from++);
 }
 
+
+/*mobjtype_t SL_DropWeapon(AActor* target)
```
fuzztooth: Not sure why this was added.  Please remove.

fuzztooth (changes requested):

fuzztooth (commented):
Also please be sure to update your branch to the latest stable.

common/p_interaction.cpp
ghost: Done.

fuzztooth (approved):

Issue #239, opened 2021-03-18
Doomguy's Mugface doesn't turn around if hit online.


Ch0wW:
On Singleplayer, whenever the Doomguy is hit, he will turn his face left or right, depending on the angle of the attacker.

However on multiplayer, it will always face straight, and never turn his face around.

It looks like that on the client, the attacker is never declared, triggering this bug.

However, it looks like it will require a protocol change to fix this.

leximayfield:
This requires a protocol change and was fixed in the `protobuf`  branch, so will be released with that version.

Ch0wW:
I have added this feature on the Protobuf branch. I'll reopen it if not totally fixed.

PR #332, opened 2021-04-29
Fix mugshot not tracking the attacker online.

Ch0wW:
Send for each damaged player their attacker netID.

When your player is hit, no matter their direction, the mugface is still looking towards us, even if the attack comes from the left and right.
In fact, the server doesn't send out the attacker, and Odamex still believes the attacker is 0 (so a hazard or self-attack).

This is fixed by sending the netid of the attacker (if any) to each damageplayer packet.

Fixes issue #239.


leximayfield (approved):
LGTM

PR #351, opened 2021-05-13
Fix a crash with limited teams.

Ch0wW:
On 2-Teams CTF, if both teams were full, and a new player tried to join in, the game would immediately throw a fatal error, since the game would consider the "Green" team acceptable.

This PR fixes this major flaw by making sure it uses the sv_teamsinplay CVAR rather than MAXTEAMS, as it is properly clamped.


leximayfield:
LGTM

PR #353, opened 2021-05-13
Fix odasrv logs not being colorstripped.

Ch0wW:
In Odamex/Odasrv, every string is logged before being color-stripped.
As a result, gibberish characters appear next to usually colored strings in the logs.

This PR fixes this issue by logging after stripping the colors, not before.


leximayfield (approved):
LGTM

PR #354, opened 2021-05-13
Fix connecting to servers on less than stellar connections

leximayfield:
Oh boy this was a nightmare...

When someone connects to a server, first they get a launcher challenge, then they parlay a token from the launcher into a connection challenge, then they await the receipt of packet sequenced zero to consider the connection attempt complete.

One of our players happened to be on a connection that is downright evil, and had considerable trouble connecting to servers.  Turns out this is because his connection had very little guarantees about ordering of packets, and he was receiving packets from the server that were sequenced with numbers above 0 before he got 0, and when that happens, the client ate pavement and dropped the connection.

I fixed this in a backwards-compatible way by having the server send the first packet, but then wait until the client acked it before sending the rest of the initial state.  I found that the player has a previously-unused `PST_CONTACT` state, so I just used that to hold the player in limbo until we got their ack for seq 0, and then we switch them over, finish the connection sequence, and treat them like any other client.

I also threw in some better clientside messaging during the connection sequence for good measure.


common/c_cvarlist.cpp
```diff
@@ -332,6 +332,9 @@ CVAR(               cl_waddownloaddir, "", "Set custom WAD download directory",
 CVAR(				developer, "0", "Debugging mode",
 					CVARTYPE_BOOL, CVAR_NULL)
 
+CVAR(deb_disconnect, "0", "Show source file:line where a disconnect happens",
```
fuzztooth: deb_?

rakohus:
From a testing perspective, I can confirm that it works.

common/c_cvarlist.cpp
leximayfield: Supposed to stand for `debug` but honestly I had my doubts about that.  Was trying to keep it between 1 and 3 letters.  `d_` i thought might get confused with "Doom" or something else.

leximayfield: I went ahead and changed it to `debug_` based on conversations elsewhere.

fuzztooth (approved):

PR #231, opened 2021-02-18
Add a Serverside CVAR to disable MP-things on Coop.

Ch0wW:
This PR adds a CVAR (`g_coopthingfilter`) that removes MP-only things on Cooperative gamemode.
Values: 
- 0: Spawns all MP things normally.
- 1: Removes only Coop Weapon things, leaving only Exclusive ammo & monsters.
- 2: Removes all MP-Exclusive things no matter the type.

PR #355, opened 2021-05-13
Added a serverside CVAR to disable MP-things on Coop.

Ch0wW:
Rewrite of #231.

This PR adds a CVAR (g_coopthingfilter) that removes MP-only things on Cooperative gamemode, as often requested by people.
Values:

    0: Spawns all MP things normally.
    1: Removes only Coop Weapon things, leaving only Exclusive ammo & monsters.
    2: Removes all MP-Exclusive things no matter the type.


common/c_cvarlist.cpp
```diff
@@ -227,6 +227,9 @@ CVAR(g_preroundtime, "5", "Amount of time before a round where you can't shoot",
 CVAR(g_postroundtime, "3", "Amount of time after a round before the next round/endgame",
      CVARTYPE_INT, CVAR_SERVERARCHIVE | CVAR_NOENABLEDISABLE)
 
+CVAR_RANGE(g_coopthingfilter, "0.0",
+     "Removes cooperative things of the map.\nValues: 1 - Only Coop weapons.\n 2 - All Coop things.",
```
leximayfield: As written, if this gets written into a config file, it will break parsing of said config file.  You need to follow every `\n` with `//` so it shows up as a comment.  See some of the other cvars with multiline help for ideas on how to handle it.

leximayfield (changes requested):

common/c_cvarlist.cpp
Ch0wW: Last commit should fix it. I actually forgot about those newlines.

leximayfield (approved):
LGTM - but apparently Visual Studio disagrees

AlexMax:
Apparently CI disagreed with me, can you fix?

common/c_cvarlist.cpp
```diff
@@ -233,6 +233,12 @@ CVAR(g_preroundtime, "5", "Amount of time before a round where you can't shoot",
 CVAR(g_postroundtime, "3", "Amount of time after a round before the next round/endgame",
      CVARTYPE_INT, CVAR_SERVERARCHIVE | CVAR_NOENABLEDISABLE)
 
+CVAR_RANGE(g_coopthingfilter, "0.0",
+     "Removes cooperative things of the map. Values are:\n" \ 
+     "// 0 - All Coop things are retained (default).\n" \ 
+     "// 1 - Only Coop weapons are removed.\n" \
+     "// 2 - All Coop things are removed.",
```
leximayfield:
```
2021-05-13T12:48:14.5839674Z ##[error]D:\a\odamex\odamex\common\c_cvarlist.cpp(241,1): error C2017: illegal escape sequence [D:\a\odamex\odamex\build\server\odasrv.vcxproj]
```

Visual Studio doesn't like this, apparently.

leximayfield (changes requested):

common/c_cvarlist.cpp
leximayfield: Cool I think we're golden.

leximayfield (approved):

PR #359, opened 2021-05-15
Enable vileghosts behavior by default - try 2

leximayfield:
This is a recreated PR #326 - the second commit had a bunch of deleted files in it, and so close to the release of 0.9.1 I didn't want to wait for the ping-pong of going back and forth just to fix a 10 line patch.


Ch0wW (approved):
Looks good so far!

PR #361, opened 2021-05-15
Fix (again) vanilla demoplaybacks

Ch0wW:
Since I added the g_coopthingfilter CVAR, Vanilla demos were suddenly desyncing (DEMO1 spawning the cyberdemon for instance).

Besides, multiplayer vanilla demos crashed suddenly as it was trying to populate data that never existed.

This PR corrects both issues by:
- Tweaking the g_coopthingfilter CVAR with SERVERINFO & LATCH;
- Make that CVAR server_app only, to avoid clients to desync demos;
- Fix vanilla coop demo playback.

However, it still has one last issue with coop demos: all players are "(null)" and they all have the default colors, but I'm unsure when it was introduced.

### DIFF
--- a/client/src/cl_game.cpp
+++ b/client/src/cl_game.cpp
@@ -1834,6 +1834,8 @@ void G_DoPlayDemo(bool justStreamInput)
 			}
 		}
 
+		demoplayback = true;
+
 		if (!justStreamInput)
 		{
 			player_t &con = idplayer(who + 1);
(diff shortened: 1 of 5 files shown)


leximayfield:
Whoops, didn't realize this was a fix for something that was added in 0.9.1.

To be honest, I'm not a huge fan of this fix simply assuming serverside, but for the release of 0.9.1 I think this is "good enough".

Ch0wW:
Although the PR being closed, the server only control those spawns. I could just revert to the previous commit, but it'd unnecessary make things more complicated for no reason.

Issue #358, opened 2021-05-15
[BUG] MAPINFO errors


lemonzest79:
Hi I'm using Odamex 0.9.0 Windows Build and some wads will not load due to MAPINFO errors, I have quite a few that do not launch

Doom II
haxc.wad HacX 1.2 IWAD MAPINFO:9:Script Error: Unknown MAPINFO token "M_TWITCH"
cafebrk1.wad MAPINFO:4:Script Error:Unknown MAPINFO token "compat_nopassover"
epic2.wad ZMAPINFO:14:Script Error:Expected = got "nojump"
jadearth.wad/deh MAPINFO:5:Script Error:Bad floating point constant "nocrouch"
mayhem18o/p.wad MAPINFO:8:Script Error:Bad floating point constant "next"
mayhem19.wad MAPINFO:8:Script Error:Bad floating point constant "next"
mayhem20.wad MAPINFO:13:Script Error:Unknown lookup string "Wet My Appetite"
mayhem2048.wad MAPINFO:318:Script error:Unknown MAPINFO token "Entertext"YOU"
pc_cp.wad MAPINFO:9:Script error:Unknown MAPINFO token "DOOM 2 UNLEASHED"

Ultimate Doom
lostepis.wad MAPINFO:6:Script Error:Unknown MAPINFO token "Evil Unleashed"

Ch0wW:
I've checked the list of wads, and I've been able to fix the following on my PR. Here's the list with its upcoming state:

```
- hacx.wad : now works
- cafebrk1.wad : now works
- epic2.wad : still not fixed (newmapinfo parser error?)
- jadearth.wad : now works
- MAYhem18O/P.wad : now works
- MAYhem19.wad : now works
- mayhem20.wad : now works
- mayhem2048.wad : still unfixed (believes argument is entertext"you )
- pc_cp.wad : now works
- lostepis.wad : now works
```

rakohus:
Some things have changed between @Ch0wW's comment and @DeathEgg's new MAPINFO branch. I tested this list and here's the current situation:

```
- hacx.wad : works
- cafebrk1.wad : works
- epic2.wad : ZMAPINFO parsing error
- jadearth.wad : MAPINFO parsing error
- MAYhem18O/P.wad : MAPINFO parsing error
- MAYhem19.wad : MAPINFO parsing error
- mayhem20.wad : MAPINFO parsing error
- mayhem2048.wad : works
- pc_cp.wad : works
[...]

PR #362, opened 2021-05-16
Fix several MAPINFO tokens

Ch0wW:
This PR fixes many problems reported from issue #358 .

It currently does the following:
- Creates `compat_nopassover` flag.
- Fix "SKY1/SKY2" "scrollspeed" argument being truly optional.
- Fix the "key" and "episode" tokens not being properly parsed.
- Create a fallback for the name if the "lookup" doesn't work.
- Adds "enterpic", "interpic", "exitpic" dummy tokens.


common/g_level.cpp
```diff
@@ -3355,9 +3376,16 @@ ClusterInfos& getClusterInfos()
 
 
 // P_AllowDropOff()
-bool P_AllowDropOff()
+const bool P_AllowDropOff()
```
leximayfield: You shouldn't return a `const` value type - the `const` is useless.

common/g_level.cpp
```diff
@@ -3355,9 +3376,16 @@ ClusterInfos& getClusterInfos()
...
 {
 	return level.flags & LEVEL_COMPAT_DROPOFF || co_allowdropoff;
 }
 
-VERSION_CONTROL (g_level_cpp, "$Id$")
+const bool P_AllowPassover()
```
leximayfield: You shouldn't return a `const` value type - the `const` is useless.

common/g_level.cpp
```diff
@@ -3355,9 +3376,16 @@ ClusterInfos& getClusterInfos()
...
+{
+	if (level.flags & LEVEL_COMPAT_NOPASSOVER)
+		return false;
+
+	return co_realactorheight;
+}
+    VERSION_CONTROL (g_level_cpp, "$Id$")
```
leximayfield: Weird spacing here?

common/sc_man.cpp
```diff
@@ -448,6 +448,31 @@ BOOL SC_GetFloat (void)
 	}
 }
 
+const bool SC_CheckFloat()
```
leximayfield: You shouldn't return a `const` value type - the `const` is useless.

leximayfield (changes requested):
Just some `const` changes.

Ch0wW:
Latest commit should take care of these small errors.

leximayfield (approved):
LGTM

PR #365, opened 2021-05-16
Fix door inconsistencies.

Ch0wW:
This PR fixes several issues related to doors. Here's a small list of them:

- Fixes an issue that happened with doors. If the door sector has a linedef in it, it checked the wrong data, resulting in invalid behaviour. (This fixes lostepis.wad on E5M3 or even the generalized door on Deus Vult II MAP01)
- Fixes Door sounds being not heard when manually reopening or closing them.
- Fix a Door desync that happened since Odamex 0.6.4 as there was a 1-unit too deep check related to door movements.


Ch0wW:
This PR also fixes door desyncs starting with odamex 0.6.4 and door sounds being properly spawned when reopening/closing them.

(Also, please note the door sounds doesn't seem to happen online when not predicting sectors)

leximayfield (approved):

PR #368, opened 2021-05-18
Fix Z-Teleport on vanilla demo playback

Ch0wW:
On TNT/Plutonia, playing back vanillademos caused a desync, because the z destination was always on the ground, which definitely wasn't the proper behaviour.

This PR attempts to fix the issue by making a special case while playbacking a demo only.


leximayfield (approved):
LGTM

PR #369, opened 2021-05-18
Introduce Lump Handles, and convert all graphic patches to use them

leximayfield:
Using raw pointers to WAD content is like handling an unstable chemical - it can sometimes blow up in your face if the pointer goes stale or if you don't clean up after yourself properly.

To that end, I have introduced a concept called a lump handle.  These handles are unique ID's that can passed to WAD functions for their original lump....unless you've loaded a new WAD file, in which case they are invalid and return nothing.

The main thrust of this pull requests is to convert much of the raw `patch_t*` usage in Odamex to use these handles.  In this specific case, if the patch handle is invalid, we just return a blank patch instead of crashing.


common/w_wad.h
```diff
@@ -88,12 +88,34 @@ typedef enum {
 	ns_colormaps,
 } namespace_t;
 
+struct lumpHandle_t
+{
+	size_t id;
+	lumpHandle_t() : id(0)
```
HyperEye, line 94: Consider making the constructor explicit

HyperEye (approved):

common/w_wad.h
HyperEye: Disregard. The constructor is a default constructor. There is no constructor that takes a parameter.

PR #372, opened 2021-05-19
Fix for g_coopthingfilter

Ch0wW:
This PR does the following:

* Fix `g_coopthingfilter` that actually filtered all MP things regardless of settings.
* Don't include Chex Quest for the teleporter bug when playbacking demos.

### DIFF
--- a/common/p_teleport.cpp
+++ b/common/p_teleport.cpp
@@ -282,7 +282,7 @@ BOOL EV_LineTeleport (line_t *line, int side, AActor *thing)
 
 				fixed_t destz;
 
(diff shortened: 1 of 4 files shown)


common/p_teleport.cpp
```diff
@@ -282,7 +282,7 @@ BOOL EV_LineTeleport (line_t *line, int side, AActor *thing)
 
 				fixed_t destz;
 
-				if (demoplayback && (gamemission == pack_tnt || gamemission == pack_plut || gamemission == chex))
```
leximayfield: Why was this changed?

leximayfield (approved):

PR #381, opened 2021-05-26
Various Oscanner fixes

DeathEgg:
* fix bug where quotation mark escape characters weren't properly handled in quote strings
* fix bug where basic strings would continue reading characters in if a quote string started without whitespace/some kind of delimiter
* add function to check if the last thing read in was a quote string


leximayfield (approved):
LGTM - nice catch.

Issue #274, opened 2021-04-16
[BUG] Doom2/Plutonia/TNT Map08 Global Sound


rakohus:
**Describe the bug**
Currently, when vanilla sound is enabled (co_zdoomsound = 0), sounds are played globally at full volume like they are in EXM8 of UDoom. Over the past week, I've joined in games and players really seem to hate this happening.

**Build that the bug occurred in**
0.9 Stable

**To Reproduce**
Make sure "co_zdoomsound = 0", go to Doom 2 Map08, listen to any sound.

**Expected behavior**
Make exception only apply to EXM8? There is some argument as to whether or not this is one of the intentional "tricks" of the map.

**Screenshots, NetDemos, & Crash Dumps**
N/A

**Additional context**
N/A

Ch0wW:
I have to admit that replicating that bug is not an enjoyable feature (despite our desires to keep the port *as close* as the original) and I'm totally fine to remove it.

rakohus:
This has been solved with PR #382

PR #382, opened 2021-05-26
Update map08 sound

Ch0wW:
This PR corrects a misbehaviour related to MAP 8 that Odamex introduced.

This PR removes the fact that the sound is way louder the closer it is to the player. That seems to be a nuisance for many players. ( issue #274 )

However, Odamex never applied the minimum attenuation due to the wrong formula. To correct this, we're doing a small hack that only leaves the minimum attenuation (which is of 0.192 [= 15/128]) on vanilla sounds ONLY (=`co_zdoomsound 0`, so that the experience is still a minimum accurate to players.

Finally, that PR removes that same check for MAP08, because it was never applied, and GZDoom never cared about that specific map behavior.


rakohus:
I tested this build and it works well. Approved from a functionality standpoint.

PR #383, opened 2021-05-29
Release 0.9.2

leximayfield:
Since we're all released, just merging 0.9.2 into stable.


leximayfield:
I think it's fair to fast-track this, since it's the thing we released.

PR #388, opened 2021-05-31
Replace entire server-to-client protocol with Google's Protocol Buffers

leximayfield:
Just what it says on the tin.  The entire server-to-client protocol has been replaced with a protocol that uses Google's [Protocol Buffers](https://developers.google.com/protocol-buffers).

Why?

- A widely-used serialization format, where [definition files](https://developers.google.com/protocol-buffers/docs/proto3) are compiled into [usable source code](https://developers.google.com/protocol-buffers/docs/cpptutorial).
- We can tell how big a message is before we write it to the wire, allowing us to fit more data per packet.
- It is much easier to upgrade a message in a backwards-compatible fashion, since every field of every message is optional.
- Every number that isn't specifically marked as "fixed" is a variable-sized integer, saving space on the wire and leading to less fretting over how big an integer should be.
- This protocol is much easier to debug, because it has a built-in debug printer.
  - In fact, included with this PR is a new demo feature that allow you to debug the network messages recorded in a demo.

A few notes:

- We use Protocol Buffers 3.5, because that is the last version that supports C++98.
- SVC is the only protocol that has been significantly touched.  CLC, FArchive serialization, demo framing and SQP have not been touched.
  - Not for any high-minded reason, but more because SVC was the most critical to convert, and I don't think that the others should prevent the work done thus far from getting merged.
- Many messages have been opportunistically modified or removed - down to 65 from the 88 we have in the stable branch.
  - Any message type that did not seem to be used was removed.
  - There is now only one "force disconnect" message.
  - Compression has been removed as a message type and changed to a "flags" field added to every SVC message.
  - Skipping past resent packets is now done automatically instead of relying on a message.
  - CTF updating has been split into two messages for clarity.
  - Maplist updates have obtained a special dictionary-based compression mechanism in order to get rid of the homespun message-specific fragmentation.


leximayfield:
This has seen some testing and reception has been positive.  I think that this is safe to merge.

PR #390, opened 2021-06-01
Add BOOM DIGFONT

leximayfield:
This commit adds a slightly modified version of DIGFONT.  The font comes from BOOM, but I sourced the lumps from Eternity specifically and added a gradient at some point, though I forget if that was me or Hex.  The font is actually a transplant from the `hud-rework` branch, but I realized that I might need font sooner for various tasks, since it's so small and compact.

As an initial implementation, I replaced the font used in the demo player HUD with DIGFONT, since it's more compact.

image


Ch0wW:
I really love how readable this is!

leximayfield:
With the possible quick turnaround time for a 0.9.3 for the IDL, I retargeted to `protobreak`.  There's no chance that there is a protocol break, but it also updates `odamex.wad` and I don't want to break people's custom WAD's.

fuzztooth (approved):

PR #391, opened 2021-06-01
Add an alternate PWO system.

Ch0wW:
Odamex properly supports PWO, however some people noticed in other MP ports that weapon switching was "ignored" if they were holding `+fire` while picking up a new weapon. 

This PR adds a new, alternate PWO value called `Dynamic PWO` that follows that idea.

### DIFF
--- a/client/src/cl_cvarlist.cpp
+++ b/client/src/cl_cvarlist.cpp
@@ -218,7 +218,7 @@ CVAR_RANGE			(cl_chatsounds, "1", "Plays a sound when a chat message appears (0 
 
 CVAR_RANGE(			cl_switchweapon, "1", "Switch upon weapon pickup (0 = never, 1 = always, " \
 					"2 = use weapon preferences)",
-					CVARTYPE_BYTE, CVAR_USERINFO | CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 0.0f, 2.0f)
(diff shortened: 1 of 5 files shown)


rakohus:
Good idea, but "dynamic PWO" is not clear. This phrase does not tell the user what the setting does. "By Preference, except when firing" would be more descriptive.

client/src/cl_cvarlist.cpp
```diff
@@ -218,7 +218,7 @@ CVAR_RANGE			(cl_chatsounds, "1", "Plays a sound when a chat message appears (0
 
 CVAR_RANGE(			cl_switchweapon, "1", "Switch upon weapon pickup (0 = never, 1 = always, " \
 					"2 = use weapon preferences)",
```
leximayfield: Please update the help text here.

client/src/m_options.cpp
```diff
@@ -675,14 +675,15 @@ menu_t NetworkMenu = {
 static value_t WeapSwitch[] = {
 	{ 0.0,			"Never" },
 	{ 1.0,			"Always" },
-	{ 2.0,			"By Preference" }
+	{ 2.0,			"By Preference" },
+    { 3.0,			"Dynamic PWO"}
```
leximayfield: I agree with Ralphis.  How about something like... `Attack Cancels Pref.`

leximayfield (changes requested):
Just change some docs and this looks good to me.

PR #395, opened 2021-06-04
add functions to OScanner for getting types and checking that types are correct

DeathEgg:
What it says on the tin. Does all of the checks that otherwise would have to be handwritten every time. Also will be useful for replacing sc_man.h/.cpp in the last places it's used in.

### DIFF
--- a/common/oscanner.cpp
+++ b/common/oscanner.cpp
@@ -22,6 +22,7 @@
 
 #include "oscanner.h"
 
+#include "cmdlib.h"
 #include "version.h"
 
 #include <algorithm>
@@ -313,6 +314,9 @@ bool OScanner::compareToken(const char* string) const
 	return _token.compare(string) == 0;
 }
 
+//
+// Print given error message.
+//
 void OScanner::error(const char* message)
 {
 	I_Error("%s", message);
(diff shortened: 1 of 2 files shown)


common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
+	{
+		str[str.length() - 1] = '\0';
+	}
+
+	if (IsNum(str.c_str()) == false || str != "MAXINT")
+	{
+		error("Missing integer (unexpected end of file).");
```
AlexMax, line 432: This doesn't seem to me like an EOF condition

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
+	{
+		str[str.length() - 1] = '\0';
+	}
+
+	if (IsRealNum(str.c_str()) == false)
+	{
+		error("Missing floating-point number (unexpected end of file).");
```
AlexMax, line 457: This also does not look to me like an EOF condition.

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+void OScanner::mustGetBool()
+{
+	if (!scan())
+	{
+		error("Missing boolean (unexpected end of file).");
+	}
+	
+	if (stricmp(_token.c_str(), "true") != 0 && stricmp(_token.c_str(), "false") != 0)
+	{
+		error("Missing boolean (unexpected end of file).");
```
AlexMax, line 473: This also doesn't look like an yadda yadda you get the picture. 😃

leximayfield (approved):

leximayfield:
@DeathEgg I accidently hit "approve" instead of "Request Changes".

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+//
+// Get token as an int.
+//
+int OScanner::getTokenAsInt() const
+{
+	// fix for parser reading in commas
+	std::string str = _token;
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
```
leximayfield: Can't you use std::string::back() here?

common/oscanner.cpp
leximayfield: Seems like a bad copy-paste of an error message.

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+	if (!scan())
+	{
+		error("Missing integer (unexpected end of file).");
+	}
+
+	// fix for parser reading in commas
+	std::string str = _token;
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
```
leximayfield: std::string::back()

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+//
+// Get token as a float.
+//
+float OScanner::getTokenAsFloat() const
+{
+	// fix for parser reading in commas
+	std::string str = _token;
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
```
leximayfield: std::string::back()

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+	if (!scan())
+	{
+		error("Missing floating-point number (unexpected end of file).");
+	}
+
+	// fix for parser reading in commas
+	std::string str = _token;
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
```
leximayfield: std::string::back()

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+	{
+		str[str.length() - 1] = '\0';
```
leximayfield: Wouldn't it be better to resize the string instead of forcing a null byte here?

common/oscanner.cpp
leximayfield: Error message copy paste

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+// Ensure next token is a bool.
+//
+void OScanner::mustGetBool()
+{
+	if (!scan())
+	{
+		error("Missing boolean (unexpected end of file).");
+	}
+	
+	if (stricmp(_token.c_str(), "true") != 0 && stricmp(_token.c_str(), "false") != 0)
```
leximayfield: We have a function specifically for dealing with std::string case-insensitive comparisons:

https://github.com/odamex/odamex/blob/stable/common/cmdlib.cpp#L240

common/oscanner.cpp
leximayfield: Looks like an accidental error message copypaste.

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+		error("Missing integer (unexpected end of file).");
+	}
+
+	// fix for parser reading in commas
+	std::string str = _token;
+
+	// remove comma if necessary
+	if (str[str.length() - 1] == ',')
+	{
+		str[str.length() - 1] = '\0';
```
leximayfield: Can we not resize this string?  Forcing a null byte seems incorrect to me.

common/oscanner.cpp
```diff
@@ -326,4 +332,146 @@ bool OScanner::isQuotedString() const
...
+
+	return static_cast<float>(num);
+}
+
+//
+// Get token as a bool.
+//
+bool OScanner::getTokenAsBool() const
+{
+	return stricmp(_token.c_str(), "true") == 0;
```
leximayfield: We have a function specifically for dealing with std::string case-insensitive comparisons:

https://github.com/odamex/odamex/blob/stable/common/cmdlib.cpp#L240

leximayfield (changes requested):
Got a few changes, but nothing serious.

common/oscanner.cpp
DeathEgg: fixed

leximayfield (approved):

leximayfield:
LGTM, thanks for the fixes

PR #402, opened 2021-06-07
Fix Teamchats & filters as a spectator.

Ch0wW:
This PR corrects an oversight with the spectator filter.

Despite properly being able to mute other spectators (useful for streams), it could have been bypassed by sending a team_chat message.

### DIFF
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -2135,14 +2135,16 @@ void CL_Say()
 	}
 
 	const char* name = player.userinfo.netname.c_str();
+	printlevel_t publicmsg = filtermessage ? PRINT_FILTERCHAT : PRINT_CHAT;
+	printlevel_t publicteammsg = filtermessage ? PRINT_FILTERCHAT : PRINT_TEAMCHAT;
+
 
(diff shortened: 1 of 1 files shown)


client/src/cl_main.cpp
```diff
@@ -2153,11 +2155,11 @@ void CL_Say()
...
 	{
 		if (strnicmp(message, "/me ", 4) == 0)
-			Printf(PRINT_TEAMCHAT, "* %s %s\n", name, &message[4]);
+			Printf(publicteammsg, "* %s %s\n", name, &message[4]);
 		else
-			Printf(PRINT_TEAMCHAT, "%s: %s\n", name, message);
+			Printf(publicteammsg, "%s: %s\n", name, message);
 
-		if (show_messages && cl_chatsounds)
+		if ((show_messages && cl_chatsounds) && !filtermessage)
```
leximayfield: The extra set of parenthesis is unnecessary.

leximayfield (changes requested):
Just a small change.

fuzztooth (approved):

PR #403, opened 2021-06-07
DEHEXTRA & Partial MBF21 support

Ch0wW:
This PR does the following:
- Add support for DEHEXTRA , whose characteristics are available here: https://doomwiki.org/wiki/DEHEXTRA
- Fix the `Speed` DEHacked Token that wasn't working at all on all conditions;
- Fix the `ID #` DEHacked Token that wasn't working;
- Add support for Doom Retro's `Gib health` token (may require further testings as I have seen no example using this token)
- Fix Translated Translucent things being broken in 32Bpp colors.
- Fix small issues related to MBF's functions online.
- Partial (broken) support of BOUNCE flag.
- Moved MF_SPECTATOR to MFO_SPECTATOR, making code clearer for future instances.
- Added almost complete MBF21 support for Weapons tokens (currently lack FLEEMELEE)
- Added MBF21 support for Bits and Args1-8 support for frames.


Ch0wW:
Mancubus' shots are slower for some reason.

Ch0wW:
From last commit, most basic DEHACKED flags should work quite fine, even better than before.

Last commit also includes a very basic (yet buggy) support of the `BOUNCES` bitflag.

Ch0wW:
Latest commit fixed the translated translucent things being corrupted in 32Bpp.

I'm confident a review can be done in its current state.

Ch0wW:
Related to the last commit, about MBF21's weapon additions.

Some changes or infos have to be noted:
- "FLEEMELEE" flag cannot be done yet, as it seems to miss a MBF function.
- Odamex's default behavior in case of "ammo switching" is now MBF21's, but is still reverted to original Vanilla behavior when demoplaybacking.

leximayfield:
This was continued by #483.

PR #409, opened 2021-06-17
HUD updates

leximayfield:
- Obituaries now appear in a "killfeed" in the top right corner of the screen.  Some of the icons are kind of rough - they can be modified later.
- Obituaries no longer appear in the message line by default.
- Obituaries and pickup messages always appear in the console regardless of their message line visibility.
- HUD elements now get their own submenu under Display Settings.
- BIGFONT now works with color translations, and the big levelstate message is now color-translated as well as the big HUD elements.
- Translucent elements are now drawn fully opaque with the opaque drawers if their translucency is set to 1.0, as opposed to being drawn 99.99999999% translucent with the lucent drawers.
- Remove the "ZDoom" HUD setting.  It didn't look different enough from Odamex's fullscreen HUD to make it worth keeping around.


fuzztooth (approved):
No way to really review this, so approving and hoping for the best.

PR #438, opened 2021-07-25
Added MAPINFO flag USEPLAYERSTARTZ to enable z-height spawning.

Ru5tK1ng:
When this flag is set in MAPINFO, a level(s) will allow for z-height spawning. Drop down from mid-air and bridge thing spawning is now possible.


Ru5tK1ng:
This also fixes a trivial issue with cl_showspawns' fountains where they would always take z-height into consideration and spawn in the sky despite all spawns in the map being on the floor.

leximayfield (approved):
LGTM

leximayfield:
@Ru5tK1ng The changes looked good, but protobreak has changed a bit.  If you can merge in the latest `protobreak`, I'll get it in soon after.

leximayfield (approved):
LGTM

PR #440, opened 2021-07-30
Bring back the classic intermission screen to Odamex.

Ch0wW:
This was done originally with #80 but I rewrote it since it broke compilation with no apparent reason.

This commits reenables `wi_newintermission`'s original functionality that was disabled long ago.

It also partially fixes coop movies that were desynced due to Odamex's intermission's timer method.

Also, it is disabled automatically whenever 5+ players are playing at the same time online.

COOP demo:
image

Online: 
image

### DIFF
--- a/client/src/cl_cvarlist.cpp
+++ b/client/src/cl_cvarlist.cpp
@@ -180,7 +180,7 @@ CVAR_RANGE_FUNC_DECL(msgmidcolor, "5", "Color used for centered messages.",
 // ------------
 
 // Determines whether to draw the scores on intermission.
-CVAR(				wi_newintermission, "0", "Draw the scores on intermission",
+CVAR(				wi_newintermission, "0", "Use Odamex's intermission screen if there are 4 players or less on cooperative gamemodes.",
 					CVARTYPE_BOOL, CVAR_CLIENTARCHIVE)
 
 
--- a/client/src/wi_stuff.cpp
+++ b/client/src/wi_stuff.cpp
@@ -41,6 +41,7 @@
 #include "hu_stuff.h"
 #include "v_palette.h"
 #include "c_dispatch.h"
+#include "v_text.h"
 #include "gi.h"
 
 void WI_unloadData(void);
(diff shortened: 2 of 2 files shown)


rakohus:
Is there an elegant way we can include the player names in this scenario?

client/src/wi_stuff.cpp
```diff
@@ -1630,13 +1648,17 @@ void WI_Drawer (void)
...
-			if (multiplayer && sv_maxplayers > 1)
+			if (multiplayer)
 			{
-				// TODO: Fix classic coop scoreboard
-				//if (sv_gametype == 0 && !wi_newintermission && sv_maxplayers < 5)
-					//WI_drawNetgameStats();
-				//else
-					WI_drawDeathmatchStats();
+				if (demoplayback)
+					WI_drawNetgameStats();
```
leximayfield, line 1690: I'd prefer it if either both parts of an `if` statement have braces or neither.

client/src/cl_cvarlist.cpp
```diff
@@ -180,7 +180,7 @@ CVAR_RANGE_FUNC_DECL(msgmidcolor, "5", "Color used for centered messages.",
 // ------------
 
 // Determines whether to draw the scores on intermission.
-CVAR(				wi_newintermission, "0", "Draw the scores on intermission",
+CVAR(				wi_newintermission, "0", "Use Odamex's intermission screen if there are 4 players or less on cooperative gamemodes.",
```
leximayfield, line 183: Is there a convenient place to slot this into the menu system?

leximayfield (changes requested):
Aside from my notes, I actually like the idea that @rakohus had.  Would it be possible to just print the name in either smallfont or digfont above or below each player?  I feel that would preserve some of the usefulness of the screen.

client/src/cl_cvarlist.cpp
Ch0wW: It's already there for years... It just was left for many releases.
image

Ch0wW:
Current status online looks like this.

image

leximayfield (approved):

PR #445, opened 2021-08-02
Create QoL for lobby maps 

Ch0wW:
This PR does the following:
- Create a `SetLobby` command that works exactly like `addmap` : ```setlobby MAP <wads>``` . It only has 1 entry, and will be overwritten it you use it again.
- If all players left on a map, the server goes back to the selected lobby map.
- Once a map is completed (ie intermission), the server goes back to the selected lobby map.

As a result, it will allow server creators to not add "addmap MAP00" between two maps, simplifying the configs.


leximayfield (approved):
I'm okay with this.  I have the feeling that there's a more elegant way to check for an empty string in that instance, but I'm assuming you already tried empty and it didn't work.

Issue #424, opened 2021-06-26
[FEATURE] In team-modes with >2 teams, open up spectating to the other 2 teams once yours is eliminated


rakohus:
**Is your feature request related to a problem? Please describe.**
This particularly applies to 3-team LMS. If your team is eliminated, there is nobody left to spectate. You are forced to stare at the scoreboard in a dead state.

**Describe the solution you'd like to see**
Open up the auto-spectator to looking at the other two remaining teams if your entire team is eliminated.

leximayfield:
I've pushed a fix for this - please give it a good looksee.

leximayfield:
Not only did this make it into 0.9.4 it caused a bug that was fixed in 0.9.5.  So I think we can call this addressed.

PR #449, opened 2021-08-07
Allow spying other teams if viewers team is dead

leximayfield:
Addresses issue #424.


rakohus:
Was tested by 3 people, including myself. Seemed to work fine and nobody noticed any issues during testing.

PR #452, opened 2021-08-09
Allow finer mouse sensitivity tweaks in the settings menu

Calinou:
This is useful for mice with particularly high DPI settings. Here, I need to use 0.1 to get a good sensitivity value on my Logitech G903 Lightspeed on Linux (flat acceleration profile).


leximayfield:
LGTM

PR #454, opened 2021-08-14
Rename the wi_intermission CVAR.

Ch0wW:
This PR only changes the CVAR so that Odamex's scoreboard is still the default one.


leximayfield (approved):
LGTM.  Thanks for the quick turnaround on this.

PR #460, opened 2021-09-03
Obituary updates

leximayfield:
A few updates for our obituaries:

- New icons.  Thanks Scuba Steve!
- Console player names are now highlighted specially.
- Add a few more means of death to accommodate the new icons, as well as changing some of the existing means.


rakohus:
Was tested weeks ago. Worked well.

HyperEye (approved):

PR #466, opened 2021-09-25
Fix colored lighting being too bright

leximayfield:
During development of a mapset, I noticed that our colored lighting seemed a little bright compared to the light level.  Doing a little digging, I figured out that we did indeed have a problem.

image

Compared with ZDoom 2.8.1

image

Below is a sample colormap comparison between a color multiply done in GIMP (top) vs what the colormap looked like in Odamex (bottom).  It's purple because I was using a different map to test.

image

The trouble turned out to be that our base colors were too bright due to a different algorithm being used.  There might be a slicker way of fixing this problem, but I decided to simply copy the standard colormap generation function and add a multiply on top of that.  The result...

image

### DIFF
--- a/client/src/v_palette.cpp
+++ b/client/src/v_palette.cpp
@@ -1002,7 +1002,8 @@ fargb_t V_HSVtoRGB(const fahsv_t &color)
 /****** Colored Lighting Stuffs (Sorry, 8-bit only) ******/
 
 // Builds NUMCOLORMAPS colormaps lit with the specified color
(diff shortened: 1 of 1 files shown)


leximayfield:
There's an issue with non-default gamma that needs addressing.

EDIT: To elaborate, there is a bug that has been there since before my changes where gamma is not applied to colored sectors until after you reload the map.  You can merge this PR if you want, but it would be nice to track down the bug.

PR #470, opened 2021-10-15
Add monster damage to scoreboard

leximayfield:
This PR change the kills/deaths counter on the coop scoreboard to Monster HP Damage/Deaths.  Just what it says on the tin and no more.

It has been tested both by me and @rakohus to make sure it functions, so it just needs some code review.


fuzztooth (approved):
If it fits...

PR #475, opened 2021-11-05
Implement WDLStats Version 6

bcahue:
WDLStats version 6 offers many changes over version 5. This release allows support for all multiplayer game modes, but would still need some more work to be truly useful in coop game modes.

Other changes include:

1. Players are given IDs (based off player ids) in the players table, and events refer to those players by id instead of name. Player join position is recorded, and player renaming is handled gracefully. A "Player" in this regard is persisted as a player who was given join permissions upon reset (i.e., no connect/queue waiting/anything) and stays in the game. Once they disconnect, their player id is freed for the next player to join. As such, player joins and disconnects reference the exact player in the player table.
2. A unique map fingerprint is created from map related lumps and recorded, so unmodified maps in compilation wads can be identified. This fingerprint is generated via the FarmHash algorithm by Google.
3. Accuracy is split into 4 events, and in each event, will detail a miss or each player who was shot by a weapon, and if its a spread event, how many pellets connected.
4. Coop, Deathmatch, Team DM, Attack/Defend CTF (with or without manual flag return) and Horde all with Lives and Rounds are supported. Monster game modes are missing many monster events in order to make it truly useful for coop/horde, but it will record player events. There are some notable hangups with coop as well, like Cyberdemon rockets counting as "null player" rockets, and sparse voodoo doll support.
5. WAD hashes (MD5) are recorded in the WDL logs.
6. Accuracy events now use MOD values instead of a homegrown value for each weapon.
7. POWERPICKUP event has been repurposed into PICKUPITEM, and covers just about every item that can be picked up by Doomguy.
8. Win types and winner IDs are printed in WDL logs, along with round number, attack/defend boolean, and number of lives (0 being unlimited).
9. The game catalogs item spawns on map load and puts them in a table to be referenced by the PICKUPITEM event. For Horde, these catalog Horde cache spawners and Horde Powerup spawners.
10. The game catalogs player spawns on map load and puts them in a table to be referenced by the SPAWNPLAYER event.
11. The game catalogs flag "sockets" and records them in a table for playback purposes.
12. Added JOINGAME and DISCONNECT events for users who drop out of a game or join mid-game.
13. Player renaming is handled gracefully by renaming the player in the player table.
14. Added PROJFIRE event to identify when a player has fired a projectile, and which kind.
15. Added PLAYERBEACON event to fire every 5 tics the location of the player. This is off by default and controlled by a new var sv_playerbeacons.


common/p_interaction.cpp
```diff
@@ -935,95 +940,123 @@ void P_GiveSpecial(player_t *player, AActor *special)
...
 
 	    case SPR_SHEL:
-            val = P_GiveAmmo(player, am_shell, 1);
+		    if (special->flags & MF_DROPPED)
+		    {
+			    val = P_GiveAmmo(player, am_shell, 1);
+		    }
+		    else
+		    {
+			    val = P_GiveAmmo(player, am_shell, 1);
```
leximayfield: Dupllicate code in the conditional?

common/m_wdlstats.cpp
```diff
@@ -259,69 +542,268 @@ static bool LogDamageEvent(
...
+
+	return false;
+}
+
+/**
+ * Log a hit shot by a player
+ *
+ * Looks for an accuracy log somewhere in the backlog, if there is none, it
+ * logs a message but continues.
+ *
```
leximayfield: Blank space?

common/p_interaction.cpp
```diff
@@ -2018,6 +2085,46 @@ void P_DamageMobj(AActor *target, AActor *inflictor, AActor *source, int damage,
...
+			M_LogActorWDLEvent(WDL_EVENT_ENVIROCARRIERDAMAGE, source, target,
+			                   actualdamage, armorDamage, mod, f);
+		else if (source != NULL && targethasflag)
+			M_LogActorWDLEvent(WDL_EVENT_CARRIERDAMAGE, source, target, actualdamage,
+			                   armorDamage, mod, f);
+		else
+			M_LogActorWDLEvent(WDL_EVENT_DAMAGE, source, target, actualdamage,
+			                   armorDamage, mod, 0);
+
+		if (mod == MOD_PISTOL || mod == MOD_CHAINGUN || mod == MOD_FIST ||
```
leximayfield: Please replace this with a switch/case.

common/p_interaction.cpp
```diff
@@ -2116,20 +2196,19 @@ void P_DamageMobj(AActor *target, AActor *inflictor, AActor *source, int damage,
 		P_KillMobj(source, target, inflictor, false);
 
 		// WDL damage events.
+		// todo: handle voodoo dolls here
```
leximayfield, line 2268: Online voodoo dolls do not take damage.

common/p_interaction.cpp
```diff
@@ -2116,20 +2196,19 @@ void P_DamageMobj(AActor *target, AActor *inflictor, AActor *source, int damage,
...
 		if (source == NULL && targethasflag)
```
leximayfield, line 2269: Please make sure that the use of braces in conditionals is consistent.  They should either all have them or none of them have them.

common/p_map.cpp
```diff
@@ -2774,7 +2775,13 @@ static BOOL PIT_DoomRadiusAttack(AActor* thing)
 		dist = 0;
 
 	if (dist >= bombdamage)
-		return true;	// out of range
+	{
+		if (bombsource && bombsource->player)
```
leximayfield, line 2779: If a condition block is spread across multiple lines, could you please add braces to them?

common/p_map.cpp
```diff
@@ -2871,6 +2878,13 @@ static BOOL PIT_ZDoomRadiusAttack(AActor* thing)
 		thing->momy = momy + (fixed_t)((thing->y - bombspot->y) * thrust);
 		thing->momz += (fixed_t)momz;
 	}
+	else
+	{
+		if (bombsource && bombsource->player)
```
leximayfield, line 2885: Please add braces to this if.

common/p_pspr.cpp
```diff
@@ -1143,6 +1167,10 @@ void A_BFGSpray(AActor* mo)
 	if (!mo->target)
 		return;
 
+	if (mo->target->player)
```
leximayfield, line 1170: Please add braces for this multi-line if block.

common/p_setup.cpp
```diff
@@ -1462,8 +1464,51 @@ void P_LoadBlockMap (int lump)
...
+	levellumps.insert(levellumps.end(), W_LumpLength(vertexes), *vertexbytes);
+	levellumps.insert(levellumps.end(), W_LumpLength(segs), *segsbytes);
+	levellumps.insert(levellumps.end(), W_LumpLength(ssectors), *ssectorsbytes);
+	levellumps.insert(levellumps.end(), W_LumpLength(sectors), *sectorsbytes);
+
+	length = W_LumpLength(things) + W_LumpLength(linedefs) + W_LumpLength(sidedefs) +
+	         W_LumpLength(vertexes) + W_LumpLength(segs) + W_LumpLength(ssectors) +
+	         W_LumpLength(sectors);
+
+	memcpy(::level.level_fingerprint, W_FARMHASH(levellumps.data(), length),
```
leximayfield: This is a memory leak as-written.  `W_FARMHASH` is using malloc to allocate memory and returning a pointer to that memory.  You are then treating that return value as a temporary and discarding the pointer afterwards.

I believe that the correct fix here is to fix W_FARMHASH proper.  See my comments on that.

common/w_wad.cpp
```diff
@@ -170,6 +172,51 @@ std::string W_MD5(std::string filename)
...
 
+/*
+ * @brief Creates a fingerprint for a map via FarmHash.
+ *
+ * However, it encodes the fingerprint into a 16-byte array to be read later.
+ *
+ * @param lumpdata byte array pointer to the lump that needs to be fingerprinted.
+ * @param size of the byte array pointer.
+ */
+byte* W_FARMHASH(const byte* lumpdata, int length)
```
leximayfield: This design is very error-prone, since you are allocating a byte buffer in the function contents (with `malloc` no less) but the name of the function does not indicate any allocation is going on, nor does the docstring.

What I would do instead is return the data in a struct containing one array member by value.  That allows you to pass back arrays without them decaying to pointers.  One day I'll make a OArray, but a hand-rolled struct will do for now.

Also, I prefer `W_FarmHash` since that's how the official hash name is capitalized - you might also want to add the size of the hash at the end in bits.

leximayfield: This design is very error-prone, since you are allocating a byte buffer in the function contents (with `malloc` no less) but the name of the function does not indicate any allocation is going on, nor does the docstring.

What I would do instead is return the data in a struct containing one array member by value.  That allows you to pass back arrays without them decaying to pointers.  One day I'll make a OArray, but a hand-rolled struct will do for now.

Also, I prefer `W_FarmHash` since that's how the official hash name is capitalized - you might also want to add the size of the hash at the end in bits.

leximayfield (changes requested):

common/p_interaction.cpp
bcahue: During testing, I observed on Sunlust.WAD MAP02 on a client/server setup the voodoo dolls would throw a "ENVIROKILL" event whenever they were damaged. I thought this was the mechanism that was doing so, but I didn't look into it very long.

common/p_setup.cpp
bcahue: Fixed this to be a bit more sane.

common/p_setup.cpp
```diff
@@ -1462,8 +1464,52 @@ void P_LoadBlockMap (int lump)
...
+* 
+* @param things - Things lump offset
+* @param linedefs - Linedefs lump offset
+* @param sidedefs - Sidedefs lump offset
+* @param vertexes - Vertexes lump offset
+* @param segs - Segs lump offset
+* @param ssectors - SSectors lump offset
+* @param sectors - Sectors lump offset
+*/
+void P_GenerateUniqueMapHash(int things, int linedefs, int sidedefs, int vertexes,
```
leximayfield: You don't need this many parameters.  The order of lumps for a given map is fixed, all that really matters is if it has a behavior lump or not.  I would recommend passing in just the base lump id and then behavior true/false.

server/src/sv_game.cpp
```diff
@@ -39,6 +39,7 @@
 #include "g_game.h"
 #include "sv_main.h"
 #include "g_spawninv.h"
+#include "m_wdlstats.h"
```
leximayfield: What is using this header?

leximayfield (changes requested):
Just one or two small things.

common/p_setup.cpp
bcahue: Pared this down to a reasonable amount of params.

server/src/sv_game.cpp
bcahue: Removed unnecessary header.

bcahue:
Changes have been implemented.

leximayfield (approved):
LGTM

leximayfield:
Merged.  CI was failing the build, but the issues were in my code, not the PR's.

PR #476, opened 2021-11-14
Get cl_autorecord_horde to function

leximayfield:
Also add it to the menu.


rakohus:
Confirmed that it works functionally.

PR #480, opened 2021-11-20
Fix bug with potential retrieval of non-existing spawn.

bcahue:
This fixes a bug with removing a player spawn if it's a voodoo doll spawn. It was removing too late in the execution however, and could potentially point to a spawn that isn't the one that was supposed to be removed -- or worse, crash.


leximayfield (approved):
LGTM

PR #483, opened 2021-11-27
[FEATURE] MBF21/DEHEXTRA Implementation

bcahue:
What is MBF21?
https://github.com/kraflab/mbf21/blob/master/docs/developer_spec.md

This builds on the excellent work by Ch0wW to complete the DEHEXTRA implementation, and finish 95% of the MBF21 implementation. Unfortunately, the linedef flags cannot be completed as Odamex currently uses Hexen's SPAC system for doors and switches even in maps in DOOM format, so those bits are unavailable and get overwritten by the SPAC LUT translation. However, another pull request is in the works to fix this for Doom format levels and handle them separately than Hexen levels, which should pave the way to do this the right way.

I've tested the current flag changes to DOOM monsters (like Baron, and Cyberdemon) and ensured their Boss activities resumed even though it wasn't hard coded anymore (e.g. E2M8 cyberdemon death exit).

I've also tested using a plethora of MBF21 wads from Doomworld online, both public and unreleased. The public WADs that were tested include:
AD_MORTEM
Experimental Mixed Bag Salad
Heretic Invades Doom

https://www.doomworld.com/forum/topic/125786-new-editing-standards-resource-megathread/

Attached is my test wad for MBF21 created by the wonderfully helpful Xenaero. The map is MAP15, and should have every sector action, and the generalized linedef scrolling type working. Please note that BlockPlayers (BFG is reachable) and BlockLandMonsters (pinky can run around the map) do not work.

[Download oda-mbf21.wad](https://www.dropbox.com/s/99j2ukf2mochso6/oda-mbf21.wad?dl=0)


client/src/r_things.cpp
```diff
@@ -442,7 +442,7 @@ void R_DrawVisSprite (vissprite_t *vis, int x1, int x2)
 
 	dcol.textureheight = 256 << FRACBITS;
 
-	if (vis->mobjflags & MF_SPECTATOR)
+	if (vis->mobjflags2 & MFO_SPECTATOR)
```
leximayfield: I know this is a vissprite, but I'd much rather this be matched up 1:1 with their source bitfield names, even if that means you need to create mobjflags1-3 and then oflags.  Or, if you don't want to create an entire field just for one flag, maybe just do a "bool".

common/actor.h
```diff
@@ -163,106 +163,129 @@ class PlayerBitField
 //
 typedef enum
```
leximayfield: If you're going to make all these changes anyway, I'd recommend changing `typedef enum` to a plain old `enum mobjflag_t`.  The typedef is a holdover from C and is only used to allow you to avoid prefacing each use of the type with the word `struct`.  C++ gets rid of that silliness, so the typedef is useless.

common/d_dehacked.cpp
```diff
@@ -699,36 +776,31 @@ static int GetLine (void)
...
-		short bit;
-		short whichflags;
-		const char *name;
-	} bitnames[] = {
+	struct flagsystem_t
+	{
+		short Bit;
+		short WhichFlags;
+		const char *Name;
+	} ;
```
leximayfield: Extra space here.

common/d_dehacked.cpp
```diff
@@ -800,109 +878,473 @@ static int PatchThing (int thingy)
...
+	int *ednum, dummyed;
+	bool hadHeight = false;
+	bool gibhealth = false;
+
+	info = &dummy;
+	ednum = &dummyed;
 
 	thingNum--;
-	if (thingNum < NUMMOBJTYPES) {
+	if (thingNum < 0 || thingNum > NUMMOBJTYPES)
```
leximayfield: Seems like `thingNum > NUMMOBJTYPES` should be `thingNum >= NUMMOBJTYPES`.

common/d_dehacked.cpp
```diff
@@ -800,109 +878,473 @@ static int PatchThing (int thingy)
...
+	{
+		DPrintf("Thing %" PRIuSIZE " out of range.\n", thingNum);
+	}
+	else {
```
leximayfield: For new or changed code, please prefer Allman style, which MSVC's clang-format support should help you with.

common/d_dehacked.cpp
```diff
@@ -800,109 +878,473 @@ static int PatchThing (int thingy)
...
+			if (stricmp(Line1 + linelen - 6, " frame") == 0)
+			{
+				statenum_t state = (statenum_t)val;
+
+				if (!strnicmp(Line1, "Initial", 7))
+					info->spawnstate = state;
+				else if (!strnicmp(Line1, "First moving", 12))
+					info->seestate = state;
+				else if (!strnicmp(Line1, "Injury", 6))
+				{
```
leximayfield: Please make this brace pair consistent - since these all are one-liners you can choose if they're all braced or all braceless.

common/d_dehacked.cpp
```diff
@@ -800,109 +878,473 @@ static int PatchThing (int thingy)
...
-				info->painsound = SoundMap[sndmap];
-			else if (!stricmp (Line1, "Death sound"))
-				info->deathsound = SoundMap[sndmap];
-			else if (!stricmp (Line1, "Action sound"))
-				info->activesound = SoundMap[sndmap];
-			else if (!stricmp (Line1, "Bits"))
+		size_t val = atoi(Line2);
+		int linelen = strlen(Line1);
+
+		if (linelen > 6)
```
leximayfield: This applies to this entire `if` block, but you don't have to check for line length for simple string comparisons.  `stricmp` won't run off of the end of a null terminated string.

common/d_dehacked.cpp
```diff
@@ -800,109 +878,473 @@ static int PatchThing (int thingy)
...
 				{
-					size_t iy;
+					info->translucency = val;
+					hadTranslucency = true;
 
-					if (IsNum (strval))
+				}
+				else if (stricmp(Line1, "Dropped item") == 0)
+				{
+					// UNSUPPORTED FOR NOW...
```
leximayfield: Probably a wise idea to print a warning in the console if something isn't supported and we know about it.

common/d_dehacked.cpp
```diff
@@ -800,109 +878,473 @@ static int PatchThing (int thingy)
...
+		{
+			if (!stricmp(Line1, "Bits"))
 				{
-					info->flags = value;
-					// Bit flags are no longer used to specify translucency.
-					// This is just a temporary hack.
-					if (info->flags & 0x60000000)
-						info->translucency = (info->flags & 0x60000000) >> 15;
+					int value[4] = {0, 0, 0};
+				    bool vchanged[4] = {false, false, false};
```
leximayfield: Dodgy spacing here.

common/d_dehacked.cpp
```diff
@@ -981,9 +1439,70 @@ static int PatchFrame (int frameNum)
...
+						}
+					}
+					if (vchanged)
+					{
+						info->flags = value; // Weapon Flags
+					}
+				}
+			}
+			else
+				PrintUnknown(Line1, "Frame", frameNum);
```
leximayfield, line 1581: Please either brace both if's branches or neither.

common/p_enemy.cpp
```diff
@@ -1660,23 +1804,27 @@ void A_Mushroom(AActor* actor)
 
 	A_Explode(actor); // make normal explosion
 
-	for (int i = -n; i <= n; i += 8) // launch mushroom cloud
+	if (serverside)
```
leximayfield: I prefer an "early exit" style where instead of encasing most of the function in a big block, just return early if `!serverside`.  Also, keep in mind that in single-player, serverside AND clientside are true.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+	type = actor->state->args[0] - 1;
+	angle = actor->state->args[1];
+	ofs_x = actor->state->args[2];
+	ofs_y = actor->state->args[3];
+	ofs_z = actor->state->args[4];
+	vel_x = actor->state->args[5];
+	vel_y = actor->state->args[6];
+	vel_z = actor->state->args[7];
+
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+		return;
+
+	type = actor->state->args[0] - 1;
+	angle = actor->state->args[1];
+	pitch = actor->state->args[2];
+	spawnofs_xy = actor->state->args[3];
+	spawnofs_z = actor->state->args[4];
+
+	A_FaceTarget(actor);
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+{
+	angle_t threshold, maxturnangle;
+
+	if (!actor)
+		return;
+
+	threshold = FixedToAngle(actor->state->args[0]);
+	maxturnangle = FixedToAngle(actor->state->args[1]);
+
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+	dist = (actor->state->args[1]);
+
+	AActor* tracer = P_RoughTargetSearch(actor, fov, dist);
+
+	if (!tracer || tracer->health <= 0)
+		return;
+
+	actor->tracer = tracer->ptr();
+
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+// Clear current tracer (seek target).
+//
+void A_ClearTracer(AActor* actor)
+{
+	if (!actor)
+		return;
+
+	actor->tracer = AActor::AActorPtr();
+
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+	fov = FixedToAngle(actor->state->args[1]);
+
+	// Check FOV first since it's faster
+	if (fov > 0 && !P_CheckFov(actor, actor->tracer, fov))
+		return;
+
+	if (actor->tracer->health <= 0)
+		return;
+
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -1734,6 +1882,594 @@ void A_BetaSkullAttack(AActor* actor)
...
+	if (!actor || !actor->tracer)
+		return;
+
+	state = actor->state->args[0];
+	distance = actor->state->args[1];
+
+	if (actor->tracer->health <= 0)
+		return;
+
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_enemy.cpp
```diff
@@ -2378,17 +3115,21 @@ void A_Gibify(AActor *mo) // denis - squash thing
 //
 void A_Spawn(AActor* mo)
 {
-	// Partial integration of A_Spawn.
-	// ToDo: Currently missing MBF's MF_FRIEND flag support!
-	if (mo->state->misc1)
+	if (serverside)
```
leximayfield: See commentary above about `serverside` and returning early.

common/p_map.cpp
```diff
@@ -2757,14 +2802,29 @@ CVAR_FUNC_IMPL(sv_splashfactor)
...
+		return true;
+
+	// MBF21
+	if (P_SplashImmune(thing, bombspot))
 		return true;
 
 	// Boss spider and cyborg
 	// take no damage from concussion.
-	if (thing->type == MT_CYBORG || thing->type == MT_SPIDER)
+	if (bombspot->flags & MF_BOUNCES ?
```
leximayfield: I would much rather this ternary nonsense be spelled out with if statements.

common/p_map.cpp
```diff
@@ -2803,12 +2863,19 @@ static BOOL PIT_DoomRadiusAttack(AActor* thing)
...
+		return true;
+
+	// MBF21
+	if (P_SplashImmune(thing, bombspot))
 		return true;
 
 	// Boss spider and cyborg
 	// take no damage from concussion.
-	if (thing->flags2 & MF2_BOSS)
+	if (bombspot->flags & MF_BOUNCES
```
leximayfield: I would much rather this ternary nonsense be spelled out with if statements.

common/p_pspr.cpp
```diff
@@ -27,6 +27,8 @@
 
 #include <stdlib.h>
 
+#include "d_dehacked.h"
+#include "doomdef.h"
```
leximayfield: With `odamex.h` you should not need to include this by hand.

common/p_spec.cpp
```diff
@@ -82,6 +82,97 @@ std::list<movingsector_t>::iterator P_FindMovingSector(sector_t *sector)
...
+ *
+ * Returns nonzero if the object is under damage based on
+ * their current position.
+ */
+int P_IsUnderDamage(AActor* actor)
+{
+	const struct msecnode_s* seclist;
+	const DCeiling* cr; // Crushing ceiling
+	int dir = 0;
+	for (seclist = actor->touching_sectorlist; seclist; seclist = seclist->m_tnext)
```
leximayfield, line 108: Please brace this block, not a fan of nested braceless blocks.

common/p_spec.cpp
```diff
@@ -82,6 +82,97 @@ std::list<movingsector_t>::iterator P_FindMovingSector(sector_t *sector)
...
+ * @param player Player actor to test same teamness
+ * @returns A boolean to determine if the player is on the shooter's team.
+ */
+bool P_IsTeamMate(AActor* actor, AActor* player)
+{
+	if (sv_gametype == GM_DM)
+	{
+		return false;
+	}
+	else if (sv_gametype == (GM_HORDE || GM_COOP))
```
leximayfield: You can use `G_IsCoopGame` here instead.

common/p_spec.cpp
```diff
@@ -82,6 +82,97 @@ std::list<movingsector_t>::iterator P_FindMovingSector(sector_t *sector)
...
+/*
+ *
+ * P_IsTeamMate
+ * @brief Helper function to determine if a particular thing is a teammate.
+ *
+ * @param actor Source player actor
+ * @param player Player actor to test same teamness
+ * @returns A boolean to determine if the player is on the shooter's team.
+ */
+bool P_IsTeamMate(AActor* actor, AActor* player)
```
leximayfield: Seems like duplicated functionality with `P_AreTeammates`.  See if you can consolidate this down to one function, whichever one makes more sense.

common/p_spec.cpp
```diff
@@ -2146,7 +2326,7 @@ void DScroller::RunThink ()
 	if (m_Control != -1)
 	{	// compute scroll amounts based on a sector's height changes
 		sector_t *sector = &sectors[m_Control];
-		fixed_t height = P_FloorHeight(sector->soundorg[0], sector->soundorg[1], sector);
+		fixed_t height = sector->ceilingheight;
```
leximayfield: This seems incorrect.  Why are we getting rid of `P_FloorHeight` here?  Do we need the raw value here for some reason?

common/p_spec.cpp
```diff
@@ -2246,13 +2426,14 @@ DScroller::DScroller (EScrollType type, fixed_t dx, fixed_t dy,
 	if ((m_Control = control) != -1)
 	{
 		sector_t *sector = &sectors[control];
-		fixed_t height = P_FloorHeight(sector->soundorg[0], sector->soundorg[1], sector);
+		fixed_t height = sector->ceilingheight;
```
leximayfield: See above - why get rid of `P_FloorHeight` - aside from the fact that it's floor height and not ceiling height?

common/p_spec.cpp
```diff
@@ -2280,7 +2461,7 @@ DScroller::DScroller (fixed_t dx, fixed_t dy, const line_t *l,
...
```
leximayfield: See above.

server/src/sv_main.cpp
```diff
@@ -665,7 +665,7 @@ void SV_Sound (AActor *mo, byte channel, const char *name, byte attenuation)
 
 	sfx_id = S_FindSound (name);
 
-	if (sfx_id > 255 || sfx_id < 0)
+	if (sfx_id > numsfx || sfx_id < 0)
```
leximayfield: Is this the correct check here....shouldn't it be `sfx_id >= numsfx`?

server/src/sv_main.cpp
```diff
@@ -694,7 +694,7 @@ void SV_Sound(player_t& pl, AActor* mo, const byte channel, const char* name,
...
```
leximayfield: See above.

server/src/sv_main.cpp
```diff
@@ -728,7 +728,7 @@ void UV_SoundAvoidPlayer (AActor *mo, byte channel, const char *name, byte atten
...
```
leximayfield: See above.

server/src/sv_main.cpp
```diff
@@ -758,7 +758,7 @@ void SV_SoundTeam (byte channel, const char* name, byte attenuation, int team)
 
 	sfx_id = S_FindSound( name );
 
-	if ( sfx_id > 255 || sfx_id < 0 )
+	if (sfx_id > numsfx || sfx_id < 0)
```
leximayfield: See above.

server/src/sv_main.cpp
```diff
@@ -783,7 +783,7 @@ void SV_Sound (fixed_t x, fixed_t y, byte channel, const char *name, byte attenu
 
 	sfx_id = S_FindSound (name);
 
-	if (sfx_id > 255 || sfx_id < 0)
+	if (sfx_id > numsfx || sfx_id < 0)
```
leximayfield: See above.

leximayfield (changes requested):
Got some changes and questions.

common/d_dehacked.cpp
bcahue: Didn't test with anything that dropped anything custom, will look into implementing properly.

common/p_spec.cpp
bcahue: These (scrollers that use control sectors) never worked correctly in Odamex. The Boom scrollers that use control sectors depend on _ceiling_ height changes, and the initial height is recorded during map load and compared then on. Consider the following, line special 215 (Scroll Floor Accelerates When Sector Changes Height). This is its behavior on Odamex 0.9.5:
https://youtu.be/Qzf6i4oWI2M
Behavior on the commit that I haven't pushed yet, but didn't modify this code:
https://youtu.be/GnAreXbAfXM
And behavior for DSDA-Doom:
https://youtu.be/qqnxemrrb9A
It looks odd but we had the spec wrong from the beginning, and I believe mine is correct to the best of my knowledge.

EDIT: Looking at this again, it appears I did not get this correct, and did not read the spec correctly. Ceiling height is only half the solution, floor height is needed as well, and they're just added together. Here's some new code that I've pushed, and to take a look we will do the same thing based on that same demonstration. Here's Odamex 0.9.5:
https://youtu.be/32SfyzzB17c
Current pull request version with sector height scroller changes:
https://youtu.be/Sp-elhpSr_s
DSDA-Doom
https://youtu.be/90YV9KFOf7Q

common/d_dehacked.cpp
bcahue: Implemented and tested to ensure current monsters drop items correctly.

leximayfield (approved):
LGTM

Issue #484, opened 2021-11-28
[FEATURE] Projectile passthrough flag for allies


Xenaero:
**Is your feature request related to a problem? Please describe.**
Only a problem if there are a large amount of players jammed into a coop map or mode.

**Describe the solution you'd like to see**
A svar that allows for projectiles to clip through players so direct impacts aren't possible.

**Describe alternatives you've considered**
Not adding an option for server owners to enable for large crowded servers might be negatively impactful to the player experience in some maps and mods where rockets, for example, are prevalent.

**Additional context**
None, I want people to be able to jump in and play and enjoy even in servers slam full of monsters and players. Accessibility options help with this.

bcahue:
This can be added to the same functionality that handles sv_unblockplayers for projectiles as well as players trying to pass thru one another. I will submit a pull request for this to add teams checking for sv_unblockplayers projectiles and explode accordingly.

leximayfield:
Merged a feature for this.

PR #486, opened 2021-11-29
[FEATURE] Allow projectiles to pass thru teammates when sv_unblockplayers is on

bcahue:
(This is a pull request to resolve the feature request #484)

The following change allows projectiles to passthru friendly players (all players in coop/horde, no players in DM, and teammates only in team game modes), which should help the spam situation in crowded horde servers. Please note I couldn't use P_AreTeammates here, as clientside every player has the same ID, so the rocket would desync as it would wait for an explosion that would never arrive from the server.


leximayfield (approved):
LGTM!

PR #491, opened 2021-12-14
Implement MAPINFO lumps for FreeDoom Phase 1 and 2

bcahue:
This pull request fixes a broken FreeDoom Phase 1 and 2 implementation, broken sometime after the UMAPINFO implementation. In adding these MAPINFO lumps, Odamex can once again play FreeDoom Phase 1 and 2.


leximayfield:
@bcahue I did a little research and found out that Freedoom comes with DeHackEd lumps.  I actually committed an alternative fix in 40ee5f1 that loads the D1/D2 mapinfo for Freedoom and then lets the DeHackEd lump change things around.  Do you see an issue with that strategy?

bcahue:
This strategy will indeed work since these lumps are based off D1/D2 anyway and just change the episode number. Closing this as this particular issue is solved with commit 40ee5f1.

PR #495, opened 2021-12-28
[BUGFIX] Fix several serverside bugs for MBF21/DEHEXTRA

bcahue:
Found when testing AD MORTEM, weapon sounds play locally but not for other players. This fixes this behavior and handles it like current weapon sounds.

### DIFF
--- a/common/p_enemy.cpp
+++ b/common/p_enemy.cpp
@@ -3148,12 +3148,12 @@ void A_PlaySound(AActor* mo)
 		sndmap = 0;
 	}
 
-	S_Sound(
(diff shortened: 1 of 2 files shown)


bcahue:
I also added server playing provisions to A_PlaySound, as AD_MORTEM.wad has interactive statues that change state and play sounds, but these will only play for the activating player and no one else.

bcahue:
Also added a commit to allow state changes made by JumpIf* codepointers to broadcast to clients, so the activator isn't the only one who gets the state change.

leximayfield (approved):
LGTM

PR #497, opened 2021-12-30
Add "baselines" to netcode

leximayfield:
This branch adds the concept of a "baseline" to the netcode.  A "baseline" snapshot of a mobj is taken on spawn, and then every time the mobj is updated thereafter a diff between the baseline and current is generated and then sent to the client.  This is much more reliable than the old system and significantly reduces the amount of "floaty" monster collision that was encountered.  It also, ironically, takes less bandwidth than the previous approach.

This branch also contains a few other PR's that were made against `protobuf-baseline` instead of `protobreak`, mostly relating to MBF21.


leximayfield:
We've been testing this branch quite a bit.  The other commits in here have been reviewed in other PR's and would just be confusing to go through twice.

PR #500, opened 2022-01-01
Add a FLTK-driven boot window

leximayfield:
This branch adds a boot window to Odamex.  The aim is to significantly ease the story surrounding setting up and selecting IWADs for first-time users.

odamex_18feYEuJ5F

- It catalogs all IWAD files in your `waddirs` and allows you to select which IWAD to use.
- You can also add and remove directories from your `waddirs` if you don't see any IWADs....or even if you do.
- If you want to turn this feature off, "Don't show this again" will do what you expect.  If you want to show it again, there's a new options menu option that will re-enable it.
- A number of command line params like `-file` and `-warp` short-circuit past it.


rakohus:
Tests fine

leximayfield:
The CentOS workflow is having an issue that has nothing to do with our code, so I'll go ahead and merge this and hope the workflow turns green this time.

Issue #499, opened 2021-12-31
[BUG] Crash when restarting AD_MORTEM.WAD MAP05


bcahue:
**Describe the bug**
"szp pointer was NULL" serverside error when restarting AD_MORTEM.WAD MAP05.

**Build that the bug occurred in**
Branch `protobreak` commit 78d074b

**To Reproduce**
Configure a server for survival, with AD_MORTEM.wad. Skip to MAP05, run off the edge of the map to die and trigger a restart.

**Expected behavior**
Map restarts and does not crash.

**Screenshots, NetDemos, & Crash Dumps**
N/A bug is reproducible every time.

**Additional context**
When debugging the server and triggering this crash, I discovered that this map has 3 BOOM-style thing Pushers. The crash occurred in `void DPusher::Serialize (FArchive &arc)` and crashed because the serializer was trying to read a null actor pointer...which is created by the DPusher initialization function.

PR #501, opened 2022-01-02
Serialize and unserliaze pusher sources

leximayfield:
A fix for #499 .


leximayfield:
Confirmed fixed by @bcahue.

Issue #502, opened 2022-01-03
[BUG] MBF21 A_RadiusDamage 'radius' arg is ignored


XaserAcheron:
**Describe the bug**
MBF21's A_RadiusDamage function doesn't currently use the `radius` arg (Arg 1) -- it's still using the `damage` arg (Arg 0) to determine the size of the explosion.

The issue isn't within `A_RadiusDamage` itself, but in `PIT_DoomRadiusAttack` in p_map.cpp -- right now it's still using vanilla logic and needs a couple of tweaks to make use of the new radius fields. Offhand, I think the necessary changes are:
- [This line](https://github.com/odamex/odamex/blob/protobreak/common/p_map.cpp#L2883) needs to check `bombdistance` instead of `bombdamage`
- The damage passed into P_DamageMobj on [this line](https://github.com/odamex/odamex/blob/protobreak/common/p_map.cpp#L2897) (i.e. the `(bombdamage - dist)` part) needs to be replaced with a formula that factors in both damage & radius -- feel free to yoink the formula from [here](https://github.com/kraflab/dsda-doom/blob/master/prboom2/src/p_map.c#L2587), since GPL is cool like that. :P

I'd make a PR myself but it may be a while before I can get a dev environment set up, and it's best not to wait on me for most things. :P

**Build that the bug occurred in**
`v10.0.0 (protobreak g3c2a-1)` -- whatever that number means. It's an alpha build from the `protobreak` branch that Ralphis DM'd me yesterday. :P

**To Reproduce**
A quick way to repro this is to load up [Vesper](https://www.doomworld.com/forum/topic/123800) and fire a Nox Grenade (slot 5 weapon) at the wall, then step into the poison gas cloud. It's intended to have a rather large radius, but in current Oda you have to get really close to the center of the cloud to begin to start taking damage; with the fix in place, you'll start to suffocate much further away. :P

**Expected behavior**
A taco is dispensed when I press the taco button.

**Additional context**
Moose are good animals.

bcahue:
Opened pull request #503 to resolve this bug.

rakohus:
Corrected behavior and will be present in Odamex 10.

PR #503, opened 2022-01-04
[BUGFIX] Propose fix for bug #502

bcahue:
This pull request solves the issue with xa-vesper.wad in bug #502.


leximayfield (approved):
LGTM

PR #505, opened 2022-01-06
Remove sc_man

DeathEgg:
Removes the last vestiges of sc_man and replaces them with OScanner.
Also adds an OScanner function that lets you know if the last scan found a new line.


leximayfield (changes requested):
LGTM

leximayfield:
Hit the wrong button in my review, it's merged.

PR #508, opened 2022-01-12
split common code between r_things.cpp/.h versions in client and server builds into separate common file

DeathEgg:


leximayfield (approved):
LGTM

PR #523, opened 2022-01-24
fix bug 518

DeathEgg:


rakohus:
Tested and confirmed to be working. Just needs code review

common/p_enemy.cpp
```diff
@@ -2787,12 +2787,14 @@ void A_BossDeath(AActor *actor)
...
-	else if (level.flags & LEVEL_CYBORGSPECIAL && actor->flags3 & MF3_E4M6BOSS)
+	else if (level.flags & LEVEL_CYBORGSPECIAL && actor->flags3 & MF3_E4M6BOSS &&
+	         level.mapname == "E4M6")
 	{
 		EV_DoDoor(DDoor::doorOpen, NULL, NULL, 666, SPEED(64), 0, NoKey);
 		return;
 	}
-	else if (level.flags & LEVEL_SPIDERSPECIAL && actor->flags3 & MF3_E4M8BOSS)
+	else if (level.flags & LEVEL_SPIDERSPECIAL && actor->flags3 & MF3_E4M8BOSS &&
+	         level.mapname == "E4M8")
```
leximayfield: I'm not sure the right play is hardcoding the map name.  Is there no way to do this otherwise?

leximayfield (changes requested):

common/p_enemy.cpp
leximayfield: Nevermind, I see what's going on.  How big of a pain in the neck would it be to add support for the following MAPINFO tokens?

- SpecialAction_ExitLevel
- SpecialAction_OpenDoor
- SpecialAction_LowerFloor
- SpecialAction_KillMonsters 

Seems like it just be a couple of one-note tokens that set a flag.

rakohus:
Tested on newest CI build.

D2 Map07 broken. After mancs are killed it drops the walls as lifts instead of a floor lower special. After arachs are killed, no floor stair raise

Could not load into doom.wad

bcahue:
I took a look at this PR and played DOOM2 MAP07 and found it to be working. However, loading into DOOM.WAD caused an error at MIType_SpecialAction_OpenDoor() in g_mapinfo.cpp on line 1096.

bcahue:
I did however see the same DOOM2 MAP07 behavior with the CICD artifact. Not sure what's happening there.

rakohus:
Tested doom.wad e1m8, e2m8, e3m8, e4m6, and e4m8. Tested doom2.wad map07 and map30. Everything appears to be working fine now!

leximayfield (approved):
LGTM.  Thanks for the cleanup.

PR #526, opened 2022-01-26
Add --version parameter to client and server

leximayfield:
Client and server get a new `--version` parameter.  On everything except the Win32 client, this version prints to stdout.  On the Win32 client, it outputs to a file in the CWD.


rakohus:
Per conversation between devs, seems fine to bring in to protobreak for Oda 10.

Issue #529, opened 2022-01-28
[BUG] ACS Broken in Duel2020OA.wad. Breaks/crashes server.


rakohus:
**Describe the bug**
Duel2020OA.wad. First time on the map, all the scripts run fine and you can teleport to another map by touching the map photo. Once it returns to map00, the scripts are broken. The script that makes the sky texture on the random map selection scroll doesn't run so you can see it immediately. Touching a map photo then says `P_StartScript: Unknown script 2`. If you then reload to map00 from console on odasrv, it bombs out entirely. 

**Build that the bug occurred in**
Oda 10 Dev Builds

**To Reproduce**
Load duel2020oa.wad. See description above.

**Expected behavior**
Touch map. Go to map. Return to map00. All scripts still working. Repeat.

bcahue:
I looked into this bug a bit. I reverted back to pre-MBF21 commits, to see if the MBF21 or the MapFormat implementation caused this to fail, commit ae301b53, and this behavior was still exhibited. So this bug has existed for a while. On stable, it runs the same lump with the same code, but on protobreak latest, it runs the same lump on the same code but has a different result. It's kind of hairy to debug since it's playing with bytes inside of pointers.

rakohus:
Resolved in #530. Thanks @bcahue!

PR #530, opened 2022-02-01
[BUGFIX] Fix ACS scripts bugged after initial load

bcahue:
This commit fixes the BEHAVIOR lump being changed after being processed for level.behavior in the FarmHash calculations. Since they don't really affect the structure of the map, I've taken it out. Now, ACS scripts will function correctly after level changes.


bcahue:
Fixes #529

rakohus:
Tested on local server. Confirmed that it works, behaviorally.

leximayfield (approved):

Issue #536, opened 2022-02-04
[BUG] Damage static inits broken in Odamex 10


rakohus:
**Describe the bug**
Gravity (line special 333) / Damage (335) static inits broken during map format re-do

**Build that the bug occurred in**
Odamex 10

**Screenshots, NetDemos, & Crash Dumps**
[odaedit_220204.zip](https://github.com/odamex/odamex/files/8006542/odaedit_220204.zip)

**Additional context**
First room has both specials for testing. Jumping will show gravity, which is reduced. Lava should kill immediately.

rakohus:
PR #537 mostly corrected this, but damage calc is wrong. Tested against Odamex 0.9.5, ZDaemon, and modern (g)zdoom

PR #537, opened 2022-02-05
[BUGFIX] Add Odamex Static Inits back into Doom format

bcahue:
This change adds the Damage and Gravity Odamex Static Inits back into Doom format in Odamex. Resolves bug #536.


leximayfield (approved):
LGTM

PR #541, opened 2022-02-07
Odaraven update

DeathEgg:
Took a lot from the old Odaraven branch and made it work with the current version of Odamex. It is not nearly complete, but I wanted to push what's done so far in case other want to contribute to it. All testing has been done solely with Client and not Server builds. As follows is an incomplete list of things that currently are in various stages of progress:

* Title screen is implemented, main menu is mostly functional (Info, Save/Load are incomplete. Quit message lacks Yes/No options. Player Setup will instantly crash. Haven't fully explored/compared Options menu yet)
* Levels load, have proper music, etc. that's covered by MAPINFO.
* Automap uses the correct colors and has the AUTOPAGE background. AUTOPAGE does not scroll, and the larger the screen is the more invisible lines are, which could be solved by either tiling the AUTOPAGE at a smaller size and/or making the lines thicker. Could also alias the lines like in the original. Text is also covered by the status bar sometimes. Also need to consider user-defined automap colors for Heretic.
* Status bar is functional. Widgets work, and chain works too. The tops of the heads do not properly display on screen unless the screen is at the minimum size. Ammo doesn't appear since no gameplay logic is coded. Inventory does not exist.
* All Heretic-exclusive linedef types except for those relating to the Heretic keys are implemented.
* Minor Heretic LANGUAGE additions so the menus work right.
* Probably some other minor stuff.

Things not implemented (a shortlist):

* Rating warning on title screen.
* Correct skill level functionality in Heretic.
* Proper Heretic keys/key doors.
* Any player weapons, including Tome of Power variants.
* Any kind of states for monsters/decorations.
* Any Heretic pickups or powerups.
* Really just any gameplay code for Heretic at all.
* Adapting the Odamex HUD for Heretic.
* Heretic variants of other Odamex graphics (Best solution may be PNG support and converting existing graphics to the palette?)
* Heretic-exclusive sector types.
* Intermission screens.
* Endgame screens.
* Sound effects/SNDINFO.
* Probably a bunch of other things I'm forgetting.

There's also at least two known issues this branch caused to Doom that are probably easy to fix:

* Episode select screen is weirdly off-center by a huge margin.
* If you load Heretic, play a level, and then load a Doom IWAD from the console, the background of the automap is still Heretics.

As it stands this isn't really in a shape where I'd recommend merging into protobreak any time soon, I've been careful so far but who knows if I accidentally messed up something in the Doom/Doom II code to make it behave slightly different, let alone the fact it's still easy to crash the game.


leximayfield:
Sure, this is better than what was there before.

PR #543, opened 2022-02-09
Codify auto-SR50 alias into an engine feature

leximayfield:
This PR essentially codifies the popular SR50 alias that seems to be passed around on forums and wikis.  Using `+strife` now gives you the extra sidemove that the alias would extract out of `+left` and `+right`.  You can disable this from the mouse options menu if you're not a fan.

There's also a speedometer added.  Currently not exposed in the menus because it's a quick implementation that probably has issues when f12ing someone online, but was good enough to test the SR50 implementation.


leximayfield:
This got testing from Ralphis, so we're good to go.

PR #544, opened 2022-02-09
Add support for MAPINFO skill definitions

DeathEgg:
Adds support for the ability to define custom skills and serializes the default IWAD skill definitions. The following ZDoom MAPINFO skill definition tokens are implemented:

* AmmoFactor
* DoubleAmmoFactor
* DamageFactor 
* FastMonsters
* DisableCheats
* EasyBossBrain
* RespawnTime 
* SpawnFilter (modified so baby = easy and hard = nightmare, since Odamex doesn't support 5 difficulty mode spawn filters)
* InstantReaction
* PicName 
* MustConfirm 
* Key 

skill_t in doomdef.h is essentially deprecated, but some horde mode code uses it still. That code should probably be modified to work with the new SkillInfos system instead.


DeathEgg:
I've since added HealthFactor and ArmorFactor as well as fixed previous MAPINFO parser errors I missed.

Issue #549, opened 2022-02-13
[BUG] Linedef Action 9 (Raise Donut) broken with Mapformat Changes


rakohus:
**Describe the bug**
E1M2 donut sector doesn't activate.

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
Go to E1M2 chainsaw

**Screenshots, NetDemos, & Crash Dumps**
![image](https://user-images.githubusercontent.com/22926877/153742941-20a52450-399d-4afa-b1d5-a5a1fdb5a2f8.png)

rakohus:
Fixed in #552

Issue #548, opened 2022-02-13
[BUG] Colormaps broken with new Mapformat Changes


rakohus:
**Describe the bug**
Line 242, lower tex doesn't translate colormaps into other sectors.

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
BOOMEDIT.wad. Pool inside of the first door with tag 35. 

**Expected behavior**
Colormap under the water.

**Screenshots, NetDemos, & Crash Dumps**
![image](https://user-images.githubusercontent.com/22926877/153742901-0240a277-651a-4441-8245-b1fd53010ec4.png)

rakohus:
Fixed in #552

PR #552, opened 2022-02-13
[BUGFIX] Fix multiple bugs with MapFormat

bcahue:
The following bugs on the tracker are fixed as well as multiple errors with boomedit.wad:

#548, #549


leximayfield (approved):
LGTM

leximayfield (approved):

Issue #553, opened 2022-02-14
[BUG] odahorde_22a crashes on map13 (Breathless)


rakohus:
**Describe the bug**
Server crashes after zdoom-hexen sector special 78 (door open after 5 minutes) activates 5 minutes into the game.

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
Wait 5 minutes on odahorde_22a, map13.

**Expected behavior**
Don't crash

**Screenshots, NetDemos, & Crash Dumps**
```
Signal number: 11
Errno: 0
Signal code: 1
Fault Address: 0x4c
/opt/odamex/bin/odasrv[0x56a300]
/opt/odamex/bin/odasrv[0x56a480]
/lib64/libpthread.so.0(+0xf630)[0x7fc8fd6d9630]
/opt/odamex/bin/odasrv(_ZN5DDoor8RunThinkEv+0x144)[0x5a5a94]
/opt/odamex/bin/odasrv(_ZN8DThinker11RunThinkersEv+0x4e)[0x533a7e]
/opt/odamex/bin/odasrv(_Z8P_Tickerv+0x6d)[0x5f329d]
/opt/odamex/bin/odasrv(_Z11SV_StepTicsy+0x27)[0x649bf7]
/opt/odamex/bin/odasrv(_Z10SV_RunTicsv+0x92)[0x64a5f2]
/opt/odamex/bin/odasrv(_ZN19CappedTaskScheduler3runEv+0x3b)[0x52f64b]
/opt/odamex/bin/odasrv(_Z9D_RunTicsPFvvES0_+0x1b9)[0x52b9c9]
/opt/odamex/bin/odasrv(_Z10D_DoomLoopv+0x17)[0x62b407]
/opt/odamex/bin/odasrv(_Z10D_DoomMainv+0x559)[0x62bae9]
/opt/odamex/bin/odasrv(main+0xd5)[0x512265]
/lib64/libc.so.6(__libc_start_main+0xf5)[0x7fc8fc8f6555]
/opt/odamex/bin/odasrv[0x51516f]
```

rakohus:
Client dies too if in single player.
[odamex_g5880b_30652_20220214T085221.zip](https://github.com/odamex/odamex/files/8059259/odamex_g5880b_30652_20220214T085221.zip)

bcahue:
Fixed with #555

rakohus:
Confirmed fixed via #562. Thanks Blair.

PR #555, opened 2022-02-14
[BUGFIX] Fix 5 minute door, generalized perpetual lifts

bcahue:
The following fixes odahorde map13 (5 minute door crash when activating) and boomedit.wad (half the time, perpetual lifts would get in invalid state on activation and not work)


bcahue:
Fixes #555

rakohus:
^^ #553

Issue #559, opened 2022-02-17
[BUG] Boom Smooth Lighting effect broken


rakohus:
**Describe the bug**
Boom smooth lighting effect broken. 

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
Play mbfedit!.wad. Go inside, first room on the right.

rakohus:
Confirmed fixed in #562. Thanks Blair.

Issue #551, opened 2022-02-13
[BUG] Eviternity Map10 Crashes on Server After Hitting a Switch


rakohus:
**Describe the bug**
Eviternity.wad. Was testing for bug #469 and came across this. Server crashes shortly after switch is hit on the southern end of the map. The switch lowers a floor, then the voodoo doll crosses a line that starts crushers with fast damage. Client seems unaffected.

**Build that the bug occurred in**
Odamex 10. Tested on Windows and Centos 8 servers, so not platform specific.

**Screenshots, NetDemos, & Crash Dumps**
![image](https://user-images.githubusercontent.com/22926877/153744940-3b1a2702-e014-4bd0-b7dd-3dbe4f074206.png)

![image](https://user-images.githubusercontent.com/22926877/153744955-eb52790e-fb09-4288-9821-e9c0aa34394e.png)

[Server Crash Dump - odasrv_g98d97_44300_20220213T081651.zip](https://github.com/odamex/odamex/files/8055180/odasrv_g98d97_44300_20220213T081651.zip)

rakohus:
This bug occurs when `co_blockmapfix` = 1

rakohus:
Confirming fixed in #562. Thanks Blair.

PR #562, opened 2022-02-18
[BUGFIX] Multiple bugfixes

bcahue:
The following GitHub issues are fixed with this pull request:

#559 (MBF Smooth Lighting not working)
#553 (5 minute door crashes on activation)
#551 - This was actually caused by a crusher blowing up a barrel and then damaging an AActor when co_blockmapfix was on. Fixed.

I also fix a bug not listed here, that had to do with generalized sectors that use model sectors, but don't copy sector flags, only reset.

Also a bugfix with perpetual genlifts getting into an invalid state and not activating. Noticeable on boomedit.wad.

Finally, I re-added the side test for ZDoom format maps for use and push line types, so you can no longer activate lifts and such from the opposite side.

### DIFF
--- a/common/p_boomfspec.cpp
+++ b/common/p_boomfspec.cpp
@@ -449,8 +449,7 @@ lineresult_s P_CrossCompatibleSpecialLine(line_t* line, int side, AActor* thing,
 		if (EV_LineTeleport(line, side, thing))
 		{
 			result.lineexecuted = true;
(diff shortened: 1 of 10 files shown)


leximayfield (approved):
LGTM

leximayfield:
The build failures seem to be problems with the CI.

Issue #261, opened 2021-03-29
Allow linking to system JsonCpp and MiniUPnPc libraries


ghost:
Currently the internal jsoncpp and miniupnpc libraries are unconditionally built and statically linked. Respective `USE_INTERNAL_JSONCPP` and `USE_INTERNAL_MINIUPNPC` CMake build options should be implemented to enable users to link against their system-provided libraries.

leximayfield:
I know our version of miniupnpc could probably use some updating, but JsonCpp's modern versions require C++11 and I don't know what that means for API compatibility with our older version that uses C++98.  Are you aware of any compatibility issues with either of these libraries?

ghost:
Gentoo has been linking odamex with the latest version of miniupnpc for some time now with no apparent issues so it can probably be updated with no expected issues. We did encounter trouble when attempting to link with a jsoncpp 1.9.3, so there does appear to be some API difference there; however, jsoncpp 00.11.0 (released last year) was able to link successfully so we can at least update to that for the various bug fixes without worrying about api issues of migrating to the jsoncpp-1.x.x series just yet.

chewi:
I've dealt with miniupnpc and have just had a go at JsonCpp. Surprisingly, I was able to build Odamex against JsonCpp 1.9.5 by just changing `CXX_STANDARD` to `11` instead of `98`. Only the server requires this change, but it works for the client too. So the question is whether you mind bumping the C++ standard at this point. Bear in mind that even the ageing CentOS 7 supports C++11. Otherwise I could make it conditional on `USE_INTERNAL_JSONCPP`, but you may not notice if it breaks later.

leximayfield:
> So the question is whether you mind bumping the C++ standard at this point.

Odamex is currently C++98 for compatibility with retrocomputing targets, most notably the OG Xbox port which has historically been compiled using Visual C++ 2003 and the Microsoft XDK.  There might be a day where we update to C++11, but that day is not going to be anytime soon.

That said, incompatibility with distro libraries has caused issues for us before.  On shorter timescales, we are currently investigating Flatpak as a way of distributing Linux binaries.  At that point, how the distro libraries are configured would be academic for folks who just want to play Odamex on Linux.

PR #569, opened 2022-02-20
Allow building against the system miniupnpc library

chewi:
`USE_INTERNAL_MINIUPNP` defaults to true and ignores `USE_INTERNAL_LIBS` because users are unlikely to have it installed.

Although miniupnpc uses pkg-config, it doesn't add any include paths to the flags because it expects your include directives to include the miniupnpc directory. We should therefore do the same with the internal build so that either can be used.

Bug: https://github.com/odamex/odamex/issues/261

----

Tested on Gentoo Linux with and without the internal miniupnpc.


leximayfield:
I'm going to give a tentative thumbs up but I am more interested in what the CI does.

leximayfield (approved):

Issue #560, opened 2022-02-17
[BUG] Autoaim selection in menu can't switch to "Never"


leximayfield:
**Describe the bug**
Just what it says on the tin - you can't select the "Never" option in autoaim settings.  It just skips over the option.

**Build that the bug occurred in**
10.0

rakohus:
Fixed in #579. Tested and confirmed. Thanks @Ru5tK1ng!

PR #579, opened 2022-02-24
Full autoaim range can now be selected in options. Addresses #560.

Ru5tK1ng:
Fixes ancient autoaim bug (at least 10 years old) and cleans up a 'for' loop for better readability.


leximayfield (approved):
LGTM

PR #587, opened 2022-02-26
[BUGFIX] Fix ZDoom Generalized Secrets

### LINKED ISSUE #586: [BUG] Hexen Format sector action 1024 no longer triggers secrets
**Describe the bug**
Hexen Format sector action 1024 (found through "Generalized Effects" in map editors like Doom Builder/2) no longer triggers secrets, including HUD messages (if enabled) and "Secrets Found" in Intermission Screen tallies. This is a regression from 0.9.5. Other generic sector actions like damage still seem to work. Sector action 8 for Doom Format maps still works.

v10.0.0 g5880b-8048

**To Reproduce**
Walk through any sector with action 1024 in a Hexen format map.

**Expected behavior**
Hexen sector action 1024 should trigger secret alerts and count towards "Secrets Found" in the Intermission.

An attachment with map MAP01 is included.
[SECRETS.zip](https://github.com/odamex/odamex/files/8145416/SECRETS.zip)

bcahue:
Added a condition to check if a player is in a secret sector to fix this, as both map formats have instances where sectors with the SECF_SECRET flag may not also have a damageamount or sector special to trigger `map_format.player_in_special_sector`. Fixes #586 (wad in that bug report works with this pull request.)

### DIFF
--- a/common/p_user.cpp
+++ b/common/p_user.cpp
@@ -932,7 +932,10 @@ void P_PlayerThink (player_t *player)
 	P_MovePlayer (player);
 	P_CalcHeight (player);
 
-	if (player->mo->subsector && (player->mo->subsector->sector->special || player->mo->subsector->sector->damageamount))
+	if (player->mo->subsector &&
+		(player->mo->subsector->sector->special ||
+		player->mo->subsector->sector->damageamount ||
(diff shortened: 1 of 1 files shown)


leximayfield (approved):
It appears the CI issue here is with the build system - not your problem.

Issue #515, opened 2022-01-23
[FEATURE] Remove Demo Bar when HUD is maximized/gone


Xenaero:
**Is your feature request related to a problem? Please describe.**
Requesting demo playback bar be included with the rest of the hud elements removed when the hud is at max size

**Describe the solution you'd like to see**
Demo playback bar vanishes from detection by my eyeballs when I hit the + key a few times

**Describe alternatives you've considered**
Turning off the demo playback bar via a bind and I don't like that solution

**Additional context**
I like making things easier on everyone

rakohus:
PR #589 merged. Thanks @Ru5tK1ng

PR #589, opened 2022-02-26
Demobar is now removed with max screenblocks. Addresses #515.

Ru5tK1ng:
Maxing out the screen size/blocks now removes the demobar during local playback.


leximayfield (approved):
LGTM

Issue #554, opened 2022-02-14
[BUG] If Automap uses custom colors, client crashes


rakohus:
**Describe the bug**
If custom automap colors are enabled, 

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
Turn on Odamex Custom Automap, Open automap

**Expected behavior**
Don't crash

**Screenshots, NetDemos, & Crash Dumps**
Crash dump

[odamex_g5880b_22724_20220214T085339.zip](https://github.com/odamex/odamex/files/8059268/odamex_g5880b_22724_20220214T085339.zip)

Acts19quiz:
I would like to add that turning on the automap overlay also causes a crash as soon as one tries to open the automap.

rakohus:
Confirmed both issues listed are fixed in PR #590.

PR #590, opened 2022-02-27
[BUGFIX] Fix automap crashes

bcahue:
This bugfix _du jour_ fixes bug #554. This was caused by a check for backsector not properly being made, and this would allow the automap loop to fall to a condition that would allow a crash. This is fixed by removing the condition that would always make linedefs without backsectors never ring true.


leximayfield (approved):
LGTM

PR #591, opened 2022-02-27
Allow building against the system FLTK library

chewi:
`USE_INTERNAL_FLTK` defaults to true and ignores `USE_INTERNAL_LIBS` because users are unlikely to have it installed.

Odamex makes of use screen scaling features in FLTK that have not yet been released. This change therefore checks for the `Fl::screen_scale` symbol and skips the associated code if it is absent. In practise, this only affects the size of the dialog window on HiDPI screens. The game window is unaffected.

Tested against FLTK 1.3.5 on Gentoo Linux.


leximayfield:
I decline to merge this PR.

As the person who wrote the boot window for Odamex using FLTK, the intention was always to eventually move on and do more bundled graphical applications and utilities for Odamex in FLTK.  And when I do, I want to write against 1.4 without worrying about using API's too new for 1.3.

leximayfield:
Also, @chewi , if you were planning on trying to do use the system Protobuf library, I'd also have to stop you.  Protobuf 3.5 is the last version that supports C++98 so we can't go higher, and the library is such a vital piece of our netcode and touches (and will touch) so many systems that I would have serious reservations about using anything other than the exact vendored version we have in-tree for all clients and servers.

chewi:
I understand. I wish FLTK would put a release out, but I guess they're still nailing down the API. There was 1.3.8 in November, but it doesn't have the screen scaling bits. I'll revisit this when 1.4 comes along.

I was planning to try Protobuf, but not yet, because their CMake build is currently broken for use in distributions, and you can't use the Autotools build here. They're fixing that up and plan to drop Autotools later. I hear you though, so I'll park this for the foreseeable future.

PR #593, opened 2022-02-27
Allow building against the system JsonCpp library

### LINKED ISSUE #261: Allow linking to system JsonCpp and MiniUPnPc libraries

chewi:
`USE_INTERNAL_JSONCPP` defaults to true and ignores `USE_INTERNAL_LIBS` because users are unlikely to have it installed.

More recent versions of JsonCpp require C++11, but Odamex targets C++98 for compatibility with older platforms. The standard is therefore only changed to C++11 when `USE_INTERNAL_JSONCPP` is false, and only for the server where JsonCpp is used. Note that C++11 still works when building against an older JsonCpp version.

Tested against JsonCpp 1.9.5 on Gentoo Linux.

Closes: https://github.com/odamex/odamex/issues/261


leximayfield:
Seems to compile OK, but have you checked to ensure that the JSON banlist can be properly written and read with the newer version of the library?

chewi:
I didn't know where JSON was used, but I've given adding, deleting, and listing a try with both versions now. All looks good.

leximayfield:
Fantastic.  If the CI isn't cross I'll merge it.

PR #596, opened 2022-02-28
ANIMDEFS temporary fix

DeathEgg:
now with just the important commit

### DIFF
--- a/common/p_spec.cpp
+++ b/common/p_spec.cpp
@@ -522,67 +522,77 @@ static void ParseAnim(OScanner &os, byte istex);
 //
 static void P_InitAnimDefs ()
 {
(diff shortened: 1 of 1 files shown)


common/p_spec.cpp
```diff
@@ -522,67 +522,77 @@ static void ParseAnim(OScanner &os, byte istex);
...
+					{
+						os.error("Unknown error reading in ANIMDEFS");
+					}
 				}
 			}
 		}
 	}
+    catch (CRecoverableError &)
+    {
+
```
leximayfield, line 583: I feel like this could use a warning, but not worth holding up the commit over.

leximayfield (approved):

PR #600, opened 2022-03-01
MAPINFO skill definitions and misc bugfixes

DeathEgg:
Adds support for the ability to define custom skills and serializes the default IWAD skill definitions. The following ZDoom MAPINFO skill definition tokens are implemented:

* AmmoFactor
* DoubleAmmoFactor
* DamageFactor 
* FastMonsters
* DisableCheats
* EasyBossBrain
* RespawnTime 
* SpawnFilter (modified so baby = easy and hard = nightmare, since Odamex doesn't support 5 difficulty mode spawn filters)
* InstantReaction
* PicName 
* MustConfirm 
* Key 
* HealthFactor
* ArmorFactor

skill_t in doomdef.h is essentially deprecated, but some horde mode code uses it still. That code should probably be modified to work with the new SkillInfos system instead.

Also, a bunch of parser bugs and crashes are fixed too.


leximayfield (approved):
LGTM

PR #603, opened 2022-03-02
replace header guards with pragma once

DeathEgg:
as stated, replaces every header guard with pragma once throughout odamex and odasrv, except for header libraries that weren't from Odamex


leximayfield (approved):
LGTM.  Seems wise to merge this ASAP.

PR #607, opened 2022-03-05
[BUGFIX] Mask extended flags to preserve compatibility on E2M7

bcahue:
This is a workaround to fix a bug report in Doomworld post https://www.doomworld.com/forum/post/2462959

This bug is caused by linedef flags being masked with 0xFE00 on E2M7 and E2M7 only. DSDA has a flag called comp_reservedlineflag that handles these erroneous flags from early maps. How I decided to solve it, since E2M7 is the only map this is known to happen, I've created a check that occurs once to see if the E2M7 FarmHash fingerprint exists, and if so, apply the reserved flag.

### DIFF
--- a/common/p_setup.cpp
+++ b/common/p_setup.cpp
@@ -50,6 +50,7 @@
 #include "p_setup.h"
 #include "p_hordespawn.h"
 #include "p_mapformat.h"
+#include <cinttypes>
 
 void SV_PreservePlayer(player_t &player);
 void P_SpawnMapThing (mapthing2_t *mthing, int position);
@@ -897,7 +898,8 @@ void P_LoadLineDefs (const int lump)
 	                        (uint64_t)(::level.level_fingerprint[14]) << 48 |
 	                        (uint64_t)(::level.level_fingerprint[15]) << 56;
 
-	StrFormat(levelHash, "%16llx%16llx", reconsthash1, reconsthash2);
+	StrFormat(levelHash, "%" PRIx64 "%" PRIx64, reconsthash1,
+	          reconsthash2);
 
 	bool isE2M7 = (levelHash == e2m7hash);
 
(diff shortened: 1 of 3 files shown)


common/p_setup.cpp
```diff
@@ -870,14 +870,44 @@ void P_LoadLineDefs (const int lump)
...
+	uint64_t reconsthash2 = (uint64_t)(::level.level_fingerprint[8]) |
+	                        (uint64_t)(::level.level_fingerprint[9]) << 8 |
+	                        (uint64_t)(::level.level_fingerprint[10]) << 16 |
+	                        (uint64_t)(::level.level_fingerprint[11]) << 24 |
+	                        (uint64_t)(::level.level_fingerprint[12]) << 32 |
+	                        (uint64_t)(::level.level_fingerprint[13]) << 40 |
+	                        (uint64_t)(::level.level_fingerprint[14]) << 48 |
+	                        (uint64_t)(::level.level_fingerprint[15]) << 56;
+
+	sprintf(levelHash, "%16llx%16llx", reconsthash1, reconsthash2);
```
leximayfield: Please do not use `sprintf`.

Additionally, I don't think that `ll` is compatible with C++98.  I would recommend using pstdint.h's `PRINTF_INT64_MODIFIER`.

leximayfield (changes requested):
Use of dangerous function and printf specifiers that are only available in modern compilers.  Small potatoes.

common/p_setup.cpp
bcahue: Changed to use StrFormat.

llx isn't C++98 but then again, neither is uint64_t. They're both C99 and C++11. I've changed this to the PRIx64 macro (pstdint.h was trying to redefine a few types).

leximayfield (approved):
LGTM

PR #608, opened 2022-03-06
fix P_GetButtonTexturePtr accessing NULL

DeathEgg:
what it says on the tin, fixes accessing data that is potentially NULL due to how bossaction line actions are used


bcahue:
Instead of this route, I would consider the following in P_UseSpecialLine:

```
if (!bossaction)
{
	P_ChangeSwitchTexture(line, repeat, true);
	OnChangedSwitchTexture(line, repeat);
}
```

This way, the server doesn't send a texture change command to clients.

leximayfield (approved):
LGTM

PR #609, opened 2022-03-06
[BUGFIX] Fix damage end sectors for Doom

bcahue:
This fixes an occasional death in sector 11 (E1M8), when sector 11 is supposed to keep you at 1hp. This now works for Doom and ZDoom maps.


leximayfield:
LGTM!

PR #613, opened 2022-03-08
Replace some BOOLs with bools

DeathEgg:
what it says, replace some typedef'd ints with actual bools throughout Odamex


leximayfield:
Unfortunately, this still has a number of unrelated changes in it.  It looks like you squashed your previous commits into one, instead of cherry picking or interactive-rebasing the bool commits out of your old branch.

I'm going to close this for now.

PR #617, opened 2022-03-10
SNDINFO Improvements

DeathEgg:
Ported over a few ZDoom SNDINFO upgrades, namely $alias and $random. I also converted the menu quit to use it (but not the randomized enemy sounds for demo compatibility reasons) as a part of abstracting gameinfo so it can be serialized instead of hard coded.

### DIFF
--- a/client/src/s_sound.cpp
+++ b/client/src/s_sound.cpp
@@ -585,14 +585,20 @@ static void S_StartSound(fixed_t* pt, fixed_t x, fixed_t y, int channel,
 
 	sfxinfo_t* sfxinfo = &S_sfx[sfx_id];
 
-	if (sfxinfo->link != sfxinfo_t::NO_LINK)
+	while (sfxinfo->link != sfxinfo_t::NO_LINK)
+	{
+		sfx_id = sfxinfo->link;
 		sfxinfo = &S_sfx[sfxinfo->link];
+	}
 
 	if (!sfxinfo->data)
 	{
 		I_LoadSound(sfxinfo);
-		if (sfxinfo->link != sfxinfo_t::NO_LINK)
+		while (sfxinfo->link != sfxinfo_t::NO_LINK)
+		{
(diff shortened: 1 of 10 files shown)


client/src/s_sound.cpp
```diff
@@ -1145,221 +1159,307 @@ static struct AmbientSound {
...
+				{
 					// Hexen-style $MAP command
-					sndinfo = COM_Parse (sndinfo);
-					sprintf (com_token, "MAP%02d", atoi (com_token));
-					level_pwad_info_t& info = getLevelInfos().findByName(com_token);
-					sndinfo = COM_Parse (sndinfo);
+					char mapname[8];
+
+					os.mustScanInt();
+					sprintf(mapname, "MAP%02d", os.getTokenInt());
```
leximayfield, line 1384: At some point this needs to be removed.

common/s_sound.h
```diff
@@ -39,136 +38,142 @@ typedef struct sfxinfo_struct sfxinfo_t;
...
-	unsigned int ms;					// [RH] length of sfx in milliseconds
-	unsigned int next, index;			// [RH] For hashing
-	unsigned int frequency;				// [RH] Preferred playback rate
-	unsigned int length;				// [RH] Length of the sound in bytes
+	char name[MAX_SNDNAME + 1]; // [RH] Sound name defined in SNDINFO
+	unsigned normal;            // Normal sample handle
+	unsigned looping;           // Looping sample handle
+	void* data;
+
+	int link = NO_LINK;
```
leximayfield: This is not valid C++98.

leximayfield (changes requested):
Please fix this one C++11-ism and you're good.

common/s_sound.h
DeathEgg: Fixed in latest commit

leximayfield (approved):

PR #619, opened 2022-03-13
Adjustments to horde

leximayfield:
This commit has a number of adjustments for Horde.

- At the start of waves, you now get a difficulty grade for that wave based on the maximum group spawn HP.
  - This is the same value used for ordering waves so easy waves appear at the start of the game and hard waves at the end.
  - The values are as follows, and yes, this is the Dean's grading system.
    - 0-499: E
    - 500-999: D
    - 1000-1999: C
    - 2000-2999: B
    - 3000-4999: A
    - 5000+: X
- Also the start of waves, you now get a notification if there are new weapons available for you to pick up.
- The E4M1 wave no longer spawns health bonuses as powerups.
  - It should be far more likely to get a Rocket Launcher powerup as a result, which can help with the bosses.
- Most significantly, monster spawning has changed.
  - The speed of spawning is now a continuum that depends on how many monsters the map has.
  - Using default settings, in a relatively empty map, it will spawn monsters every 1-3 seconds.  In a relatively full map, it will spawn monsters every 2-6 seconds.
  - The rate of spawning is governed by four cvars.
    - `g_horde_spawnempty_min`: Minimum number of seconds it takes to spawn a monster in an empty horde map.  Defaults to 1.
    - `g_horde_spawnempty_max`: Maximum number of seconds it takes to spawn a monster in an empty horde map.  Defaults to 3.
    - `g_horde_spawnfull_min`: Minimum number of seconds it takes to spawn a monster in a full horde map.  Defaults to 2.
    - `g_horde_spawnfull_max`: Maximum number of seconds it takes to spawn a monster in a full horde map.  Defaults to 6
  - If the map gets too full, monsters will stop spawning.  However, as soon as some are killed, more will spawn - there is no more RELAX state at all.
  - If the map gets too empty, monsters will spawn in faster than the minimum.
  - The specific values that "relatively empty" and "relatively full" designate can be seen using the `hordeinfo` command, under "Min Total Health" and "Max Total Health".  This command also gives you insight into how they are calculated.
  - There is a mechanism that tries not to spawn too many monsters in the first 60 seconds of a wave.
  - There is another mechanism that slows down the spawn rate during the boss battle.  Essentially, the "alive" health used to determine the spawn rate is bumped to halfway between the current alive health and the maximum health.
    - In other words, the closer you are to the minimum health, the more help this boost is.  However, it never exceeds the maximum spawn rate and it doesn't cut off any sooner.
- You can now override a base wave by using the same name as the earlier wave.


rakohus:
The pre-game countdown where sv_gametypename is displayed, and the post-game message, no longer work in this branch

rakohus:
Confirm that e7743a5 fix worked.

Not sure if this is a new bug, but no message is displayed on client when chainsaw is picked up from supply box.

leximayfield:
This is enough for one minor release.

Issue #632, opened 2022-03-22
[BUG] Slow Ceiling Crusher ([141] W1 - Crusher Start with Slow Damage) behavior incorrect


rakohus:
**Describe the bug**
Action 141 behaves differently than expected.

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
Run the test map. Walk forward. Keen will be crushed.

**Expected behavior**
In vanilla, keen is crushed slowly. 

**Screenshots, NetDemos, & Crash Dumps**
Behavior in DSDA:
https://user-images.githubusercontent.com/22926877/159506907-a7bb81b5-06f8-44c8-8ba0-3bc7e82ea66e.mp4

**Additional context**
[keencrushtest.zip](https://github.com/odamex/odamex/files/8325213/keencrushtest.zip)

rakohus:
Confirmed fixed. Thanks @DCGRetrowave for bringing this to our attention.

PR #634, opened 2022-03-23
[BUGFIX] Fix Bug 632 (Incorrect crusher behavior)

bcahue:
This bug was caused by improperly gating ZDoom crusher behavior and Doom crusher behavior. While doing this, I decided to also match up the silent behavior exhibited by generalized sectors in Boom, by not making a plat stop sound if the crusher is a silent generalized crusher (reference wad: boomedit.wad). This differs from Doom silent crushers which do make a plat stop sound, but don't make a "ceiling move" sound.

Fixes bug #632


rakohus:
Confirmed this fixes #632

PR #639, opened 2022-03-25
[BUGFIX] Trigger ZDoom action specials on items that are picked up

### LINKED ISSUE #638: [BUG] Hexen Format: Tagged Actions on Pickups Do Not Work
**Describe the bug**
Picking up items with defined actions do not activate on tagged sector targets. Killing a monster DOES activate the tagged action correctly.

Odamex 10

**To Reproduce**
Run attached wad. Red keycard should lower the floor. Blue keycard should drop the lift. Yellow keycard should open the door.

**Expected behavior**
Activate targets.

Test Wad:
[pickupactiontest.zip](https://github.com/odamex/odamex/files/8354006/pickupactiontest.zip)

image

bcahue:
Fixes #638. Previously only action specials would play on killed monsters. Now, items that are consumed by a player will trigger an action special trigger (and then remove said trigger).

### DIFF
--- a/common/p_interaction.cpp
+++ b/common/p_interaction.cpp
@@ -1299,6 +1299,16 @@ void P_TouchSpecialThing(AActor *special, AActor *toucher)
 	if (!serverside && (!cl_predictpickup || !P_SpecialIsWeapon(special)))
 		return;
 
+	// [Blair] Execute ZDoom thing specials on items that are picked up.
+	// (Then remove the special.)
+	if (map_format.getZDoom() && special->special)
+	{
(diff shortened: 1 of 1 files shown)


rakohus:
Doesn't work online in coop where the pickup stays on the map

rakohus:
Re-tested on newest commit and it seems fine. Must have been user error on my side.

bcahue:
Here's a vid to confirm: https://www.youtube.com/watch?v=YxGqeJ2daKM

leximayfield (approved):
LGTM

PR #641, opened 2022-03-27
[BUGFIX] Return if unrecognized sound effect is encountered

DeathEgg:
Returns if the sound effect that is attempted to be played doesn't exist rather than attempting to play it and crashing


rakohus:
Unfortunately, didn't work. 
image

rakohus:
Seems good now

leximayfield (approved):
LGTM

PR #642, opened 2022-03-28
[BUFIX] Allow generic locked doors to make sounds when opening

bcahue:
Does what it says in the subject. Generic locked doors weren't making sounds when opening, now they do.


leximayfield (approved):
LGTM

PR #643, opened 2022-03-28
Use OLumpName in place of some lump names that use char[8]/char[9]

DeathEgg:
I didn't get every one of them since some were more difficult to untangle, but I've replaced a lot of char[8]/char[9] with OLumpName, since that class takes care of keeping things uppercase for you and generally is safer to use for lump names.


leximayfield:
Delay until 10.1+1

fuzztooth (approved):

PR #645, opened 2022-04-02
Enable all serverside settings on -warp

leximayfield:
This makes them fully client controlled, every time you `-warp`.  Previously, this only happened on new game.


leximayfield:
This patch was tested by Ralphis.

PR #648, opened 2022-04-05
[BUGFIX] Fix "Dark Lift" ZDCTFMP MAP08 Phenomenon

bcahue:
This was caused by the client clamping specials to `byte`, thereby changing their intended special completely. This was generally okay in all maps both ZDoom and Boom, except for Generalized specials, which extend far beyond the range of `byte` and thus had unintended consequences. This is a clientside-only change, the server does the correct thing already.

https://cdn.discordapp.com/attachments/236518337671200768/959427197335121920/unknown.png

_Goodbye spooky lift, 2022-2022._


leximayfield (approved):
LGTM

PR #651, opened 2022-04-10
[BUGFIX] Fix Multiplayer Vanilla Demos

### LINKED ISSUE #557: [BUG] Can't play old Vanilla netdemo
Tried to load an old demo with Doom 2 1.9 using Odamex 10.0
```
]playdemo nos_meg1.lmp
Playing DOOM demo C:\Games\odamex\nos_meg1.lmp

MAP11: "'o' of destruction!"
```
*crash*

[nos_meg1.lmp](http://www.doom2.net/noskill/nos_meg1.lmp)

bcahue:
Fixes #557. This was caused by a player that was local to the client but didn't have a team assigned. Servers seem to be able to handle this but clients choke. The fix allows the demo in the bug to play again without hard crashing. Fixing this also revealed another bug with multi-player demos, where players who weren't the consoleplayer can't pause or resume the game, the pause command just gets dropped, causing the demo to desync. This is fixed.


leximayfield (approved):
Looks mostly good to me.  I did notice the `strcpy` but since it was existing code I won't make you change it - getting rid of those is a project-wide concern.

PR #656, opened 2022-04-16
Use initializer lists instead of individual push_backs when initializing MapInfoDataContainer

DeathEgg:
Exactly as said, replaces macros for individual push_backs to std::vector with one initializer list. easier to read, write code for, and faster


leximayfield:
For some reason github is not giving me the option to update this branch to the latest non-CentOS 7 commit.  I'll merge it anyway and see how it goes.

PR #658, opened 2022-04-17
Use constexpr where possible

DeathEgg:
Change types to constexpr where it makes sense throughout Odamex


fuzztooth:
Bumping - is this still useful?

electricbrass:
Yes, this would still be useful.

client/src/c_console.cpp
```diff
@@ -1331,8 +1331,7 @@ int C_BasePrint(const int printlevel, const char* color_code, const std::string&
 
 		// We need to know if there were any new lines being printed
 		// in our string.
-
-		int newLineCount = std::count(logStr.begin(), logStr.end(), '\n');
+		const int newLineCount = std::count(outline, outline + strlen(outline), '\n');
```
electricbrass, line 1334: Looks like this line got messed up in a merge, should be `const int newLineCount = std::count(logStr.begin(), logStr.end(), '\n');`

client/src/c_console.cpp
```diff
@@ -1366,6 +1365,55 @@ int C_BasePrint(const int printlevel, const char* color_code, const std::string&
 	return newStr.length();
 }
 
+FORMAT_PRINTF(1, 2) int STACK_ARGS Printf(const char* format, ...)
```
electricbrass, line 1368: These FORMAT_PRINTF functions no longer exist in protobreak, looks like they accidentally got brought in from stable at some point. They should be deleted.

client/src/r_sky.cpp
```diff
@@ -174,8 +174,6 @@ void R_GenerateLookup(int texnum, int *const errors); // from r_data.cpp
 
 void R_InitSkyMap()
 {
-	fixed_t fskyheight;
-
```
electricbrass, line 178: This variable still needs to be declared somewhere. Looks like an earlier commit had it declared at the point it's first used down below.

client/src/d_netinfo.cpp
```diff
@@ -34,7 +34,7 @@
 
 // The default preference ordering when the player runs out of one type of ammo.
 // Vanilla Doom compatible.
-const byte UserInfo::weapon_prefs_default[NUMWEAPONS] = {
+constexpr byte UserInfo::weapon_prefs_default[NUMWEAPONS] = {
```
electricbrass, line 37: Making this constexpr causes the Ubuntu 18.04 and Fedora builds to fail. To keep it constexpr, the initialization can be moved to the header.

electricbrass (changes requested):
Overall this looks good. There's a couple spots where it looks like it got messed up during merges.

PR #660, opened 2022-04-27
[BUGFIX] Hash levels properly (Fixes E2M7 hack not always working)

bcahue:
This ensures the E2M7 hack will work properly every time, no matter what level you run beforehand. This was caused by a static member not being cleared in the fingerprint function. Instead of clearing this, I simply made it non-static as it's not needed to be static.


common/p_setup.cpp
```diff
@@ -1528,7 +1528,7 @@ void P_GenerateUniqueMapFingerPrint(int maplumpnum)
 	unsigned int length = 0;
 
 	typedef std::vector<byte> LevelLumps;
-	static LevelLumps levellumps;
+	LevelLumps levellumps;
```
leximayfield, line 1531: oof

leximayfield (approved):
LGTM

PR #663, opened 2022-05-05
[BUGFIX] Fix #662 (Sector Flags Not Serializing on Save Game)

### LINKED ISSUE #662: [BUG] Non-secret sectors are reported as secrets and insta-kill the player
**Describe the bug**
In certain WADs, specific sectors will produce one or both of two behaviors depending on the build settings:

1. The sectors, which aren't secrets, will periodically produce the "A secret is revealed!" message. Walking back and forth will sometimes cause the message to be shown repeatedly, incrementing the secrets counter on the map screen past the number of actual secrets.
2. The same sectors will occasionally insta-kill the player, playing the exploded-body sound effect.

Odamex v10.1.0 (Stable, G5BE2-8307) compiled from commit 5be225f3bb6aad88d96c965c136145c75e067893 (HEAD at time of writing) on Ubuntu 20.04 using:

- `$ cmake ..` (bugs 1 and 2)
- `$ cmake .. -DCMAKE_BUILD_TYPE=Release` (bugs 1 and 2)
- `$ export CC=/usr/bin/clang; export CXX=/usr/bin/clang++; cmake ..` (**only** bug 1)
- `$ export CC=/usr/bin/clang; export CXX=/usr/bin/clang++; cmake .. -DCMAKE_BUILD_TYPE=Release` (bugs 1 and 2)

gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
clang version 10.0.0-4ubuntu1
cmake version 3.16.3

**To Reproduce**
1. Build as described above
2. Run $ odamex -iwad DOOM2.WAD -file ICARUS.WAD
3. Start a new game on MAP01 (difficulty: ITYTD and HNTR both produced issue), and get the blue key for access to the chapel area
4. Save the game before entering the blue key door
5. Enter the room, walk onto the wooden chair, and kill the imps that appear
6. Walk around the center area of the room, near the stairs and the green floor design

**Expected behavior**
No secret is reported, and the player does not go splat.

[odamex_secret_insta_kill.zip](https://github.com/odamex/odamex/files/8598251/odamex_secret_insta_kill.zip)

**Additional context**

- Tested in single player with all compatibility options set to the default "OFF"
- In testing, it can take anywhere from a few seconds to 30 seconds for the death bug to trigger while walking around the room
- The "A secret is revealed!" message may appear several times before the player dies, as often as every few seconds
- The issue will occur consistently after reloading a save before the affected area
- I've observed the same bugs together in other maps (e.g., TNTR.wad MAP32). In every case, the bugs occur in a consistent location in the map.
- As visible in one of the screenshots, I also saw hall-of-mirrors artifacts on the grey stairs in the Icarus room; these did not appear in Crispy Doom when I tested the same room there.

bcahue:
This will fix the phenomenon of players dying, or secrets getting revealed twice on certain sectors (seem to be secret sectors) when saving/loading. The same function is called on level complete, so its likely this is called in certain situations that aren't game saving/loading. Fixes #662.

### DIFF
--- a/common/p_floor.cpp
+++ b/common/p_floor.cpp
@@ -804,6 +804,10 @@ DFloor::DFloor(sector_t *sec, DFloor::EFloor floortype, line_t *line,
 		{
 			sec->floorpic = line->frontsector->floorpic;
 			sec->special = line->frontsector->special;
+			sec->flags = line->frontsector->flags;
+			sec->damageinterval = line->frontsector->damageinterval;
+			sec->leakrate = line->frontsector->leakrate;
+			sec->damageamount = line->frontsector->damageamount;
 		}
 		break;
 
@@ -816,6 +820,9 @@ DFloor::DFloor(sector_t *sec, DFloor::EFloor floortype, line_t *line,
 		// --> should not affect compatibility <--
 		m_NewSpecial = sec->special;
 		m_NewFlags = sec->flags;
+		m_NewDmgInterval = sec->damageinterval;
+		m_NewLeakRate = sec->leakrate;
+		m_NewDamageRate = sec->damageamount;
 
 		//jff 5/23/98 use model subroutine to unify fixes and handling
 		sec = P_FindModelFloorSector (m_FloorDestHeight,sec);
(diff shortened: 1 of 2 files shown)


bcahue:
The serialization only fixed half of it. Doom's Floor lower and change and raise and change functions don't set sector flags as well as leakrate, damageinterval and damagerate, which end up being wild pointers when the floor runs tics and the change is made. I set those as well to fully fix this issue.

leximayfield (approved):
LGTM

PR #665, opened 2022-05-08
[BUGFIX] Fix resurrected monsters not having height when using A_HealCorpse

bcahue:
A leftover from DSDA, this will ensure that A_HealCorpse restores actor height depending on co_novileghosts, similar to how A_VileChase is done. Fixes #664.

PR #666, opened 2022-05-12
[BUGFIX] Fix judgment.wad loading

bcahue:
The following PR will fix the loading of JUDGMENT.WAD, a megawad that showcases MBF21 capabilities. This was caused by the program selecting an improper overload for OStringToUpper. This fix dispels that ambiguity.

(Please note, this fix has code from #665, which should be merged before this one.)


leximayfield (approved):
Thanks, LGTM.  The more I write C++ the less I like overloading unless it's 99.9% unambiguous.

PR #670, opened 2022-05-19
[BUGFIX]Fix A_PlaySound

bcahue:
Previous to this fix, A_PlaySound would play locally and online, which resulted in the activator hearing the same thing twice. Ambient sounds also play twice online, as seen in AD_MORTEM.WAD MAP02. This will fix this behavior.


leximayfield (approved):
LGTM - nice catch

PR #672, opened 2022-05-22
Add "AUTOMAP" graphic for said section in options menu.

Acts19quiz:
When going through Odamex's options, going into the AUTOMAP section (OPTIONS>DISPLAY OPTIONS>AUTOMAP) confusingly displays the graphic "MESSAGES" when that is a different section. This hopes to alleviate that in a satisfactory manner.

[Screenshots of current menu.](https://user-images.githubusercontent.com/63941237/169673057-b6c29e62-52d4-4928-bd7d-ea89f14f7a7e.gif)


leximayfield:
This is really cool lookin'.  I'll let the build run and we'll see where that ends up.

Acts19quiz:
If you're okay with it, I could expand the scope of the PR to include something similar for "HEADS-UP DISPLAY", or alternatively file a separate PR for that. Though "DISPLAY OPTIONS" sort of works for that. This (AUTOMAP) was more egregious. What do you think?

PR #677, opened 2022-06-18
Add visible values for sliders

leximayfield:
This commit adds visible numbers for slider controls.

Not all of them, RGB kind of speaks for themselves.

odamex_HL9pMA8gO1


rakohus:
Looks great. My only suggestion here is to translate the values to a different color, like the typical doom green, so they really pop out and are differentiated in the menu.

Issue #674, opened 2022-05-25
[FEATURE] Improve kill feed display durations


rakohus:
**Is your feature request related to a problem? Please describe.**
The kill feed can scroll by somewhat quickly, especially given Doom's pace. 

**Describe the solution you'd like to see**

- Either increase the kill feed messages by 1 second, or make it variable.
- Make any message that the console player is involved in last longer (linger at the top spot longer). Source games do this.

rakohus:
PR #678 addressed this. Thanks @AlexMax

PR #678, opened 2022-06-18
Make feed fadeout time adjustable

leximayfield:
This adds a new `hud_feedtime` cvar and menu item that adjust the amount of time that the killfeed obituaries stay onscreen.

This PR addresses #674 .


rakohus:
This is very good. My only suggestion is to change the menu text to "Kill Feed Timeout" and "Kill Feed Display" instead of "Feed Obits". The current shorthand might not be easy to understand for non-native speakers.

Issue #679, opened 2022-06-21
[FEATURE] Introduce a `limit` field to HORDEDEF


rakohus:
**Is your feature request related to a problem? Please describe.**
We have a `chance =` token, it would be good to have a `limit` token. Limit the total amount of any monster type that is active on the map.

**Describe the solution you'd like to see**
Example:
`addmonster "Archvile" {
  limit = 1
}`
This would stop the game from spawning any Archviles if 1 is already alive on the field. Once it dies, the game could spawn another. It could accept any integer value.

If you want to get really tricky, you could use this to spawn two separate monsters as a boss simultaneously but that's a bit outside of the scope of what I'm trying to accomplish here i.e. stop 10 archviles from spawning in and crushing everyone in the 3rd wave.

electricbrass:
Do you want bosses to be counted towards the limit? I see a few possibilities. 
1. They do count and are treated the same as any other spawn and can be blocked by the limit.
2. They don't count at all towards it nor is it checked for their spawns.
3. They count toward the limit, but it is ignored when spawning bosses, letting the count exceed the limit.
4. Non-boss monsters are killed if bosses would exceed the limit.
5. Boss and non-boss monsters of the same type can have their own limits (useful for your suggestion at the end probably)

I went with #2 in my PR, and assume #1 is not wanted, but thought I'd ask what is thought to be best.

rakohus:
Definitely 2 at a minimum, but 5 is the natural extension I think

electricbrass:
Alright, I think I can change it to 5 tonight

electricbrass:
Done

PR #680, opened 2022-06-22
Add `limit` field to hordedef

electricbrass:
Allows limiting the total number of a particular monster type alive in the map at once.

Addresses issue #679


common/p_horde.cpp
```diff
@@ -202,6 +203,8 @@ class HordeState
 	int m_nextSpawn;
 	int m_nextPowerup;
 	corpseCollector_t m_corpses;
+	std::unordered_map<mobjtype_t, int> m_monsterCounts;
+	std::unordered_map<mobjtype_t, int> m_bossCounts;
```
leximayfield: Odamex 10.0 is C++98.  `OHashTable` is what we use until 11.0.

electricbrass:
I think I fixed any C++11-isms alongside conforming to the style

common/p_hordedefine.cpp
```diff
@@ -295,6 +295,7 @@ bool P_HordeSpawnRecipe(hordeRecipe_t& out, const hordeDefine_t& define,
 
 	out.type = outType;
 	out.count = outCount;
+	out.limit = config->limit;
```
leximayfield: I think that making the limit part of the recipe is the wrong strategy.  Instead, why not take the limit into account when constructing the recipe?  That way, the spawner doesn't have to know about the limit and work any differently, and if there's no viable way to spawn any monsters of a given type due to the limit being hit, then you can try another attempt at making a recipe with a different monster.

Also, I think that it might be a wise idea to make this limit subject to the `sv_skill` health multiplier.  You can convert the monster limit to an HP limit and then scale it from there.

common/p_horde.cpp
```diff
@@ -202,6 +203,8 @@ class HordeState
 	int m_nextSpawn;
 	int m_nextPowerup;
 	corpseCollector_t m_corpses;
+	std::map<mobjtype_t, int> m_monsterCounts;
+	std::map<mobjtype_t, int> m_bossCounts;
```
leximayfield: I appreciate you trying to make this C++98 compatible, but I would much prefer if we used `OHashTable` instead of a `std::map`.

leximayfield (changes requested):

common/p_horde.cpp
electricbrass: Okay, changed it. Still figuring out both C++ and Odamex.

common/p_hordedefine.cpp
electricbrass: Okay, I'll give it a go tomorrow. Now that you say this, I definitely agree.

common/p_hordedefine.cpp
```diff
@@ -213,7 +215,7 @@ size_t P_HordePickDefine(const int current, const int total)
  * @param wantBoss Caller wants a boss.
  */
 bool P_HordeSpawnRecipe(hordeRecipe_t& out, const hordeDefine_t& define,
-                        const bool wantBoss)
+                        const bool wantBoss, MobjTypeTable monsterCounts)
```
leximayfield: I think you meant to pass in by const reference here?

electricbrass: Yes thank you

leximayfield:
Okay, here's the deal.  There are one or two things about this commit that need fixing, and I do think this commit needs some smoke-testing as well.  However, this is a feature that I was actually going to implement myself before I got so busy, and I am incredibly grateful to you for your contribution.

So here's what I'm going to do.  I'm going to accept your PR not into `stable` but into a branch to itself.  This means that the CI will build the branch for us, plus the buildserver will host the binaries and we can test your feature thoroughly.  It may or may not make it in time for 10.2, but it _will_ make it in.

Thanks again!  We really appreciate it!

PR #684, opened 2022-06-27
[BUGFIX] Fix client scrolling desyncs

### LINKED ISSUE #683: [BUG] Boom Floor Scroll Special 253 (scroll + move things) is faster on a connected client than server or local game
**Describe the bug**
Floors that use the boom scrolling special 253 move FASTER on the client when it is connected to a server. This causes the client to mispredict on boom conveyor belt scripts. This bug does NOT occur when playing a regular single player game.

You can use the 'version' console command in Odamex to find this.

**To Reproduce**
Run the wad in single player first. Press the switch to open the barrel stopping door and observe. At the blue pole, the doors should open. At the red pole, they should close. At the eye, the barrel should teleport back to the beginning of the conveyor and repeat. It works flawlessly in single player.

Now run it on a server (default config is fine). Connect your client. Press the switch and observe. The flats are moving much faster and the client _thinks_ the barrel is moving much faster as well. If you continue to watch, you will see the barrel predicted to teleport back to the beginning of the conveyor belt before the server corrects it with the teleporter.  

**Expected behavior**
Uniform behavior

[scroll_net_test.zip](https://github.com/odamex/odamex/files/8985911/scroll_net_test.zip)

bcahue:
In Odamex, scrollers are designed to be initiated by the server ONLY, and any client scrollers that start independent of the server will desync as there is no mechanism to reconcile improper scroller/actor position from the client/server. This seems to manifest itself by the scrollers "adding" the scroll value between client and server on the client.

This introduced a bug when MapFormat was introduced, as Boom scrollers were initiated on clients as well as servers. ZDoom scrollers still had the correct behavior.

This fixes #683.

### DIFF
--- a/common/p_boomfspec.cpp
+++ b/common/p_boomfspec.cpp
@@ -1689,6 +1689,10 @@ void P_SpawnCompatibleSectorSpecial(sector_t* sector)
 
 void P_SpawnCompatibleScroller(line_t* l, int i)
 {
+	// [Blair] don't run scrolling on clients to prevent desyncs
+	if (IgnoreSpecial)
+		return;
+
 	fixed_t dx = l->dx >> SCROLL_SHIFT; // direction and speed of scrolling
 	fixed_t dy = l->dy >> SCROLL_SHIFT;
 	int control = -1, accel = 0; // no control sector or acceleration


leximayfield:
LGTM - CI errors look spurious.

PR #687, opened 2022-06-27
[BUGFIX] Clear specials on the client as well as the server

### LINKED ISSUE #685: [BUG] W1 Teleport remains as a 'fake' active trigger when playing online
**Describe the bug**
Playing online in Sunlust Map07 showcases how a linedef (ID 3756, 3741) with the special action of 39 - W1 Teleport will remain active after the one time use. This could be only in relation to the fact that an item triggers this event. A super shotgun is sent through these linedefs to tele it across the corridor. Walking over the linedef after this events will instantly flash the teleport destination on screen, for a brief second, and also fixes the player's camera and aim to be in line as the teleport destination's angle, but not physically move the player over and no fog is shown. This also stutters movement and locks the mouse in place (similar to when one would take a normal teleporter).

10.1.0

**To Reproduce**
I can consistently recreate this on the WDL Sunlust Survival server.

**Expected behavior**
W1 should only be used once, and shouldn't be able to manipulate player view and controls.

Demo Attached

**Additional context**
[Sunlust Map07 Demo (2).zip](https://github.com/odamex/odamex/files/8994867/Sunlust.Map07.Demo.2.zip)

bcahue:
Due to the logic, the client would fail to clear special lines for single-activation items and cause a desync, mainly for teleport lines where the client tries to handle it.

Fixes #685

### DIFF
--- a/common/p_spec.cpp
+++ b/common/p_spec.cpp
@@ -1942,9 +1942,12 @@ void P_CrossSpecialLine(line_t*	line, int side, AActor* thing, bool bossaction)
 		result = map_format.cross_special_line(line, side, thing, bossaction);
 	}
 
-	if (serverside && result.lineexecuted)
+	if (result.lineexecuted)
 	{
-		SV_OnActivatedLine(line, thing, side, LineCross, bossaction);
+		if (serverside)
+		{
+			SV_OnActivatedLine(line, thing, side, LineCross, bossaction);
+		}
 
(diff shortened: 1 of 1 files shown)


leximayfield (approved):
LGTM

leximayfield:
It appears that the CI failure is spurious and down to some strange MinGW error.

PR #688, opened 2022-06-27
[BUGFIX] Fix last frag not being counted in WDLStats Deathmatch Modes

bcahue:
This occurred because P_KillMobj would fire before recording the frag, which would call the end game functions and finalize the log. Now, WDL events will record the kill before the log is finalized.


leximayfield (approved):
LGTM

PR #689, opened 2022-06-28
[BUGFIX] Share all scroller info

bcahue:
Because Odamex depends on the server to tell the client when scrolling begins, the client is completely dependent on the server's initial information for the scroller. Odamex currently omits some information so some scrollers will cause desyncs as their online behavior will not match their offline behavior. Take for instance, this SSG disappearing trick in Sunlust MAP07:

Here's what it looks like on Odamex 10.1 Stable:
https://youtu.be/GHlYsH6NZbE

And here's what it looks like for this patch (Odamex 11 Protobreak):
https://youtu.be/xwikmdbOo2o

And how it looks in single-player (Odamex 11 Protobreak):
https://youtu.be/8vsCZMjoAsc


leximayfield:
I think this could actually be applied to `stable` with a small tweak.  All you did was add two fields to the protocol and all unsent fields are assumed to be 0.  I think you would need the client to assume `control` is -1 if the server is < 10.2, but in other cases it should just work.

Issue #558, opened 2022-02-15
[FEATURE] Setting for how long chat messages are shown


weatherman115:
**Is your feature request related to a problem? Please describe.**
Reading longer chat messages within ~3 seconds of getting them can be hard to do, especially if you're focused on gameplay.

**Describe the solution you'd like to see**
In-game setting for how long a chat message can stay on screen.

**Describe alternatives you've considered**
Additional option for how many lines of messages can be shown at once
Ability to scroll through previous chat messages

**Additional context**
Happened a lot during the Horde event.

rakohus:
This was addressed by AlexMax in #692, merged into stable, and will be in 10.2. Thank you for the request @weatherman115.

PR #692, opened 2022-06-30
Add message line and mid message timeouts to menu

leximayfield:
We actually technically had support for this in-tree, but now it's surfaced in the menus as well (under messages).  The cvars have also been changed to floats, to allow non-integer second wait times.

Addresses #558.


rakohus:
Tested it. checks out

PR #713, opened 2022-07-21
Skip artifact upload if actions run on fork

electricbrass:
Let me know if there's some reason this isn't wanted. I just keep finding it annoying when I try to figure out for a minute why something failed before realizing it was just because it was on a fork.


leximayfield:
Thanks for this.

Issue #707, opened 2022-07-17
[FEATURE] Compat option to uncap Pain Elemental-spawned Lost Soul limit


Xenaero:
**Is your feature request related to a problem? Please describe.**
Ran into this often during Horde sessions, where the limit for Pain Elemental's Lost Souls on the map was very quickly reached, rendering Pain Elementals a non-threat and easily neutralized.

**Describe the solution you'd like to see**
While this behavior will definitely aid new and amateur players, and maintain compatibility for obvious reasons, I think the best implementation would be a compatibility var that toggles this limit on or off. Server administrators can then disable it for MAXIMUM LOST SOUL MAYHEM.

**Describe alternatives you've considered**
None.

**Additional context**
None.

rakohus:
Pushed to stable in #716. Thanks for the patch @electricbrass

PR #716, opened 2022-07-23
Add a lost soul limit cvar

electricbrass:
I know there is a desire to limit the number of cvars, but this keeps being an issue with pain elemental bosses in horde mode.

Addresses #707

### DIFF
--- a/client/src/m_options.cpp
+++ b/client/src/m_options.cpp
@@ -127,6 +127,7 @@ EXTERN_CVAR (cl_deathcam)
 EXTERN_CVAR (co_fineautoaim)
 EXTERN_CVAR (co_nosilentspawns)
 EXTERN_CVAR (co_boomphys)			// [ML] Roll-up of various compat options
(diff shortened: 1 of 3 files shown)


rakohus:
I think there still needs to be a reasonable hard cap on souls to prevent a server from spawning thousands of lost souls and killing everything. I'm not sure where that cap should be, but it doesn't seem like a good idea to be limitless.

electricbrass:
Yea, I agree. I actually meant to ask but I forgot. Any idea on a general range? 50? 128? 256?

rakohus:
After doing some research, it looks like boom and higher remove the cap entirely. With that in mind, I would advocate for:

- removing the cap entirely when playing locally (offline)
- capping at something like 128 when playing online

As an aside, the cvar also needs to automatically default to vanilla behavior during vanilla demo playback

electricbrass:
Okay sounds good. I'll do that soon

leximayfield (approved):
LGTM!

PR #720, opened 2022-07-23
[FEATURE] Implement r_softinvulneffect

### LINKED ISSUE #703: [BUG] Custom colormap/invul range is not used
**Describe the bug**
It appears as though Odamex does not parse a pwad's custom COLORMAP or at the very least, its invul range. 

10.1.0 (G0A84-8300)

**To Reproduce**
The easiest way is to load Eviternity and give yourself Invul (idbeholdv). You will see the normal Doom invul colormap range instead of the custom one defined by the wad's COLORMAP.

**Expected behavior**
A lovely cascade of teals to show up when I am invulnerable in Eviternity. Expected working behavior is reproducible on DSDA-Doom.

N/A

**Additional context**
N/A

bcahue:
The invuln that is experienced in Horde is eye-watering at best, and a photosensitive epilepsy hazard at worst. I've went ahead and implemented a mockup by @rakohus that changes the effect to such:

- Only the weapon sprite has the invuln effect drawn.
- The light googles effect is also drawn in order to provide the added "shadow" invuln benefit of sussing out fuzzy monsters.
- Colored sectors don't affect the weapon sprite when using this and the player is invuln, as the color translation doesn't work well with inversion and colored sector lighting.

Here's a video of the new effect:
https://www.youtube.com/watch?v=L8km6GVQts4

If implemented, this will resolve #703 in spirit, since we generate colormaps off palettes, and cannot use the COLORMAP lump.

Test wad here: [invuln-color-test.zip](https://github.com/odamex/odamex/files/9186134/invuln-color-test.zip)

### DIFF
--- a/client/src/m_options.cpp
+++ b/client/src/m_options.cpp
@@ -97,6 +97,7 @@ EXTERN_CVAR (r_forceteamcolor)
 EXTERN_CVAR (r_teamcolor)
 EXTERN_CVAR (r_forceenemycolor)
 EXTERN_CVAR (r_enemycolor)
+EXTERN_CVAR (r_softinvulneffect)
 EXTERN_CVAR (cl_mouselook)
 EXTERN_CVAR (in_autosr50)
 EXTERN_CVAR (gammalevel)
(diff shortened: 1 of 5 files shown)


leximayfield (approved):
Looks good, and is a vast improvement.

PR #721, opened 2022-07-23
[BUGFIX] Fix Horde Monsters getting crunched at full health from doors

### LINKED ISSUE #715: [BUG] Doors/Lifts/Crushers instantly delete Horde-spawned things
**Describe the bug**
It appears as though that anything that has the potential to restrict height lower than a Horde-spawned object's height will cause it to instantly disappear. You'll also notice that if it's a monster that gets deleted, the bar fills up.

10.1 Release

**To Reproduce**
Run the attached map and press the switch you spawn up against when things spawn. You'll see this applies to powerups, caches and monsters. The map shows a door in action, but I was able to reproduce it with crushers and lifts.

**Expected behavior**
Probably **not** insta-delete monsters.

Test WAD is attached.

**Additional context**
N/A
[horde-door.zip](https://github.com/odamex/odamex/files/9164718/horde-door.zip)

bcahue:
This was an interesting and funny bug. It was caused by monsters in horde being flagged with MF_DROPPED. PIT_ChangeSector looks for this item flag in order to check if it can crunch items. Doors crunching horde boxes is because it has MF_DROPPED (removing that would interfere with P_TouchSpecial I believe, and other items are crunched by doors, so it's okay in my mind) but doors crunching full HP monsters is not good.

I've fixed that by removing MF_DROPPED and I ensured the monsters won't respawn by other means, which was the entire reason they were flagged with MF_DROPPED.

Resolves #715.

### DIFF
--- a/common/p_hordespawn.cpp
+++ b/common/p_hordespawn.cpp
@@ -78,9 +78,6 @@ static AActor::AActorPtr SpawnMonster(hordeSpawn_t& spawn, const hordeRecipe_t& 
 	{
 		if (P_TestMobjLocation(mo))
 		{
-			// Don't respawn
-			mo->flags |= MF_DROPPED;
-
 			if (recipe.isBoss)
 			{
 				// Heavy is the head that wears the crown.


leximayfield (approved):
This game still finds ways to surprise me.  Nice catch.

PR #722, opened 2022-07-23
[FEATURE] Show horde boss projectiles as having the same effect as horde bosses

### LINKED ISSUE #697: [FEATURE] Horde boss projectiles to mimic boss's palette status
Not related to a problem. Would just be nice to differentiate horde boss projectiles from standard projectiles, helps a player triangulate where a boss is if it cannot be seen right away. Also allows players to threat assess if they *must* take damage from a source then they can run into a normal enemy projectile than a suped up boss projectile.

Boss projectiles to look like the palette swapped versions of their owners, to be gold, sparkling, etc.

None

**Additional context**
N/A

bcahue:
This is a simple change that applies the horde boss effect (yellow fountain with yellow transition table) to fired horde projectiles. Video below:

https://youtu.be/ikGJOLdlMFc

Resolves #697

### DIFF
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -553,8 +553,9 @@ static void CL_SpawnMobj(const odaproto::svc::SpawnMobj* msg)
 
 	// Light up the projectile if it came from a horde boss
 	// This is a hack because oflags are a hack.
-	if (mo->target && mo->target->oflags)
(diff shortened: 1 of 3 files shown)


leximayfield (approved):
LGTM, in both senses of the world.  Also, don't give me ideas about boss projectile buffs...

Issue #694, opened 2022-07-03
[FEATURE] Have command "hordenextwave" resurrect players fallen in the current wave


Xenaero:
**Is your feature request related to a problem? Please describe.**
If using the command **hordenextwave** during a live wave, players who had died prior to the command being issued do not regain their lives upon the new wave starting, and are left out!

**Describe the solution you'd like to see**
Players coming back to life. If you're advancing the wave, it makes sense to maintain the behavior of the last wave being completed.

**Describe alternatives you've considered**
I have not considered any alternatives.

**Additional context**
There is none.

electricbrass:
In a game with >1 max lives, should living players be given an extra life when hordenextwave is used?

electricbrass:
> In a game with >1 max lives, should living players be given an extra life when hordenextwave is used? I've got it working both ways, so can do a PR with whichever is preferred.

bcahue:
I believe its configured to give an extra life for players if g_lives is greater than current player lives, so it would make sense to mirror that functionality.

electricbrass:
Yes it is. Just wasn't sure if we wanted that for hordenextwave. I could see not wanting to give everyone lives.

rakohus:
Added to stable for 10.2 release in #726. Thanks @electricbrass

PR #726, opened 2022-07-24
Revive players on hordenextwave

electricbrass:
hordenextwave now functions the same as when a new wave starts normally: dead players are revived, queued players join, and living player gain an extra life

Addresses #694


leximayfield (approved):
LGTM!

Issue #538, opened 2022-02-06
[BUG] "Pumpfaking" Monsters


rakohus:
**Describe the bug**
Under new baseline netcode introduced in Odamex 10, monsters' position is generally more correct. However, monsters changing state will skip around and sometimes show their firing frame before they do. This is most obvious with Cyberdemons, then Revenants. Very visible in Horde mode. 

**Build that the bug occurred in**
Odamex 10

**To Reproduce**
Regularly occuring

**Expected behavior**
Minimize pumpfaking

**Screenshots, NetDemos, & Crash Dumps**
Netdemos/videos forthcoming.

rakohus:
Added to stable for 10.2 release in #727. Thanks @bcahue

PR #727, opened 2022-07-24
[BUGFIX] Workaround Pumpfaking Monsters

bcahue:
After a deep dive of the pump-faking monsters, I figured a cheap and easy win is to simply force the client to only honor missilestate/meleestate changes from the server. This does not scale well with a ton of ping, so its a simple workaround and not really a permanent fix, but the results speak for themselves for clients with average ping. Attached to this ticket are horde gameplay videos of this client against 10.1 Release servers.

https://youtu.be/lpsrdJ_iXro
https://www.youtube.com/watch?v=IikPTb7UwoQ
https://youtu.be/nTwIUA1AU94

Partially solves #538


leximayfield (approved):
Amazing catch!  LGTM.

PR #732, opened 2022-08-01
[BUGFIX] Show K/D instead of Wins on scoreboard for DM

bcahue:
After the recent scoreboard changes, I went ahead and fixed a simple check for a round game that was causing wins instead of K/D to be displayed in normal deathmatch modes.


leximayfield (approved):
Oops.   Thanks.

PR #737, opened 2022-08-06
[BUGFIX] Propagate Line Special Clearing

bcahue:
In Odamex, specials are cleared by the server and sent to clients primarily thru the `P_UpdateButtons(client_t*)` (on connect) and `SVC_Switch(line_t&, uint32_t, uint32_t)` (on demand) functions. This acts upon the `wastoggled` property of `line_t` to see if a linedef was triggered. The nomenclature is confusing, because you'd think only switches would be applicable here, but here it, in fact, updates all linedefs, and only applies switch logic to it if the texture is a switch.

This fixes the DOOM II MAP20 blue switch offline and online (line special wasn't cleared on the D1 immediately before the switch). This was fixed offline by removing logic for changed switches, as that logic isn't necessary as downstream logic will only apply switch logic to actual switch textures.

This also fixes DOOM II MAP27 online, when lowering the cyberdemon and spider mastermind, a 2nd player could come in and also lower the cyberdemon and spider mastermind to the floor, a map bug. This fixes that. Also fixed, are the cleared line specials sent to the client upon connect. Before, one time switches and such could be activated a 2nd time by a client if they were to reconnect. This is fixed.

### DIFF
--- a/common/doomdef.h
+++ b/common/doomdef.h
@@ -335,12 +335,6 @@ enum GameMission_t
 // Texture scrollers operate at a rate of x/64 units per tic.
 #define SCROLL_UNIT 64
 
-struct lineresult_s
-{
-	bool switchchanged;
-	bool lineexecuted;
-};
-
 // Index of the special effects (INVUL inverse) map.
 #define INVERSECOLORMAP 32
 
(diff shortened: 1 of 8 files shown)


leximayfield (approved):
LGTM.  There are some instances where I noticed some spurious `break` after `return` but unreachable break can always be cleaned up later.

PR #738, opened 2022-08-06
[BUGFIX] Clear secrets on clients when discovered by players other than the consoleplayer.

bcahue:
Does what it says on the tin. MapFormat changed the secret format to closely resemble Boom, but these changes weren't carried over when parsed by a client from a message by the server. This fixes that.


leximayfield (approved):
LGTM.

PR #740, opened 2022-08-07
[BUGFIX] Fix "fast" scrollers

bcahue:
This is a simple fix to scrollers. Before, scrollers would spawn 2 scrollers on the map if the map underwent a soft map reset, such as when maps count down to play time (ala Survival). Demo shown as an example, on SUNLUST.WAD MAP01 as a demonstration.

Odamex 10.2 RC2 (Old build)
https://youtu.be/p1MBoMyTA-I

Odamex 10.2 Stable (This patch)
https://youtu.be/uAR-mWr9SiY


leximayfield (approved):
LGTM.

PR #742, opened 2022-08-08
[BUGFIX] Restore Vanilla Demo Compatibility

bcahue:
The recent changes to propagate line special changes inadvertently broke vanilla demo compatibility. Doom II DEMO2 was the main culprit. Now, all the IWAD demos properly play, and I even improved our working demo count for OdaTests from 52 to 59. I also cleaned up the special line functions a bit.


leximayfield (approved):
LGTM

PR #745, opened 2022-08-10
[BUGFIX] Fix random bad lighting glow that would sometimes happen on soft map reset

bcahue:
I was afraid I would have to do this when I did the scroller fix, as lighting glow is transmitted the same way and has the same issues.

The reason why lighting glow would sometimes mess up (Odahorde map01 is a good example) online, was on map reset -- The map would 1.) not clear the previous thinkers, which wasn't a big deal for glow things as it just uses a sector as input, but the big deal was 2.) original light levels are not restored, even if the Glow thinker is destroyed, so when the thinker is created, it takes the minimum light level of an already in-progress glow, which can only be lower than the original light level.

This problem becomes very evident on SUNLUST MAP01, which has glow lights prominently on the first spawn. After a few failed attempts, the lights eventually stop glowing.

This is fixed by 1.) destroying light thinkers on map reset and 2.) restoring the original light level of every sector on map reset.

Also, I've re-added the null check for scroller control sectors. This has the side effect of not allowing sector 0 as a control sector, but I figure the alternative is that 10.2 clients simply won't scroll when playing on 10.1 servers, so the devil you know...

### DIFF
--- a/common/p_setup.cpp
+++ b/common/p_setup.cpp
@@ -97,6 +97,8 @@ line_t* 		lines;
 int 			numsides;
 side_t* 		sides;
 
+int*			originalLightLevels; // Needed for map resets
+
 // [RH] Set true if the map contains a BEHAVIOR lump
 bool			HasBehavior = false;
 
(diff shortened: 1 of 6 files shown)


common/p_setup.cpp
```diff
@@ -97,6 +97,8 @@ line_t* 		lines;
 int 			numsides;
 side_t* 		sides;
 
+int*			originalLightLevels; // Needed for map resets
```
leximayfield: Is there a reason why you used an `int*` and `new[]` here instead of ` std::vector`?  Also, I don't see a `delete[]` anywhere.

leximayfield (changes requested):

common/p_setup.cpp
bcahue: I redid this to use std::vector instead.

leximayfield (approved):
Perfect.  Thank you.

PR #747, opened 2022-08-12
[FEATURE] Add new Steam directories for the consolidated Doom store listings

bcahue:
Adds a few more Steam directories since Id software updated the directories it installs WADs to for Final Doom.


leximayfield (approved):
LGTM, thanks!

PR #749, opened 2022-08-14
[BUGFIX] Send line specials activated via ACS through the proper logic path

bcahue:
This was a bug discovered during Hordamex testing. Map36 has a bug where scrollers created via an Enter script would run at light speed because it was being applied twice. Map45 has an issue with a perpetual lift created via infinite looping ENTER script. The lift will shoot above where it's supposed to because it runs both clientside and serverside.

This fix ensures ENTER scripts happen on the server first, and then are propagated to clients via SV_ExecuteLineSpecial to ensure things run at the proper time, and don't run twice. LS_ACS_Execute already ensures it only runs on servers (and is only executed on line executions), but no check existed for OPEN/ENTER scripts.


leximayfield (approved):
Looks good, let's add it to the stable builds.

Issue #739, opened 2022-08-06
[BUG] Automap scrolling does not work with follow mode off


thefatwilly:
**Describe the bug**
When toggling 'follow mode' while in the Automap, attempting to move the cursor with the arrow keys will move the point to a (center of the map?) fixed location and not really budge at all.

**Build that the bug occurred in**
10.1 & 10.2 pre-release

**To Reproduce**
Just go into any map and use the automap to toggle follow mode off then try to move the map around

electricbrass:
Doesn't seem to be the center of the map. Seems consistently up and to the right of the center (So the map ends up more in the bottom left of the screen).
I'm pretty sure the issue is somewhere in AM_changeWindowLoc.

rakohus:
Fixed by @electricbrass patch. Thanks for report @thefatwilly

PR #750, opened 2022-08-16
[BUGFIX] Fix automap scrolling with follow mode off

electricbrass:
The map coordinates were being set to double the pan increment instead of being incremented by the pan increment. Now it works properly.

Addresses #739


leximayfield (approved):
Nice catch!  LGTM!

Issue #492, opened 2021-12-19
[BUG] PortMIDI is not handling pitch bends too well


Some1NamedNate:
**Describe the bug**
In MAP23 of TWANGO1.WAD, the guitar's bend is more intense than what the MIDI specified.

**Build that the bug occurred in**
0.9.5

**To Reproduce**
With "Music Backend" set to PortMIDI, load TWANGO1.WAD and warp to MAP23.

**Expected behavior**
The pitch bend should be handled correctly.

boomlinde:
The problem here is probably that the MIDI bend range was changed (perhaps in a song playing on another map). Ideally, I think bend range should be reset to +/-2 semitones between song changes. This can be done  with RPN message 0, i.e. by setting RPN address to 0 with CC#100 and CC#101, and then setting the value to two in CC#6.

Other parameters should also be reset. In the best of worlds, all controllers, but I don't know what they all initialize to. But a common problem in Doom ports is that they leave the modulation wheel unchanged between map changes, so sometimes you get crazy vibratos where you shouldn't have them. This should be reset by setting CC#1 to 0.

Various MIDI modules also have reset functions available via SysEx. For SC-55 for example, you can send it the SysEx packet sequence "F0 41 10 42 12 40 00 7F 00 41 F7" (hex encoded) to make it reset all controllers and settings and set up for GM mode. I think allowing users to specify a SysEx sequence to run between each song change would be a great feature. I wonder what Doom does...

Some1NamedNate:
A reset command in between MIDI's (Map -> Intermission -> Map, etc.) would be nice.

rakohus:
Fixed by @ceski-1 patch. Thanks for report @Some1NamedNate

PR #754, opened 2022-08-19
fix portmidi pitch bend, volume, panning, and effects

ceski-1:
**Pitch bend:**
- **Bug:** Follow the steps described [here](https://github.com/coelckers/prboom-plus/issues/101) for Three is A Crowd. The music on map03 changes the pitch bend range to 12 semitones from a default of 2. The changed value carries over to map10 causing the music to sound strange.
- **Fix:** Set the pitch bend range back to 2 whenever the music changes.

**Pan, delay (detune), and reverb:**
- **Bug:** Download [Sunlust](https://www.doomworld.com/forum/topic/68089-sunlust-on-idgames/). Launch the game with `odamex.exe -file sunlust.wad -warp 1` and listen to the music. Now launch the game with `odamex.exe -file sunlust.wad`, start a new game manually, and listen again. The celesta (ch. 1 and 2) echoes badly and the slap bass (ch. 5) is positioned far right. Both have too much reverb. This is due to the pan, delay, and reverb values carrying over from the title screen music.
- **Fix:** Set pan to 64 (center), delay to 0, and reverb to 40 whenever the music changes.

**Notes:**
- Fixes for volume, chorus, and other effects are also included.
- Fixes incorrect control numbers in i_midi.h (0x3 is undefined).
- All values follow [official midi specifications](https://www.midi.org/specifications).
- Should also address [#492](https://github.com/odamex/odamex/issues/492)


rakohus:
I tested this and functionally, it's good. Awaiting code review. Thanks for the patch @ceski-1!

leximayfield (approved):
Wow, this is good stuff.  I don't know enough about the inner workings of MIDI to know correctness, but the code looks okay, so in it goes!  Thanks!

PR #760, opened 2022-08-25
[BUGFIX] Fix MacOS build workflow

bcahue:
When testing builds on my local fork, I noticed macOS builds failing. This is because Homebrew has updated SDL to 2.24.0, which causes compiling issues with Odamex currently (https://formulae.brew.sh/api/formula/sdl2.json).

This is fixed by using a (hacky) way to ensure 2.0.22 is used for Odamex builds.

### DIFF
--- a/ci/macos-buildgen.sh
+++ b/ci/macos-buildgen.sh
@@ -8,7 +8,9 @@ IFS=$'\n\t'
 set -x
 
 # Install packages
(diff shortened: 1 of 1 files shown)


leximayfield:
I'll pull this for now, but I suspect the real problem is something stupid like "SDL header isn't right" or "missing define causes half the file to not be included." - not anything about 2.0.24 specifically that causes issues.

PR #761, opened 2022-08-26
[FEATURE] DoomBattleZone v10.2.0-DBZ1 😈

ghost:
This is a modification to odamex that adds zoning plus
several new features and improvements:

1. Zoning - supported through new console commands
2. BattleMode - client requests server address
3. HUD markup - server controlled client UI
4. Level & spectator timers
5. Fixed width font support
6. Respawning barrels! 💥
7. Developer improvements

To try odamex in battle mode, visit https://doombattle.zone/.


leximayfield:
@DoomCoder Hi there.

We appreciate the fact that you've spent a lot of time on these patches, however I'm not sure I can accept this PR in it's current form.  There's two reasons, one practical, the other philosophical.

Practically, this is a gigantic patch that adds many features at once, that was simply PR'ed without prior discussion.  We would've preferred a discussion beforehand and a more piecemeal "This feature looks cool, this other thing could've been done differently" with separate PR's for each feature.

Philosophically, we don't like the idea of zoning between servers in a server cluster.  I think I would've been more receptive to the idea a decade ago, but today, we think that large server clusters should be in the past tense.  Instead, we think that allowing players to more easily run their own servers (so players can play whatever they want) or to allow players to configure remote servers (so you don't need large server clusters) is the way forward.

ghost:
@AlexMax, Thank you for the detailed response. Your feedback is very reasonable. My apologies on dropping the bomb with the large commit/PR. It was rushed as we wanted to have the code available to Odamex when our site opened last week.

On the practical front, I can split things up but it sounds like on the philosophical side, the key feature of zoning is not in the future plans for Odamex.

If you have any interest in any of the changes we have now or in the future, reach out to me or submit an issue to our [repo](https://github.com/DoomBattleZone).

The work you & the Odamex team is doing is amazing. It is the best choice of client/server DOOM with a great future. If you ever reconsider, we definitely want to be part of Odamex. Classic DOOM online at [Odamex.net](https://odamex.net/) or clustered DOOM online at [DoomBattle.Zone](https://doombattle.zone/) or at future sites to come.

Issue #764, opened 2022-08-27
[BUG] Tag 667 Arachnotron Script Does Not Execute


thefatwilly:
**Describe the bug**
Map07 specific tag 667 will not execute in singleplayer or online mode

**Build that the bug occurred in**
10.2.0 Stable Online and in Singleplayer

**To Reproduce**
Play Doom2 Map07 and kill all arachnotrons

**Expected behavior**
Floor tagged "667" should raise when all arachnotrons are dead

PR #769, opened 2022-08-29
[BUGFIX] Fix Bossactions (667 not working, 666 causes crashes, etc)

### LINKED ISSUE #766: [BUG] Tag 666 causes Oda to crash, unknown how to recreate
**Describe the bug**
Sometimes on Doom 2 Map07 when executing Tag 666 this will cause Odamex to crash suddenly.

10.2.0 Stable Singleplayer so far but this causes a client crash likely leaving me to believe this is possible in online

**To Reproduce**
Unknown as no one method has consistently caused this to crash. Only constant is that in every crash, I opened the console and input some commands such as 'bind' 'alias' and random inputs

**Expected behavior**
To not crash

Test wad
[Map07testwad.zip](https://github.com/odamex/odamex/files/9438487/Map07testwad.zip)

Crash dmps
[Map07crashdmps.zip](https://github.com/odamex/odamex/files/9438488/Map07crashdmps.zip)

bcahue:
A simple change to prevent bossactions from triggering switch texture commands, which would occur on bunk lines and cause a crash sometimes. Also fixed are bossspecials not returning false on `P_UseSpecialLine` if it couldn't find a special. Resolves #766 and #764.

Seems like a scary, vanilla breaking change. I'm happy to report OdaTests reports 30 broken demos instead of 32, so I actually increased the compatible demos using this change (which seems normal, considering Cross line bossspecials weren't working or would crash Odamex...)

### DIFF
--- a/common/p_boomfspec.cpp
+++ b/common/p_boomfspec.cpp
@@ -1793,6 +1793,8 @@ void P_SpawnCompatiblePusher(line_t* l)
 bool P_UseCompatibleSpecialLine(AActor* thing, line_t* line, int side,
                                         bool bossaction)
 {
+	bool reuse = false;
+	bool trigger = false; // used for bossactions
 	// e6y
 	// b.m. side test was broken in boom201
 	if (side) // jff 6/1/98 fix inadvertent deletion of side test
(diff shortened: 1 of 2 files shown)


leximayfield (approved):
LGTM.  Thanks!

PR #770, opened 2022-08-29
[FEATURE] Add PWAD select tab to boot window

electricbrass:
Adds a tab to the client boot window. The left column displays `.wad`, `.deh`, and `.bex` files found in the directories specified in the resource locations tab. Recognized IWADs are excluded from the list. Files selected in this column show up in the right column. The buttons to the right move the selected file up and down, to change load order.

image


rakohus:
I was unable to test this. When using the autobuild in github actions, the client runs but never loads or shows the launch ui.

electricbrass:
It works for me. I just tried the same and I get the ui. Are you using a existing config where "Don't show this again" has been selected maybe?

leximayfield:
This looks great as far as it goes, functionality-wise.  I have not code-reviewed it yet.  However, one thing I can immediately recommend is adding a "delete WAD" to the right hand drop-down, because having to go back and uncheck the WAD to get rid of it is kind of a pain.

> I was unable to test this. When using the autobuild in github actions, the client runs but never loads or shows the launch ui.

This might be because the launcher is trying to index WAD files.  We do this anyway to find the IWAD files we support, but if we need to do the same thing for all WAD files period we might want to add a progress bar that shows this happening.  Or, even better, load the main launcher without scanning PWADs, but start scanning when the interface loads.

Also, there's the question of if scanning all WAD files is even a good idea - there's a reason why ZDL loads so quickly, and that's because it leaves the full directory scan to the "Select WAD" dialog box.

electricbrass:
I can add the delete WAD option sometime soon hopefully, currently very busy with school. 

> I have not code-reviewed it yet.

I'm sure there will be some code issues since I am still very much not a C++ programmer.

> We need to do the same thing for all WAD files period we might want to add a progress bar that shows this happening. Or, even better, load the main launcher without scanning PWADs, but start scanning when the interface loads.

Here you mean when loading the PWAD interface?
I agree that we don't want the main launcher loading slowly because of indexing all the files. I considered doing it the same as ZDL, it's what I personally use and would likely continue using, but I also think there is something to consider for having all your wads in one list in the ui instead of having to search for and add them every time. For general users, I think being able to add the directory once and just having everything there is easier. And ZDL then is an alternative for those who want that ui.

Also maybe have a progress bar *and* wait to load PWADs til selecting that tab? That way there isn't also the possibility of user thinking it has frozen if it takes a bit to load when changing to that tab.

leximayfield:
Better, but for me, there's still a noticable pause when selecting the new tab as it indexes my entire WAD files directory, with no visual feedback.

electricbrass:
I'm planning on adding a progress bar (and speeding up the process if I can). I've just been very busy.

leximayfield:
With commit 1237b7d8 I was able to speed things up significantly.

rakohus:
PWAD selector tab crashes for me as of `45f0418`. Dump attached. 
[odamex_g45f0_11084_20221025T205817.zip](https://github.com/odamex/odamex/files/9864186/odamex_g45f0_11084_20221025T205817.zip)

leximayfield:
> PWAD selector tab crashes for me as of `45f0418`. Dump attached. [odamex_g45f0_11084_20221025T205817.zip](https://github.com/odamex/odamex/files/9864186/odamex_g45f0_11084_20221025T205817.zip)

I believe my most recent commit has fixed that.

Issue #768, opened 2022-08-27
[BUG] Cannot change install location in official 10.2.0 Windows installer


stewboy:
**Describe the bug**
When running the 10.2.0 installer on Windows, the installer offered no option to change the install path. This was a problem for me because the default install location was on a drive letter that I no longer had a drive mapped to. On trying to install there, I got the error 'The drive or UNC share you selected does not exist or is not accessible. Please select another.' The installer then quit, offering no option to change the drive or path. I had to temporarily map a drive to that letter to install it, then change it back - I only noticed just now that I also had the option to download a .zip file instead.

**Build that the bug occurred in**
10.2.0

**To Reproduce**
Run the installer.

**Expected behavior**
I expected at some point to be able to change the install location.

rakohus:
Hi @stewboy, are you saying you don't see this browse button next to the path? I just tested with both user-level and admin and am able to select path in both instances.

![image](https://user-images.githubusercontent.com/22926877/187587832-2389d4e6-64ce-44f0-bb92-aa9911e0bf0e.png)

![image](https://user-images.githubusercontent.com/22926877/187587935-56bb41af-5164-4bb5-a90d-92fc40d5a3b8.png)

leximayfield:
This occurs when you are upgrading from a previously installed version of Odamex.  It will upgrade to the same location and also keep the same type of install (user-permission Programs vs admin-permission Program Files).

If you want to change the location or installation type, please uninstall Odamex, then reinstall it.  This is a limitation of InnoSetup and although I've hacked on the installer before, I don't know that you can bypass that limitation easily.

bcahue:
In current `stable`, the Inno Setup installer will prompt if Odamex is installed and notify the user how to change the path in the case of Odamex already being installed.

In this case, however, we encounter a common problem with Inno Setup (and NSIS and Windows Installer, for that matter) -- if you install to a removable drive then remove that drive, you can no longer run an installer for that product unless you run the uninstaller in appwiz.cpl (Add or Remove Programs). These uninstallers are supposed to not fail if it can't find a file.

With that, I'm closing this.

PR #772, opened 2022-09-03
[FEATURE] Installer: Prompt for downgrading and prompt for uninstall if Odamex is already installed

bcahue:
This change to the Inno Setup file is a pascal script that first checks if the version of Odamex you are installing is a downgrade to an already installed version of Odamex, and another prompt to ask the user if they want to uninstall the current version of Odamex before installing the new one, which would in effect free up the user to install Odamex to a different directory if wanted, or to change the install to a User install rather than a Computer-wide install, or vice versa.

Since it would notify the user what to do to change the path, I would consider this as solving #768 

The script is skipped if the install is marked /SILENT or /VERYSILENT.

Attached is an example installer created using this script. It's of a fictitious (at the moment) version of 10.3.0. You can play around with the logic by changing the `DisplayVersion` in `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\{2E517BBB-916F-4AB6-80E0-D4A292513F7A}_is1`

[Odamex-Installer.zip](https://github.com/odamex/odamex/files/9483265/Odamex-Installer.zip)


leximayfield:
I like the idea, however I have two issues with it.

First off, I don't get the "select install mode" screen after uninstalling Odamex.  That probably has to be fixed.

image

Secondly, I think the message is unclear and front-loads the idea of uninstalling Odamex, which really should only be done in exceptional circumstances.  There's also no way to cancel the installation at that point.

Try this copy on for size:

> Odamex has been detected on this machine.  Would you like to do an in-place upgrade?
> 
> If you want to upgrade your copy of Odamex in-place, press "Yes."  This is the suggested choice.
> If you want to do a fresh installation of Odamex, in order to change your installation directory or permissionss, press "No."
> If you want to exit the installation, press "Cancel."
>
> [Yes} [No] [Cancel]

bcahue:
@AlexMax 
> I like the idea, however I have two issues with it.
> 
> First off, I don't get the "select install mode" screen after uninstalling Odamex. That probably has to be fixed.
> 
> image

I will likely need to just simply exit after uninstall and the user should simply launch the installer again to get the system-wide/user-wide prompt. I am not able to control this, as far as I can tell.

> Secondly, I think the message is unclear and front-loads the idea of uninstalling Odamex, which really should only be done in exceptional circumstances. There's also no way to cancel the installation at that point.
> 
> Try this copy on for size:
> 
> > Odamex has been detected on this machine.  Would you like to do an in-place upgrade?
> > If you want to upgrade your copy of Odamex in-place, press "Yes."  This is the suggested choice.
> > If you want to do a fresh installation of Odamex, in order to change your installation directory or permissionss, press "No."
> > If you want to exit the installation, press "Cancel."
> > [Yes} [No] [Cancel]

Agreed, this prompt is doable and does make more sense in a flow sense. I'll update when these are implemented.

fuzztooth (approved):

PR #775, opened 2022-09-08
Relax SDL2 minor version checking

electricbrass:
With the change in the sdl2 version numbering scheme, minor version changes no longer imply an break in backwards compatibility. This relaxes the sdl version check, and reverts the changes to the macOS workflow now that they aren't need.


leximayfield:
LGTM

Issue #786, opened 2022-09-25
[BUG] Boom Format Line Action 192 Does the Opposite


thefatwilly:
**Describe the bug**
In boom the line special of 192 which is "SR Light Change to Brightest Adjacent" currently does the exact opposite of the action and turns the sector brightness down to an indeterminate amount.

**Build that the bug occurred in**
10.2.0 Stable currently offline unknown if possible in online

**Expected behavior**
For sector to become bright relative to an adjacent brighter sector

**Additional context**
Test Wad attached, careful as it has multiple testing related areas as well.
[testing wad.zip](https://github.com/odamex/odamex/files/9641797/testing.wad.zip)

bcahue:
This can be closed I believe

fuzztooth:
Fix merged with no additional feedback.  Closing.

PR #788, opened 2022-09-25
[BUGFIX] Fix Boom Format Line 192

electricbrass:
EV_LightTurnOn only checks brightness of nearby sectors if the value passed in for bright is less than 0. 0 was being passed in, no searching would be done, and then the brightness would be set to 0.

Addresses #786


leximayfield (approved):
Thanks!  LGTM!

Issue #637, opened 2022-03-24
[FEATURE] Flash task on OS when a minimized player should be aware of an in-game event


rakohus:
**Is your feature request related to a problem? Please describe.**
Players often minimize the game window during periods where their play is suspended. Examples include, but are not limited to, being dead in a survival mode and waiting for their next attempt, waiting for a competitive game to start, etc. I have observed this player behavior in both Horde and CTF modes recently.

**Describe the solution you'd like to see**
Flash the taskbar under the following conditions:

1. The Odamex client window is not in focus (minimized); AND
2. When ANY of the following game events occur:
a. The player is spawned in-game and a `restart` is passed by the server;
b. The player is spawned in-game and the pre-game countdown begins;
c. The player is brought back to life during any survival mode;

**Describe alternatives you've considered**
We have tinkered with some sound cues before i.e. Horde respawn sound. However, in the midst of a lot of action or background noise, it is not always noticable. Some OS, which support [SDL_FlashWindow](https://wiki.libsdl.org/SDL_FlashWindow) may not play in-game sounds while the window is not focused as well. 

**Additional context**
[SDL_FlashWindow](https://wiki.libsdl.org/SDL_FlashWindow) was implemented in SDL 2.0.16. The function seems pretty simple. May want to be careful of error handling when a negative error code is returned as we don't want Odamex client bombing out over this feature.

bcahue:
Was fixed in #790

PR #790, opened 2022-10-04
[FEATURE] Flash window on respawn

electricbrass:
Flashes the taskbar any time the player is spawned in-game and the window is not focused.

Address #637

### DIFF
--- a/client/sdl/i_video_sdl20.cpp
+++ b/client/sdl/i_video_sdl20.cpp
@@ -719,7 +719,7 @@ bool ISDL20Window::isFocused() const
 //
 // Flashes the taskbar icon if the window is not focused
 //
-void ISDL20Window::flashWindow() {
+void ISDL20Window::flashWindow() const {
 	if (!this->isFocused())
 		SDL_FlashWindow(mSDLWindow, SDL_FLASH_UNTIL_FOCUSED);
 }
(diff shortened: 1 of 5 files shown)


client/sdl/i_video_sdl20.h
```diff
@@ -171,6 +171,10 @@ class ISDL20Window : public IWindow
 
 	virtual bool isFocused() const;
 
+	#if defined(SDL2016)
+	virtual void flashWindow() const;
```
leximayfield, line 174: I don't really like the idea of changing the shape of the class based on the SDL version.  I think a better approach would be for this method to always exist, but ifdef out the actual implementation in the method definition.

client/sdl/i_video_sdl20.cpp
```diff
@@ -712,6 +712,19 @@ bool ISDL20Window::isFocused() const
...
+#if defined(SDL2016)
+
+//
+// ISDL20Window:flashWindow
+//
+// Flashes the taskbar icon if the window is not focused
+//
+void ISDL20Window::flashWindow() const {
+	if (!this->isFocused())
+		SDL_FlashWindow(mSDLWindow, SDL_FLASH_UNTIL_FOCUSED);
```
leximayfield, line 723: I don't really like the idea of changing the shape of the class based on the SDL version. I think a better approach would be to move the `#ifdef` pair to be inside of the `ISDL20Window::flashWindow()` implementation.

leximayfield (changes requested):
I like the change, but I think it would be cleaner to only require a single ifdef at inside the method definition itself.

electricbrass:
I tried just one but did it wrong and so added the 2nd. I agree this is better.

leximayfield (approved):
Nice!  LGTM.

Issue #759, opened 2022-08-24
[FEATURE] Tell the user they have a different version of the IWAD than the server


rakohus:
**Is your feature request related to a problem? Please describe.**
During the leadup to a decino event, many users were reporting the `Refusing to download commercial IWAD file.`. They often had version 1.666 from a CD. 

**Describe the solution you'd like to see**
Improve the messaging here. If the IWAD they have is valid and we know the hash, tell them their version and that the server is using version x. 

**Additional context**
![image](https://user-images.githubusercontent.com/22926877/186537959-7672b47f-9244-4e00-8509-38451c52c4de.png)

drivingspectrum:
I believe this is now fixed? I downgraded doom2.wad to 1.666 and connected to a server, it told me to use Omniscient to repatch the iwad

PR #791, opened 2022-10-09
Show better error when Odamex can't find IWAD

leximayfield:
This PR gives a more in-depth reason when a user doesn't have the correct version of a commercial IWAD.  It also cuts off any attempt to even start the download process in the first place.

Additionally, I came across two other issues that I fixed:
- During testing, I found that the Unity versions of the various Doom games were not a part of the IWAD database.  This has been addressed.
- If the game was in `GS_LEVEL` state when the level data had been unloaded, Odamex would still attempt to render the level.  This specifically came up in the context of disconnecting after the server switched to an IWAD the client doesn't have, but I can't imagine this is the only circumstance when it happened.

Addresses #759.

### DIFF
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -1717,6 +1717,14 @@ bool CL_PrepareConnect()
 	}
 	else if (!ok && !missingfiles.empty() || cl_forcedownload)
 	{
+		if (::missingCommercialIWAD)
+		{
+			Printf(PRINT_WARNING,
+			       "Server requires commercial IWAD that was don't have.\n");
+			CL_QuitNetGame(NQ_ABORT);
+			return false;
+		}
+
 		OWantFile missing_file;
 		if (missingfiles.empty())				// cl_forcedownload
 		{
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -1720,7 +1720,7 @@ bool CL_PrepareConnect()
 		if (::missingCommercialIWAD)
 		{
 			Printf(PRINT_WARNING,
-			       "Server requires commercial IWAD that was don't have.\n");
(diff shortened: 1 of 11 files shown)


client/src/cl_main.cpp
```diff
@@ -1717,6 +1717,14 @@ bool CL_PrepareConnect()
 	}
 	else if (!ok && !missingfiles.empty() || cl_forcedownload)
 	{
+		if (::missingCommercialIWAD)
+		{
+			Printf(PRINT_WARNING,
+			       "Server requires commercial IWAD that was don't have.\n");
```
bcahue: "Server requires commercial IWAD that was don't have."

Should this be: "Server requires commercial IWAD that **we** don't have." ?

client/src/cl_parse.cpp
```diff
@@ -801,6 +801,13 @@ static void CL_LoadMap(const odaproto::svc::LoadMap* msg)
 
 	if (!missingfiles.empty())
 	{
+		if (::missingCommercialIWAD)
+		{
+			Printf(PRINT_WARNING, "Server requires commercial IWAD that was don't have.\n");
```
bcahue: "Server requires commercial IWAD that was don't have."

Should this be: "Server requires commercial IWAD that **we** don't have." ?

electricbrass: I would think it should be "Server requires commercial IWAD that **you** don't have." or "Server requires commercial IWAD that **wasn't found**."

client/src/cl_main.cpp
leximayfield: Good eye.  Fixed.

client/src/cl_parse.cpp
leximayfield: Fixed.

PR #793, opened 2022-10-23
[BUGFIX] Lower BIGFONT comma offset

electricbrass:
The comma in the BIGFONT had it's offset set to 0,0 so it appeared like a comma. I lowered it to 0,-6 so that it lines up with the bottom of the text

Before:
image

After:
image


electricbrass:
Ah oops, has the pwad select stuff

PR #800, opened 2022-11-04
[BUGFIX] Fix initial spawn weapon desync/missing spawninventory and introduce client replay system

### LINKED ISSUE #386: [BUG] Players telefrag each other on `callvote restart`
**Describe the bug**
Players telefrag each other when respawning after a "restart" callvote.

Odamex 0.9.2

**To Reproduce**
Callvote restart.

**Expected behavior**
Green spawn fog should appear on first spawn and players should not spawn in the same spots (when it can be avoided).

### LINKED ISSUE #647: [BUG] Weapon pickup desync, and server not getting firing signal from client
**Describe the bug**
SSG in round 51 is not registered as being picked up by the client, and rocket launcher fires on client but not on the server. SSG in round 52 appears to do damage despite not a proper firing frame and forces switch to pistol on client.

10.0.0 (G98D97-8033)

**To Reproduce**
I have no idea, seems random. Only occurred on the first spawn of both rounds, and only on that map

**Expected behavior**
No desyncs

See attached, and view Rounds 51 and 52. Problem described occurs within the first engagement of both rounds.

**Additional context**
N/A
[lazaruswepdesync.zip](https://github.com/odamex/odamex/files/8407744/lazaruswepdesync.zip)

bcahue:
As you may have followed the saga on #oda-development, I have finished the work for initial map load/reset map weapon desyncs. This involved keeping reborns BEFORE PlayerInfo commands. I've also added a ClientReplay system that runs every tic for weapon pickups that run on the client but the item hasn't arrived yet. In the future, we can just send `mobjtype_t` with the svc_touchspecial message, and use the replay system for item destruction instead.

Also included are updates to `infomap.cpp` used for debugging information.

Should fix #647, and *might* fix #386.


leximayfield (approved):
Really nice idea.  LGTM.

PR #801, opened 2022-11-07
[BUGFIX] Item Replay/Weapon Correction Fixes

bcahue:
I had previously tried to fix the edge case of an item replay and a weapon correction firing at the same time, causing the weapon to switch to itself after firing. But in doing this, I caused a desync in the scenario you advance a map in coop and keep your ready weapon out in the next map. This fixes that bug, and properly accounts for the edge case of an item replay occurring the same tic or before a weapon correction, triggered by firing immediately before your spawn weapon arrives, which seems quite common in such maps with large amounts of items such as SmartCTF2 MAP12.

Also included is a provision to only "play" the tics from the last ready state so the client matches up with what the server sees.


leximayfield (approved):
Looks good, nice catch.

PR #802, opened 2022-11-10
[BUGFIX] Fix numpad usage with and without Numlock

### LINKED ISSUE #380: [BUG] Odamex detecting AltGr as a chat macro
**Describe the bug**
Whenever you press AltGr (in AZERTY keyboards) + 0-9 after opening the chat, it erases my whole sentence and replaces it with the Doom chat macros.

Odamex 0.9.1 (Build GC96C-1) _(and earlier down to 0.8.x)_

**To Reproduce**
_-For AZERTY/Other keyboards that have the AltGr key_: open the chat then press AltGr + 0-9
_-For QWERTY keyboards_: open the chat then press Ctrl+Alt (look at _Additional Context_ for more explanation) + 0-9

**Expected behavior**
Since this button is used as a shortcut for special characters, I was expecting it to type out a character corresponding to its position between 0-9 (~ # { [ | ` \ ^ @).

https://user-images.githubusercontent.com/84753774/119567150-dc672c00-bda3-11eb-9799-d4e319e70753.mp4

 **__Notes:__** Since the keyboard overlay shows characters instead, here are the corresponding numbers:
_& = 1_ / _é = 2_ / _" = 3_ / _' = 4_ / _( = 5_ / _- = 6_ / _è = 7_ / _ = 8 / _ç = 9_ / _à = 0_

**Additional context**
After running a keyboard overlay, I noticed that the AltGr key is actually a shortcut for Ctrl+Alt. So you might be able to replicate this issue on QWERTY keyboards with those two keys.

### LINKED ISSUE #799: [CHANGE] Change console to use numpad based on how numlock has been toggled
While using the console: if numlock is off, numpad keys 2 4 6 8 behave as the arrow keys, which is expected. 0 , 1 3 7 9 do nothing. If numlock is on, 2 4 6 8 still behave as if numlock were off, while 0 , 1 3 5 7 9 behave as expected. 
Note: due to regional keyboard reasons, mine has a comma instead of a dot on the numpad.

If numlock is enabled, I want all the number keys to actually give me numbers. If numlock is off, I want them to have their alternate behaviour. I am requesting the change because the current behaviour is flat out confusing and weird and means I can't actually use the numpad to input numbers while using the console, which is the only time I'd use it while having the console open in the first place. Also, page up, page down, and delete do have uses for the console, so if somebody wants to use the numpad with numlock off for the console then they would likely want those to function as expected, too. I can't see a use case for having this current weird hybrid behaviour.

**If applicable, describe alternatives you've considered**
I guess if somebody in the dev team is dead set on using numpad 2 4 6 8 in the console instead of arrow keys while having numlock enabled and refuses this change, then just make it behave like numlock was off for all the keys and not just those, regardless of how it's toggled. At least that way it will be less misleading to others, and you'll get page up/down and delete for the numpad while at it.

bcahue:
This was achieved by using SDL's mod system for determining what key combinations were used with keys, and use that to determine their final function. Previously, Odamex would handle this functionality, which worked well for keys you have to keep pressing for the mod to work, but does not with mods like capslock, scrolllock and numlock.

To keep everything consistent, I've converted the Shift and Ctrl functionality for console to this mod method, and converted the options menu to be able to use numpad as well.

Fixes #799 and #380. I also added functionality for chat macros -- they can now be performed on numpad keys if numlock is on.

These changes are only performed on the console, chat macros and the options menu. The rest of the key code is intact.

### DIFF
--- a/client/sdl/i_input.cpp
+++ b/client/sdl/i_input.cpp
@@ -803,7 +803,8 @@ static int I_GetEventRepeaterKey(const event_t* ev)
 	int button = ev->data1;
 	if (button == OKEY_CAPSLOCK || button == OKEY_SCRLCK ||
 		button == OKEY_LSHIFT || button == OKEY_LCTRL || button == OKEY_LALT ||
-		button == OKEY_RSHIFT || button == OKEY_RCTRL || button == OKEY_RALT)
+		button == OKEY_RSHIFT || button == OKEY_RCTRL || button == OKEY_RALT ||
+		button == OKEY_NUMLOCK)
 		return 0;
 	else if (button >= OKEY_HAT1 && button <= OKEY_HAT8)
 		return button;
--- a/client/sdl/i_input_sdl12.cpp
+++ b/client/sdl/i_input_sdl12.cpp
@@ -279,6 +279,8 @@ void ISDL12KeyboardInputDevice::gatherEvents()
 					ev.data2 = ev.data3 = '\r';
 				else if ((sdl_ev.key.keysym.unicode & 0xFF80) == 0)
 					ev.data2 = ev.data3 = sdl_ev.key.keysym.unicode;
+
+				ev.mod = sdl_ev.key.keysym.mod;
 				
 				if (ev.data1)
 					mEvents.push(ev);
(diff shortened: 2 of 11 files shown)


leximayfield (approved):
Looks good.

PR #803, opened 2022-11-11
[BUGFIX] Fix latched cvar strings in console to show pending value instead of broken text

### LINKED ISSUE #798: [BUG] Changing a non-boolean latched cvar and checking what it will change to doesn't show the new value in console
**Describe the bug**
If you change a non-boolean latched cvar in console and then check what it will be changed to, the console doesn't print the new value.

Odamex v10.2.0

**To Reproduce**
Open console, change a non-boolean latched cvar, check what it will be changed to before it's had a chance to be changed.

**Expected behavior**
Typing in the cvar again should actually show what value it will be changed to for non-boolean cvars.

https://puu.sh/JqwZJ/fce7970219.png
https://puu.sh/Jqx3b/67ee397165.png

**Additional context**

bcahue:
Due to how strings were expressed for latched cvar pending change retrieval, they'd create broken strings every time. I fixed this behavior by making the pending change a string from the beginning.

Old behavior:
image

New behavior: 
image

Fixes #798

### DIFF
--- a/common/c_cvars.cpp
+++ b/common/c_cvars.cpp
@@ -612,12 +612,12 @@ static std::string C_GetLatchedValueString(const cvar_t* var)
 		return C_GetValueString(var);
 
 	if (var->flags() & CVAR_NOENABLEDISABLE)
-		return '"' + var->latched() + '"';
+		return '"' + std::string(var->latched()) + '"';
 
 	if (atof(var->latched()) == 0.0f)
 		return "disabled";
 	else
-		return "enabled";	
+		return "enabled";
 }
 
 BEGIN_COMMAND (set)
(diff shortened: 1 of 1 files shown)


common/c_cvars.cpp
```diff
@@ -612,12 +612,12 @@ static std::string C_GetLatchedValueString(const cvar_t* var)
 		return C_GetValueString(var);
 
 	if (var->flags() & CVAR_NOENABLEDISABLE)
-		return '"' + var->latched() + '"';
+		return '"' + std::string(var->latched()) + '"';
```
leximayfield: Yeah, chars are actually considered numeric types according to C.

I am still not a fan of what's going on here, though.  I think this would be much better off as a `StrFormat` instead of doing string concatenation.

leximayfield (changes requested):

common/c_cvars.cpp
bcahue: Done, changed to use StrFormat and verified functionality:

image

leximayfield (approved):
LGTM

Issue #321, opened 2021-04-27
[FEATURE] More console shortcuts


Ch0wW:
**Is your feature request related to a problem? Please describe.**
Odamex's console is quite good the way it is right now ; however, compared to other Quake-Engine games or ports, I noticed we miss some useful shortcuts.  

**Describe the solution you'd like to see**
Add a few more console shortcuts, such as the following:
- ALT/CTRL + BackSpace = delete the last word
- ALT + DEL = delete the word before the cursor

rakohus:
Could be good to have.

bcahue:
This was resolved with #804 but I forgot to close this ticket.

PR #804, opened 2022-11-12
[FEATURE] Add ALT+DEL and CTRL+Backspace to delete words to the right and left of the cursor in console, respectively.

bcahue:
This feature resolves issue #321, and includes more shortcuts to manipulate text in the console. Alt+Del will delete the word to the immediate right of the cursor, while Ctrl+Backspace will delete the word on the left of the cursor. Ctrl+Backspace has logic to delete an extended whitespace or the word, but not both. Also included is a fix to ConsoleCommandLine::moveCursorLeftWord(). It now treats a body of whitespace as a word that can be skipped, rather than moved one cursor length at a time.

With these changes, a user who is privy to shortcut usage can manipulate text in the console as fast as any other text input in the operating system.

This also contains code for PR #802 and should be merged after that one is.

### DIFF
--- a/client/src/c_console.cpp
+++ b/client/src/c_console.cpp
@@ -360,7 +360,7 @@ void ConsoleCommandLine::moveCursorLeftWord()
 
 	const char* str = text.c_str();
 
-	if (str[cursor_position - 1] == ' ')
+	if (cursor_position > 0 && isspace(str[cursor_position - 1]))
 	{
 		spaceWord = true;
 	}
@@ -369,18 +369,18 @@ void ConsoleCommandLine::moveCursorLeftWord()
 	{
 		if (spaceWord)
 		{
-			if (str[cursor_position - 1] != ' ')
+			if (!isspace(str[cursor_position - 1]))
 			{
 				break;
 			}
 		}
 		else
 		{
-			if (firstSpacesCleared && str[cursor_position] == ' ')
+			if (firstSpacesCleared && isspace(str[cursor_position]))
 			{
 				break;
 			}
-			else if (str[cursor_position] != ' ' && !firstSpacesCleared)
+			else if (!isspace(str[cursor_position]) && !firstSpacesCleared)
 			{
 				firstSpacesCleared = true;
 			}
@@ -458,7 +458,7 @@ void ConsoleCommandLine::deleteLeftWord()
 	const char* str = text.c_str();
 
 	// Delete trailing spaces instead of the word
-	if (str[cursor_position - 1] == ' ')
+	if (cursor_position > 0 && isspace(str[cursor_position - 1]))
 	{
 		spaceWord = true;
 	}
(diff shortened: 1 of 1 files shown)


client/src/c_console.cpp
```diff
@@ -352,12 +355,39 @@ void ConsoleCommandLine::moveCursorRight()
...
 {
-	if (cursor_position > 0)
-		cursor_position--;
+	bool firstSpacesCleared = false;
+	bool spaceWord = false;
 
 	const char* str = text.c_str();
-	while (cursor_position > 0 && str[cursor_position - 1] != ' ')
+
+	if (str[cursor_position - 1] == ' ')
```
leximayfield: What happens if `cursor_position` is 0?

Also, I'm not super sure it matters for the console, but in general checking for whitespace is done with `isspace` or some other white-space detection function instead of hadrcoding 0x20.

client/src/c_console.cpp
```diff
@@ -418,6 +448,84 @@ void ConsoleCommandLine::deleteCharacter()
...
+void ConsoleCommandLine::deleteLeftWord()
+{
+	size_t word = 0;
+	bool firstSpacesCleared = false;
+	bool spaceWord = false;
+
+	const char* str = text.c_str();
+
+	// Delete trailing spaces instead of the word
+	if (str[cursor_position - 1] == ' ')
```
leximayfield: Again, what happens if `cursor_position` is 0 here?

client/src/c_console.cpp
```diff
@@ -418,6 +448,84 @@ void ConsoleCommandLine::deleteCharacter()
...
+	doScrolling();
+}
+
+void ConsoleCommandLine::deleteRightWord()
+{
+	bool spaceWord = false;
+
+	const char* str = text.c_str();
+
+	if (str[cursor_position] == ' ')
```
leximayfield: What happens if `cursor_position` is equal to the length of the string?

leximayfield (changes requested):

client/src/c_console.cpp
bcahue: Changed to use `isspace()` and check `cursor_position` for all of these.

client/src/c_console.cpp
bcahue: Changed to use `isspace()` and check `cursor_position` for all of these.

client/src/c_console.cpp
bcahue: Changed to use `isspace()` and check `cursor_position` for all of these.

client/src/c_console.cpp
bcahue: Also check bounds here to make sure `cursor_position` is less than `text.length()`

leximayfield (approved):
LGTM

PR #805, opened 2022-11-15
[BUGFIX] Improve MBF21 obituary messages and toasts.

### LINKED ISSUE #784: [BUG] Obituaries w/ MBF21 projectiles are confusing
**Describe the bug**
This is a bug via perception. Possibly better considered a change, but players consider it a bug. 

Obituaries in the kill feed don't work well with MBF21 projectiles that come from a player's weapon. They appear the same way a suicide would. See: VesperDM-RC2.wad and the screenshots below.

10.2

**To Reproduce**
Play VesperDM-RC2, kill or be killed via a projectile weapon other than the rocket launcher. 

**Expected behavior**
Absent a larger feature request for custom obituaries for new projectiles, there should at least be a standard message for MBF21 projectile kills in both the kill feed and classic obituaries.

i.e.

Killfeed: ```Ralphis [generic kill icon] AlexMax```
Text obits: ```Ralphis killed AlexMax.```

Current behavior when killing another player with an MBF21 projectile:

image

image

bcahue:
This change functionally resolves #784, and adds MBF21 obituary messages and toasts so users always know who killed them with custom PVP weapons, and also, Odamex will now determine means of death for most MBF21 monsters and display it. Barring a complete system for promoting MOD messages and icons, this should get us to that point.

The only hiccup with this is A_Explode and A_Detonate work by making the source of the explosion mo->target, which will be the player that it's targeting if the A_Explode or A_Detonate is on a frame that doesn't belong to a projectile. This makes death mods from those functions a suicide. I can't change this without messing up vanilla compat.

Odamex_DM_20221115_025912


leximayfield (approved):
Nice!

PR #807, opened 2022-11-29
Don't include d.WAD in the PWAD select tab

electricbrass:
Installations of The Ultimate Doom from Steam include a file called d.WAD in the same directory as doom.wad. It is an exact copy of doom.wad, but because of the different filename, it currently gets considered a PWAD and is added to the PWAD select. This is likely to confuse users who are uncertain of what this file is or where it is, and when loaded it is difficult to realize at first what is happening. This change checks the MD5 hash of a file ONLY if it has the name "d.WAD" and does not include it if it matches a known IWAD hash.

### DIFF
--- a/common/m_resfile.cpp
+++ b/common/m_resfile.cpp
@@ -397,7 +397,14 @@ std::vector<scannedPWAD_t> M_ScanPWADs()
 				continue;
 
 			OWantFile file;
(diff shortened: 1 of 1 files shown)


leximayfield (approved):
Looks good to me

PR #816, opened 2022-12-16
Hack to make the PlayerSetup hack keep track of the menu stack.

Acts19quiz:
The following is a pull request to address the issue where the "hacked" way of getting PlayerSetup to work with Odamex's menu doesn't keep track of the menu stack. Specifically, this is to address the bug where upon entering "PLAYER SETUP", exiting the menu immediately goes directly back to the game, not the previous OPTIONS menu.

This introduces a check for whether or not PSetupDef is active, and a boolean to catch differing situations from reaching PLAYER SETUP from Escape or from F4.

### DIFF
--- a/client/src/m_menu.cpp
+++ b/client/src/m_menu.cpp
@@ -112,7 +112,7 @@ short				whichSkull; 		// which skull to draw
 bool				drawSkull;			// [RH] don't always draw skull
 
 // hack for PlayerSetup
-bool				F4menu;
+int					PSetupDepth;
 
 // graphic name of skulls
 char				skullName[2][9] = {"M_SKULL1", "M_SKULL2"};
@@ -510,7 +510,7 @@ BEGIN_COMMAND (menu_main)
     S_Sound (CHAN_INTERFACE, "switches/normbutn", 1, ATTN_NONE);
 	M_StartControlPanel ();
 	M_SetupNextMenu (&MainDef);
-	F4menu = false;
+	PSetupDepth = 2;
 }
 END_COMMAND (menu_main)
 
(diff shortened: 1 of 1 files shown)


ghost:
That is an annoying bug. We also fixed it with a different approach which is a little simpler. Here's the patch for reference:
[player-menu-fix-patch.txt](https://github.com/odamex/odamex/files/10246739/player-menu-fix-patch.txt)
The full source is in a fork here: https://github.com/DoomBattleZone/odamex/tree/doombattlezone

Acts19quiz:
I searched issues and pull requests to make well sure I wasn't doing something already done. Oh well. I guess the devs can do whatever with this. I need to touch up the formatting anyway (the etiquette on single-action if blocks in the code is to not use brackets, which I did, probably because I shuffled a few lines around and forgot to remove them).

AlexMax:
@Acts19quiz Found a bug in this version.

If you bind a key to `menu_options`, press it, and then hit escape, you will immediately exit the menu stack instead of going to the main menu.  However, if you bind a key to `menu_player` and then press it, pressing Escape will hit the Options menu, and not back you out of the menus completely.

That's the only issue I found - the normal route from Options to Player Setup seems to work just fine.

leximayfield (changes requested):
@Acts19quiz Found a bug in this version.

If you bind a key to `menu_options`, press it, and then hit escape, you will immediately exit the menu stack instead of going to the main menu.  However, if you bind a key to `menu_player` and then press it, pressing Escape will hit the Options menu, and not back you out of the menus completely.

That's the only issue I found - the normal route from Options to Player Setup seems to work just fine.

Acts19quiz:
I changed the bool to an int to catch more means of entry into the menu.

leximayfield (approved):
LGTM!  Thanks for the patch!

PR #817, opened 2022-12-19
Rewrite PortMidi driver

ceski-1:
Volume fixes:
- Set to default (scaled) before each song
- Set to MIDI spec default on shutdown
- Updated whenever song changes volume
- Updated whenever song has a SysEx reset
- Controlled by MSB value only (LSB ignored)

Reset device between songs:
- SysEx: GM, GS (default), or XG
- Non-SysEx: reset commonly used controllers only
- Reset type selectable via menu option or CVAR: `snd_midireset` (0: None, 1: GM, 2: GS (default), 3: XG)
- Optional delay after reset for hardware devices, CVAR only: `snd_mididelay` (0 (default) to 2000 milliseconds)

Read SysEx events from songs:
- Process complete messages only (`0xF0 ... 0xF7`)
- Ignore escape sequences (no illegal characters sent to device)
- Ignore multi-packet or malformed messages
- Process SysEx resets (e.g. switch from GS to XG as needed)

Instrument fallback support (aka "capital tone fallback" emulation)
- Prevent wrong or silent notes when song uses unsupported instruments
- Emulated at an SC-55 level in GS mode
- Useful for Roland-based synths (MS GS Synth (Windows default), SCVA, SC-55mkII, etc.)
- Disabled when song contains a SysEx reset
- Toggle via CVAR: `snd_midifallback` (0: Disable, 1: Enable (default))

General music system clean up:
- Easier to extend MIDI support in the future (e.g. FluidSynth)


ceski-1:
Some examples:

XG MIDI songs play correctly (Alien Vendetta MAP24):

https://user-images.githubusercontent.com/56656010/208369250-6af2ec18-71a4-41c1-902d-bda8f892fab6.mp4

Instrument fallback for MIDIs that use incorrect instrument banks (e.g. MAYhem19 title screen). This is especially important for the default Windows MIDI device (Microsoft GS Wavetable Synth) in GS mode:

https://user-images.githubusercontent.com/56656010/208369272-044fcee3-ad06-4d65-88b4-d1bac2d46a73.mp4

Delay after reset option for hardware MIDI devices (real Roland SC-55 playing DOOM.WAD):

https://user-images.githubusercontent.com/56656010/208369293-5660ac74-b723-4f8e-8969-ffcf6404bd48.mp4

Menu option for reset type:

menu_option

fuzztooth (dismissed):
Generally looks ok to me assuming there aren't better ways to handle the init stuff.

I'll let @AlexMax weigh in.

ceski-1:
It looks like the PortMidi library has some custom changes? I would recommend updating it to a more current build.

leximayfield:
@ceski-1 I haven't looked at the PR quite yet, but I recall looking at the portmidi discrepancies in the past and seeing that they were mostly changes to make it compatible with 64-bit architectures (which I believe have been upstreamed), as well as changes to the CMake file to allow only compiling the one library we need instead of the whole enchilada.  There might be more, but those two were the ones I remember off the top of my head.

EDIT: Okay yeah the last time I looked at this portmidi was mostly unmaintained, I was unaware that recent updates had been made.  Adding portmidi as a submodule seems like the right play.

ceski-1:
@AlexMax I built Odamex with the latest Portmidi master offline and it works great. I'll add back the submodule change ~~but some of the build checks will fail until some compiling tweaks are applied.~~ Edit: Good to go now.

fuzztooth (approved):

PR #818, opened 2022-12-25
PortMidi: Ignore SysEx messages and use GM mode by default

ceski-1:
~~This is a small patch to ensure that instrument fallback isn't auto-disabled when the user's MIDI device is the default Windows synth (Microsoft GS Wavetable Synth).~~ Edit: See post below.

An easy test for this is DBP37_AUGZEN.wad at the title screen or MAP22.

### DIFF
--- a/client/sdl/i_musicsystem_portmidi.cpp
+++ b/client/sdl/i_musicsystem_portmidi.cpp
@@ -65,8 +65,24 @@ PortMidiMusicSystem::PortMidiMusicSystem()
 		return;
 	}
 
-	m_outputDevice = Pm_GetDefaultOutputDeviceID();
 	std::string prefdevicename(snd_musicdevice.cstring());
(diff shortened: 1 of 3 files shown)


ceski-1:
On second thought, I'm going to drop the platform specific hacks, disable SysEx from MIDI files by default, and set the default mode to "GM." This is the most conservative configuration that will be transparent to users when they upgrade to the next version of Odamex. Everything will play correctly this way (even songs that have historically had problems) at the expense of any advanced GS/XG features that a song may contain. Power users are welcome to dive into the config (or change the "MIDI reset" menu option) at the risk of occasionally exposing the issues of poorly written/ripped songs. My apologies for the back and forth.

fuzztooth (approved):
Looks good.  Thank you very much for your contributions!

Issue #322, opened 2021-04-27
[FEATURE] Vanilla Player Color Presets


OpenRift412:
I think it would be kind of cool to see the ability to have player color presets in the player settings menu, akin to how the Eternity Engine does it. 

Here's an example of what I mean:
![image](https://user-images.githubusercontent.com/29780788/116296779-9e58f700-a768-11eb-94b7-38efd63d9bd9.png)

Ch0wW:
That is actually a great idea, and would be a great complement on another "feature restoration" that is getting back the classic intermission screen! The idea to get a preset of the 4 playercolors and a 5th one that would be "custom" is actually something that shouldn't be too troublesome to make.

Acts19quiz:
With 10.4.0, this can now be closed as done. Thanks everyone! (:

OpenRift412:
> With 10.4.0, this can now be closed as done. Thanks everyone! (:

I was actually taking a look at this, and the player 2, 3, and 4 colors don't seem to match up properly with their vanilla counterparts. 

Here are some comparisons between Odamex and Zandronum, which has vanilla-accurate colors.

**Indigo**
Zandronum/Vanilla:
![image](https://github.com/odamex/odamex/assets/29780788/e7135528-ad8c-48fd-9c5b-b7d8a37b298f)
Odamex:
![image](https://github.com/odamex/odamex/assets/29780788/079a1913-f66a-4328-b527-f7f6455c6933)

**Brown**
Zandronum/Vanilla:
![image](https://github.com/odamex/odamex/assets/29780788/76a08eb9-eed0-46e6-a496-3543a6a91cfc)
Odamex:
![image](https://github.com/odamex/odamex/assets/29780788/1a405335-dca7-4022-a676-152ececa49d5)

**Red**
Zandronum/Vanilla:
![image](https://github.com/odamex/odamex/assets/29780788/37ab0205-62a9-47f4-a75c-5edd547d6dcc)
Odamex:
![image](https://github.com/odamex/odamex/assets/29780788/7b5cc083-ff6d-49b1-b744-bd19a2d1b2d1)

Acts19quiz:
Ah, yes. That. There was a conversation about this in the Discord server when I first started this. Some copy/paste:

I have tweaked them some, though outright Vanilla matching might not be possible as-is.

[...]

PR #819, opened 2022-12-28
Add to PLAYER SETUP the ability to switch between custom and preset colors. [FEATURE] #322

Acts19quiz:
In response to [FEATURE] request #322, here is the PLAYER SETUP menu rigged up with the ability to switch between current CUSTOM COLOR, as well as five select presets. This commit introduces changes to four files and two new CVARS, cl_customcolor and cl_colorpreset. cl_colorpresets is unchanged. cl_colorpreset keeps track of where players have that option set, and cl_customcolor saves the slider positions for RED, GREEN and BLUE when a preset color is chosen.

If this is setup correctly, these two new CVARs are stored locally only. The server in netplay should never see them since only cl_color itself is relevant.

I'm open to changes to some of the color presets given there's some wiggle room to try to matchup Odamex's RGB values to Vanilla DOOM's indexed palette.

Included is a short video.

https://user-images.githubusercontent.com/63941237/209765647-33f84af1-fed8-4449-abec-6fe8cd4fe1c0.mp4

### DIFF
--- a/client/src/m_menu.cpp
+++ b/client/src/m_menu.cpp
@@ -178,8 +178,8 @@ static void M_EditPlayerName (int choice);
 static void M_PlayerNameChanged (int choice);
 static void M_ChangeGender (int choice);
 static void M_ChangeAutoAim (int choice);
-static void M_ChangeColorPreset (int choice);// Acts 19 quiz
-static void SendNewColor (int red, int green, int blue);// Acts 19 quiz
+static void M_ChangeColorPreset (int choice);
+static void SendNewColor (int red, int green, int blue);
 static void M_SlidePlayerRed (int choice);
 static void M_SlidePlayerGreen (int choice);
 static void M_SlidePlayerBlue (int choice);
@@ -339,7 +339,7 @@ enum psetup_t
 	playerteam,
 	playersex,
 	playeraim,
-	playercolorpreset,// Acts 19 quiz
+	playercolorpreset,
 	playerred,
 	playergreen,
 	playerblue,
@@ -352,7 +352,7 @@ oldmenuitem_t PlayerSetupMenu[] =
 	{ 2,"", M_ChangeTeam, 'T' },
 	{ 2,"", M_ChangeGender, 'E' },
 	{ 2,"", M_ChangeAutoAim, 'A' },
-    { 2,"", M_ChangeColorPreset, 'C' },// Acts 19 quiz
+    { 2,"", M_ChangeColorPreset, 'C' },
 	{ 2,"", M_SlidePlayerRed, 'R' },
 	{ 2,"", M_SlidePlayerGreen, 'G' },
 	{ 2,"", M_SlidePlayerBlue, 'B' }
@@ -1272,7 +1272,8 @@ void M_QuitDOOM(int choice)
 void M_DrawSlider(int x, int y, float leftval, float rightval, float cur, float step);
 
 static const char *genders[3] = { "male", "female", "cyborg" };
-static const char *colorpresets[11] = { "custom", "blue", "indigo", "green", "brown", "red", "gold", "jungle green", "purple", "white", "black" };// Acts 19 quiz the order must match d_netinf.h
+// Acts 19 quiz the order must match d_netinf.h
+static const char *colorpresets[11] = { "custom", "blue", "indigo", "green", "brown", "red", "gold", "jungle green", "purple", "white", "black" };
 static state_t *PlayerState;
 static int PlayerTics;
 argb_t CL_GetPlayerColor(player_t*);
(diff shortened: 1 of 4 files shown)


leximayfield:
This might be a minor nitpick, but it's pretty universal that the "Grey" player is called "Indigo" in several places by Doom itself.  [Source](https://doomwiki.org/wiki/Player), also [source](https://github.com/chocolate-doom/chocolate-doom/blob/e55356d93a1408ca9db9a8cc57ec243f9fc8f99a/src/doom/d_englsh.h#L310).  Might be a better name for it.

Also, it would be nice if there was a nice round number of additional colors that cover different ranges.  Colors like Gold, Jungle Green, Pitch-black and Purple are under-represented.

Acts19quiz:
There was some back-and-froth in the Discord on how to handle differences between the way Vanilla achieved palette changes with palette range changes, and how Odamex uses HEX RGB values. For those colours mentioned, do you have recommendations, or should I just pick something and ship it out?

The Indigo thing is also news to me. I legit was thinking that would be some dark blue. It seems weird to pick Indigo and get a gray/grey colour in response, but if that's Vanilla behaviour and that's what you want, I can put that change in.

leximayfield:
> There was some back-and-froth in the Discord on how to handle differences between the way Vanilla achieved palette changes with palette range changes, and how Odamex uses HEX RGB values. For those colours mentioned, do you have recommendations, or should I just pick something and ship it out?

I'd say pick RGB colors that you think look good.  I just thought it'd be nice to have more colors than just the basic four plus blue.  If you need guidance, Doom Legacy and Eternity Engine might have some worth borrowing.

> It seems weird to pick Indigo and get a gray/grey colour in response, but if that's Vanilla behaviour and that's what you want, I can put that change in.

I'll leave that up to you.  "Gray" is probably more widely understood, but the original game did name it "Indigo" and if you're intending to fill out the four vanilla colors to me it makes sense to name it the same thing.

Acts19quiz:
I'm not sure what Jungle Green is supposed to look like, so I just took a guess.

https://user-images.githubusercontent.com/63941237/212980012-958631e6-ae4f-4517-8a65-65bf60874456.mp4

client/src/m_menu.cpp
```diff
@@ -1638,10 +1706,17 @@ static void M_PlayerTeamChanged (int choice)
...
 static void SendNewColor(int red, int green, int blue)
 {
-	char command[24];
+	char colorcommand[24];
+	char customcolorcommand[30];// Acts 19 quiz
+	int colorpreset = D_ColorPreset(cl_colorpreset.cstring());// Acts 19 quiz
 
-	sprintf(command, "cl_color \"%02x %02x %02x\"", red, green, blue);
-	AddCommandString(command);
+	sprintf(colorcommand, "cl_color \"%02x %02x %02x\"", red, green, blue);
```
leximayfield: So, as a rule, if you need to change a line that has a `sprintf` in it, please change it to something that's not `sprintf`.  The function is incredibly dangerous and there are safer alternatives, both standard and specific to our codebase.

In this instance, there is no reason for the buffer to be fixed-length, so we have a function called `StrFormat` that operates on `std::string` that I would recommend switching to.

client/src/m_menu.cpp
```diff
@@ -1638,10 +1706,17 @@ static void M_PlayerTeamChanged (int choice)
...
+	AddCommandString(colorcommand);
+	if (colorpreset == COLOR_CUSTOM)// Acts 19 quiz
+	{
+		sprintf(customcolorcommand, "cl_customcolor \"%02x %02x %02x\"", red, green, blue);
```
leximayfield: Ditto.

leximayfield (changes requested):
I like it, as far as it goes.  Only thing I would change would be to replace two instances of `sprintf` that I see and you're good.

client/src/m_menu.cpp
Acts19quiz: I added the one sprintf command because that was how the then-color command was being sent to the console (I renamed command to colorcommand since there would be two commands, and added customcolorcommand). If I'm understanding the request correctly, you want me to change not only my addition, but also how the standard color command is currently being handled since I'm already monkeying with that. (that's what I went with, anyway)

Acts19quiz: Uh oh, I got the same Mac error that electricbass got on #836. Did I foul something up, or is something else going on?

electricbrass: Something else is going on, I got that on my fork just from syncing it.

leximayfield (approved):

leximayfield:
The mac build is failing for reasons other than your code.

PR #820, opened 2022-12-28
Make Odamex translate the Chex Quest smallfont (STCFN***).

Acts19quiz:
The way Odamex works with index palette entries for translating pixel colours causes hiccups sometimes. Specifically, with Chex Quest. All of the menu functions get lost in a sea of green, and while it's easy to suggest that 'Chex text should always be green', that works only in a Vanilla EXE vacuum. With ZDoom-style expanded menus, mono-colour text works a whole lot worse.

Included are two screenshots. The first shows the way it is now. Note that both the option and choice are the same colour. Worse, the distinct colours for the chosen message colours don't even show their true colours. It really is kind of a mess for Chex Quest. ZDoom/GZDoom/Zandronum outright disables Chex Quest 3's custom textcolo specifically for this reason.

This patch addresses this issue--at least for Chex Quest. HacX, another Odamex supported megawad, suffers the same issue. However, unlike Chex Quest's green colour, the blue colour used by HacX is split up across the palette. E.g., while there are still 16 primary colours like with DOOM and Chex Quest, HacX has them split across a palette range--because that's how DOOM's palette does it. It's possible a bigger expansion of the current system could address this, but that's adding code upon code. Wholesale changes may be in order to fix HacX.
DOOM_translate
Chex_translate

### DIFF
--- a/client/src/r_draw.cpp
+++ b/client/src/r_draw.cpp
@@ -216,10 +216,14 @@ byte bosstable[256];
 
 static void R_BuildFontTranslation(int color_num, argb_t start_color, argb_t end_color)
 {
-	const palindex_t start_index = 0xB0;
-	const palindex_t end_index = 0xBF;
 	const palindex_t chexstart_index = 0x70;
 	const palindex_t chexend_index = 0x7F;
+	const palindex_t hacxstart_index = 0xC3;
+	const palindex_t hacxmid1_index = 0xCF;
+	const palindex_t hacxmid2_index = 0xF0;
+	const palindex_t hacxend_index = 0xF2;
+	const palindex_t start_index = 0xB0;
+	const palindex_t end_index = 0xBF;
 	const int index_range = end_index - start_index + 1;
 
 	palindex_t* dest = (palindex_t*)Ranges + color_num * 256;
@@ -229,6 +233,15 @@ static void R_BuildFontTranslation(int color_num, argb_t start_color, argb_t end
 		for (int index = 0; index < chexstart_index; index++)
 			dest[index] = index;
 		for (int index = chexend_index + 1; index < 256; index++)
+			dest[index] = index;
+	}
+	else if (gamemission == commercial_hacx)
+	{
+		for (int index = 0; index < hacxstart_index; index++)
+			dest[index] = index;
+		for (int index = hacxmid1_index + 1; index < hacxmid2_index; index++)
+			dest[index] = index;
+		for (int index = hacxend_index + 1; index < 256; index++)
 			dest[index] = index;
 	}
 	else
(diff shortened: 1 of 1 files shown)


rakohus:
This looks good visually, so great work on that. However, for the sake of discussion, I'd like to float the idea of whether or not Odamex should offer its own small menu font out-of-the-box. This would eliminate the need for corner cases like this, but has some obvious drawbacks. 

Pros:
- Will work with any custom wad that alters the small font (for instance, any doom pwad that changes the font color)
- Uniform design and look
Cons:
- Less "doomy", etc.
- Requires the creation of nice assets

The further logical step in the "pro" camp would be to implement something akin to the TTF support that @seanmleonard prototyped like a decade ago.

Acts19quiz:
GZDoom took this approach--it junked the entire Vanilla menu font and went for its own Unicode fonts. The primary reason for this was alternate languages. I like what you're floating, but that's also a really big change beyond the scope of this PR, something the primary devs would need to pass back-and-forth to determine what direction to take.

HacX is probably the biggest victim here. It's a recognized and supported IWAD, but the scattered nature of blue in the Vanilla palette index makes the current setup not work well at all. This is at least a quick and easy stopgap until the menu gets entirely reworked with a custom smallfont system.

client/src/r_draw.cpp
```diff
@@ -218,31 +218,61 @@ static void R_BuildFontTranslation(int color_num, argb_t start_color, argb_t end
 {
 	const palindex_t start_index = 0xB0;
 	const palindex_t end_index = 0xBF;
+	const palindex_t chexstart_index = 0x70;
+	const palindex_t chexend_index = 0x7F;
```
fuzztooth: Is this the kind of information that can be tied to the gamemode configurations?

Acts19quiz: Possibly. I didn't look into any part of the code related to that other than to see how the current IWAD is determined. I changed only the one file to keep it as simple as possible. If there are other files you want to outsource some of the alterations/checks to in addition to r_draw.cpp, or even instead of, I'll look into seeing if I can do it.

Acts19quiz:
After revisiting https://github.com/odamex/odamex/pull/820 and racking my brain a bit, I did find a way to make the current r_draw.cpp setup translate the HacX small font in menus. It does bloat things a bit, and I agree that some sort of larger scope change to the way menu fonts and palette translations are handled would be beneficial for reasons already mentioned. However, the HacX menu does currently not look good, and is in need of immediate improvement.

A note about the difference in "BLUE" from old to new: This is not an error in picking up the blue colours or my Rube Goldberg math falling apart. Rather, this is a known issue where the RGB translation used isn't an exact match of the original. Instead of going for light blue, the RGB translator targets pure blue. This is the same across DOOM and Chex Quest.

HacX_STCFN_old
HacX_STCFN_new

leximayfield (approved):
Looks better to me.  Good enough to merge, at least for now.

PR #822, opened 2023-01-04
Make the server console a little nicer to use on windows

drivingspectrum:
Basically this PR does the following:
* Disable QuickEdit mode on Windows, which if the user were to (accidentally) select text in the server console, would cause the whole server to stop running until the text was un-selected.
* Prevent unusual (control) characters from entering the console like arrow keys.
* Proper Ctrl-C detection to tell the server to exit gracefully (this never worked properly, it should now)


server/src/i_main.cpp
```diff
@@ -114,14 +114,20 @@ int __cdecl main(int argc, char *argv[])
...
     {
-        // Handle ctrl-c, close box, shutdown and logoff events
+        // Handle close box, shutdown and logoff events
         if (!(hEvent = CreateEvent(NULL, FALSE, FALSE, NULL)))
             throw CDoomError("Could not create console control event!\n");
 
         if (!SetConsoleCtrlHandler(ConsoleHandlerRoutine, TRUE))
             throw CDoomError("Could not set console control handler!\n");
 
-        #ifdef _WIN32
```
fuzztooth, line 124: My apologies for my ignorance here, but was the removal of this ifdef intentional?

drivingspectrum: Sorry yes I should have clarified this more in the commit, as on line 94 there is already a #ifdef _WIN32 block protecting the windows main function.

Issue #812, opened 2022-12-11
[BUG] The obituaries specific to the amount of floor damage (-5, -10, -20) aren't used, instead using OB_DEFAULT.


Acts19quiz:
**Describe the bug**
The different obituaries for different types of damaging floors (OB_SLIME, OB_LAVA) aren't used. Instead, OB_DEFAULT is used for them all.

**Build that the bug occurred in**
10.3.0 GB28EE-8578

**To Reproduce**
In multiplayer, go into different types of damaging or killing floor. They will all display OB_DEFAULT ("%o died.").

**Expected behavior**
For the correct obituary string to display for its intended floor type.

**Screenshots, NetDemos, & Crash Dumps**
[Here is a test WAD](https://github.com/odamex/odamex/files/10201327/FloorDamageTest.zip) for DOOM II to try out some different types of floors. Since offline single-player doesn't display obituaries to console anyway, these will need to be tested online, possibly in a LAN configuration. The console will need to be checked since Odamex's short-form graphical obits take the place of the old-style HUD messages.

**Additional context**
The floor obituaries haven't worked correctly in quite awhile. In testing, the current iteration has been happening since Odamex 10, and there was also some issues going on in Odamex 0.9. Maybe even earlier.

electricbrass:
Which obituaries go with which damage amounts? I think I have a fix for this but don't know which ones go with which. There's a "Player melted." and a "Player mutated."

Acts19quiz:
I don't know if I can help with that one. I put my test WAD into ZDoom 2.8.1, ZDaemon 1.10.20, Zandronum 3.1, and GZDoom 4.10.0. All of them showed nothing but "Mutated". That makes it a bit hard for me to discern what is supposed to be what.

rakohus:
Confirmed fixed. Thank you @electricbrass!

PR #836, opened 2023-01-25
[BUGFIX] Restore floor damage text obituaries

electricbrass:
Addresses #812 

Death by damaging sectors now displays "%o mutated." instead of "%o died."
This matches the behavior of other ZDoom based ports.


leximayfield (approved):
Nice catch!

PR #837, opened 2023-01-26
PortMidi: Send "reset all controllers" when a song loops

ceski-1:
Matches Vanilla Doom behavior and prevents hanging notes (NIN.WAD E1M8). Also removed an unused variable.


leximayfield (approved):
Looks good!  These MIDI fixes are really appreciated.

PR #844, opened 2023-02-07
[BUGFIX] Display levelnames specified in UMAPINFO on intermission screens

electricbrass:
This fixes two related bugs. Level names specified in UMAPINFO were not being used in intermissions, instead using the Doom 2 names. Additionally, level names specified in MAPINFO were not being shown on the ENTERING intermission screens, just leaving a blank space. Both of these are addressed.


leximayfield (approved):
LGTM!

PR #848, opened 2023-03-30
PortMidi: Update volume after "reset all controllers" event

ceski-1:
This is a workaround for a bug that appears when using PortMidi, Windows, and the default Windows MIDI synth (Microsoft GS Wavetable Synth). When the MS GS Synth receives a "reset all controllers" event it also resets the channel volume, deviating from MIDI spec. As a result, in certain cases the channel can be audible even if the music volume slider is set to zero. The workaround is to always update the volume after a "reset all controllers" event.

Test wad: [reset_all_ctrls_test.zip](https://github.com/odamex/odamex/files/11107650/reset_all_ctrls_test.zip)


leximayfield (approved):
LGTM, thank you

PR #849, opened 2023-04-19
master: include time.h on all platforms

Jan200101:
`time()` is only declared i `time.h`, I don't know what `sys/time.h` was intended for, but it compiles without it on Fedora 38


leximayfield:
Green lights across the board - nice catch.

PR #851, opened 2023-05-02
[BUGFIX] Fix ghost monsters that occur with MBF21 use of A_HealChase

bcahue:
A_HealChase was allowing MBF21 monsters to resurrect monsters clientside, which resulted in desynced "ghost" monsters that never attacked. This fixes that.


leximayfield (approved):
LGTM

PR #860, opened 2023-05-26
Add label support to UMAPINFO

DeathEgg:
Adds support for the `label` key to UMAPINFO. The `author` key is also read in for both UMAPINFO and MAPINFO, but not displayed yet. In addition, UMAPINFO is now split into its own file to make it easier to edit. Some functionality was merged into OScanner to make this possible, but only functions that could be useful with serialization elsewhere.


bcahue (approved):
Looks good to me. Tested and verified. Refactoring the umapinfo stuff into a different file is much liked as well.

PR #864, opened 2023-05-29
Improve messaging in Horde and Survival coop modes.

rakohus:
Small commit to improve messaging in Horde/Survival modes upon round and game end conditions. Please double check my code because, although it's very simple, I'm still learning best practices on where to place things.

### DIFF
--- a/client/src/st_new.cpp
+++ b/client/src/st_new.cpp
@@ -1368,9 +1368,14 @@ void LevelStateHUD()
 		break;
 	}
 	case LevelState::ENDROUND_COUNTDOWN: {
-		StrFormat(lines.title,
-		          "Round " TEXTCOLOR_YELLOW "%d " TEXTCOLOR_GREY "complete\n",
-		          ::levelstate.getRound());
+		if (G_IsCoopGame() | G_IsHordeMode())
+			StrFormat(lines.title,
+				"Attempt " TEXTCOLOR_YELLOW "%d " TEXTCOLOR_GREY "complete\n",
+				::levelstate.getRound());
+		else
+			StrFormat(lines.title,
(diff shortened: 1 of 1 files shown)


client/src/st_new.cpp
```diff
@@ -1368,9 +1368,14 @@ void LevelStateHUD()
 		break;
 	}
 	case LevelState::ENDROUND_COUNTDOWN: {
-		StrFormat(lines.title,
-		          "Round " TEXTCOLOR_YELLOW "%d " TEXTCOLOR_GREY "complete\n",
-		          ::levelstate.getRound());
+		if (G_IsCoopGame() | G_IsHordeMode())
```
bcahue: should be || instead of |

client/src/st_new.cpp
```diff
@@ -1381,15 +1386,27 @@ void LevelStateHUD()
 		else if (win.type == WinInfo::WIN_TEAM)
 			StrFormat(lines.subtitle[0], "%s team wins the round",
 			          WinToColorString(win).c_str());
+		else if (G_IsCoopGame() | G_IsHordeMode())
```
bcahue: should be || instead of |

bcahue (changes requested):
Also, please wrap your if statements in curly braces.

client/src/st_new.cpp
bcahue: should be || instead of |

client/src/st_new.cpp
bcahue: should be || instead of |

bcahue (changes requested):

bcahue (approved):

bcahue (approved):

Issue #861, opened 2023-05-27
[BUG] Weapons can be moved around on a conveyor when touching them


rakohus:
**Describe the bug**
When weapons (and possibly other items, although they disappear when picked up) are touched by a player, they can move items from their intended position. The player serves as a "blocker" of sorts, changing their trajectory. 

ODDLY ENOUGH, this behavior can even occur OFFLINE, where the weapons do not stay. The weapon can hesitate in front of the player and not pick up. 

**Build that the bug occurred in**
10.3

**To Reproduce**
Play wheeltest.wad, wait for the weapons to fan out on the wheel. Stand in front of the weapon while it is moving and press against it. It will react to the player thing and can be moved around the wheel.

**Expected behavior**
Player does not affect the weapon's trajectory. Whether or not they pick it up, their existence doesn't alter its movement.

In DSDA Doom, with out of the box settings, the weapons ALWAYS come right to player, they are picked up, and then they continue on their expected route.

**Screenshots, NetDemos, & Crash Dumps**

[wheeltest.zip](https://github.com/odamex/odamex/files/11580812/wheeltest.zip)

rakohus:
[box2.zip](https://github.com/odamex/odamex/files/11593175/box2.zip)

Hobomaster created the attached debug map

rakohus:
More demonstrative items. 
Odamex:
https://github.com/odamex/odamex/assets/22926877/ea57b3ef-f1f2-4776-b605-c6076147d0ff

https://github.com/odamex/odamex/assets/22926877/e5487ceb-6947-4259-a5ba-f7b74225f11e

PR #868, opened 2023-06-09
Fix pickup items getting caught on players when pushed by conveyors

rakohus:
Implement fix for #861. Add a check at the end of PIT_CheckThing that allows the items to pass through the player. Thanks to @nstlaurent for the research and insight on this!


rakohus:
@bcahue developed a better fix

PR #869, opened 2023-06-10
[BUGFIX] Fix Items getting desynced on conveyors

bcahue:
This was a 2 prong problem:

1.) Odamex didn't have Boom behavior when it came to items on conveyors. When they touch a non-moving player, they're supposed to pass on thru the player, unless the player is on a conveyors as well, then they'll pick the item up because they're moving. In Odamex, items wouldn't pass thru players, whether they're on a conveyors or not.

In DSDA, GZDoom, and every other port I checked, the intended behavior was the same (items passing thru players).

Here's DSDA on 2 demo maps (latest complevel):

https://www.youtube.com/watch?v=w-LEpAW4FhQ

Here's Odamex Stable with the bad behavior:

https://www.youtube.com/watch?v=fdz0DbkViZE&feature=youtu.be

And here's this patch for Odamex:

https://www.youtube.com/watch?v=PR4jVcx3Wxk

Problem number 2: when connecting, due to a timing issue, there was a chance for momentum to be applied clientside and serverside in an incorrect order. This would cause items to be flown off conveyors or have incorrect positions on them. This corrects that.

Here's how it looks with high latency:

https://cdn.discordapp.com/attachments/250157428455243777/1117157557019090975/chop.mp4


common/p_map.cpp
```diff
@@ -697,12 +697,31 @@ static BOOL PIT_CheckThing (AActor *thing)
 		// [SL] Work-around the additional height added to players
 		// in P_CheckPosition. Don't let players grab items above
 		// their real height!
-		if (!P_AllowPassover() || !tmthing->player ||
-			thing->z < tmthing->z + tmthing->height - 24*FRACUNIT)
+		if (!P_AllowPassover() || thing->z < tmthing->z + tmthing->height - 24*FRACUNIT)
```
seanmleonard: Since we're removing the check for `tmthing->player` here, I think it would be good to add a `max_z` variable as part of this check.
```
fixed_t max_z = tmthing->z + tmthing->height;
if (tmthing->player)
    max_z -= 24 * FRACUNIT;
if (!P_AllowPassover() || thing->z < max_z)
    P_TouchSpecialThing(thing, tmthing);
```

common/p_map.cpp
```diff
@@ -697,12 +697,31 @@ static BOOL PIT_CheckThing (AActor *thing)
...
+	// despite another solid thing being in the way.
+	// killough 4/11/98: Treat no-clipping things as not blocking
+	// ...but not in demo_compatibility mode
+
+	// e6y
+	// Correction of wrong return value with demo_compatibility.
+	// There is no more synch on http://www.doomworld.com/sda/dwdemo/w303-115.zip
+	// (with correction in setMobjInfoValue)
+	if (demoplayback || !co_boomphys
+	    //&& !prboom_comp[PC_TREAT_NO_CLIPPING_THINGS_AS_NOT_BLOCKING].state
```
seanmleonard, line 731: minor: it's preferable to not add commented out code. If you think it'd be helpful to add some breadcrumbs for finding this original Boom code in case we add Boom compat levels in the future, that'd would be helpful

common/p_map.cpp
```diff
@@ -697,12 +697,31 @@ static BOOL PIT_CheckThing (AActor *thing)
...
+			)
+		return !(thing->flags & MF_SOLID);
+	else
+		return !((thing->flags & MF_SOLID && !(thing->flags & MF_NOCLIP)) &&
+		         (tmthing->flags & MF_SOLID || (demoplayback || !co_boomphys)));
+
+	// return !(thing->flags & MF_SOLID);   // old code -- killough
```
seanmleonard: minor: no need to keep this around

server/src/sv_main.cpp
```diff
@@ -1165,6 +1165,15 @@ bool SV_AwarenessUpdate(player_t &player, AActor *mo)
 
 		return true;
 	}
+	else if (previously_ok && ok 
+			&& player.mo && !mo->player && mo->type != MT_AVATAR && mo->type != MT_PLAYER 
+			&& mo->on_conveyor
+			&& (gametic + mo->netid) % 120) // Update items on conveyers every 120 tics
```
seanmleonard, line 1171: Is the client predicting the position of items on conveyors and this is just to ensure sync? ~3.5 seconds is a long time for an item to be on a conveyor.

rakohus:
Tested online and local with `co_boomphys 1` and it seems to work fine. Tested against two separate wads

common/p_map.cpp
bcahue: That was the intent, eventually I am going to go back and re-engineer to allow boom comp levels one day.

server/src/sv_main.cpp
bcahue: Seems to be a timing issue that affects more than just this (the item on the client is almost always behind, and it depends on when you connect/when the item respawns for how behind it is), this is just a rubber band to get it to work at QuakeCon. 😅

common/p_map.cpp
bcahue: Removed.

common/p_map.cpp
bcahue: Fixed.

seanmleonard (approved):
As bcahue mentioned in the review, it's a band-aid fix. It needs more work (bring in clientside prediction for conveyor items or something similar) but if it works well enough on low and medium ping in testing, let's ship it. Some occasional desyncs are better than the mess it was previously.

PR #870, opened 2023-06-18
[BUGFIX] Fix +RIP projectiles

### LINKED ISSUE #850: [BUG] `+RIP` flag doesn't work as intended
**Describe the bug**
`+RIP` mbf21 flag doesn't work - projectiles blow up as normal on contact with actor.

10.3

**To Reproduce**
Get hit with a +rip projectile.

**Expected behavior**
+RIP continues through the actors until it hits wall or an actor that is immune.

bcahue:
This was caused by the MBF21 dehacked function checking for incorrect specified bits when it came to RIP. This patch fixes this. Should fix #850

### DIFF
--- a/common/d_dehacked.cpp
+++ b/common/d_dehacked.cpp
@@ -1244,19 +1244,14 @@ static int PatchThing(int thingy)
 						info->flags3 |= MF3_E4M8BOSS;
 					}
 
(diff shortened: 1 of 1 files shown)


rakohus:
Tested and confirmed against test wad in #850

rakohus (approved):

PR #871, opened 2023-06-18
[BUGFIX] Revert ticcmd changes

bcahue:
The ticcmd changes need to be reverted as they cause lag after a new game. We need to look at this issue closer at a later time.


rakohus (approved):
Tagging @seanmleonard so he's aware of the rollback. The reason for the rollback is due to the ticcmd changes blocking other testing. We can revisit shortly.

PR #872, opened 2023-06-18
[FEATURE] Add options tab to boot GUI

electricbrass:
I often find myself wishing I had the no monsters option when loading a wad from the GUI. This adds a new tab with checkboxes for options such as that. This also adds the ability to double click an IWAD in the Game Select list to start the game, same as clicking the Play button.

image

### DIFF
--- a/client/gui/gui_boot.cpp
+++ b/client/gui/gui_boot.cpp
@@ -112,6 +112,9 @@ class BootWindow : public Fl_Window
 	Fl_Hold_Browser* m_IWADBrowser;
 	Fl_Check_Browser* m_PWADSelectBrowser;
 	Fl_Hold_Browser* m_PWADOrderBrowser;
+	Fl_Check_Button* m_nomonstersCheckButton;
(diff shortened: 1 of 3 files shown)


rakohus:
Tested all options and seems good! Double click is a nice addition.

client/gui/gui_boot.cpp
```diff
@@ -527,7 +565,7 @@ class BootWindow : public Fl_Window
  */
 static BootWindow* MakeBootWindow()
 {
-	return new BootWindow(0, 0, 425, 240, "Odamex " DOTVERSIONSTR);
+	return new BootWindow(0, 0, WINDOW_WIDTH, WINDOW_HEIGHT, "Odamex " DOTVERSIONSTR);
```
seanmleonard, line 568: I would like to see the window dimensions eventually be derived dynamically but I'm not sure of the logic to control that. This is at least a step in the right direction.

seanmleonard (approved):

PR #874, opened 2023-07-01
[BUGFIX] Fix obits for MBF21 Projectiles

bcahue:
Previously, I fixed obits for MBF21 monsters and most instances of player weapons. However, I did not fix player projectiles. This fixes that.


rakohus (approved):
LGTM

PR #882, opened 2023-07-21
Fix MAPINFO token compat_invisibility from causing an error

DeathEgg:
As said in the title. Does not implement said token.


bcahue (approved):
Looks good to me.

PR #886, opened 2023-08-10
Update M_PlayerNameChanged to keep track of PSetupDepth and not immediately exit the menu.

Acts19quiz:
The current fix (https://github.com/odamex/odamex/pull/816) for keeping track of the menu stack depth in Player Setup has an oversight. Any submitted changes to player names immediately exits the menu. I can't find a probable cause for this.

This hot fix adds another integer to keep track of what the current menu stack depth is, then immediately issues a "menu_player" command after a successful name change. One side effect is that the fire background behind the player sprite refreshes. However, it does do the job of staying in the Player Setup menu after name changes while keeping track of what the menu stack depth should have been.

Another minor change is with regards to the policy on sprintf. I accordingly changed the name change command over to std::string.

### DIFF
--- a/client/src/m_menu.cpp
+++ b/client/src/m_menu.cpp
@@ -114,6 +114,7 @@ bool				drawSkull;			// [RH] don't always draw skull
 
 // hack for PlayerSetup
 int					PSetupDepth;
+int					PSetupArchive;
 
 // graphic name of skulls
 char				skullName[2][9] = {"M_SKULL1", "M_SKULL2"};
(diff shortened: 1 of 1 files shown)


bcahue (changes requested):
We'll need to debug this to figure out exactly why changes in the player menu back out completely rather than simply start the player menu every change. It's a hack I can't quite stomach.

Acts19quiz:
10-4. While a simple fix, it also doesn't really answer the question of why this happens in the first place (i.e., curing the symptom vs. the disease).

For reference, it isn't _all_ changes in the Player Menu that cause the back out, but specifically all changes to the player name. As soon as you hit return, BAM, back to gameplay.

electricbrass:
This is due to the name field reusing the code for handling the savegame slots. When the name field is selected with enter, `genStringEnter` is set to 1, and now in `M_Responder`, when the enter key is pressed, this case will be reached:
image
Depending on whether `genStringEnd` was set to `M_PlayerNameChanged` or `M_DoSave`, either the game will be saved or the player's name will be updated. Either way, `M_ClearMenus()` is called, and the menu is exited.

bcahue:
Closing this as @electricbrass 's changes provides the result intended for this PR.

Issue #881, opened 2023-07-21
[CHANGE] "Player became a spectator" message


rodylg:
**What is the existing behavior? Please describe.**
The message on the hud says "Player became a spectator."

**What is the desired behavior? Why are you requesting the change?**
The message should read "Player became a spectator. (n frags, m deaths)". That will make advanced scorekeeping using demo replays easier.

**If applicable, describe alternatives you've considered**
N/A

**Additional context**
It should look similar to the topmost message
![Screenshot](https://github.com/odamex/odamex/assets/49259699/afda70b8-eb64-4388-8aec-59c0deeb3a4d)

bcahue:
Thanks to @seanmleonard for the fix, this is now implemented in `stable`.

PR #887, opened 2023-08-12
Display the number of kills & deaths when an active player becomes a spectator

seanmleonard:
https://github.com/odamex/odamex/issues/881

Resolves the feature request to have a kill/death count when a player becomes a spectator, similar to what's broadcast when a player disconnects from the server.


rakohus:
Tested, works well

bcahue (approved):
Looks good.

PR #888, opened 2023-08-15
[FEATURE] Allow player to 🦎

devnoot:
Ever wish you could play as your all-time favorite species? 

**Well now you can!** 

Join the ranks of the world's elite like while you're shredding demons to bits.

This PR adds a feature allowing you to choose *LIZARD* in the player config.

image


rakohus:
As a person with reptilian tendencies, this PR has my interest. However, `lizard` is certainly better classified as an `order` rather than a `gender`. Indeed, to name a `lizard` as a species would be to err. Such a change should compel discussion regarding our player classifications in general.

electricbrass:
This actually crashes the client when a message using a gendered pronoun would be shown, if a player is a lizard. This appears to be because it is accessing an out of bounds index of `genderstuff` in `p_interaction.cpp`. I do think a 4th option that uses they/them/their in the messages would be a good addition.

devnoot:
> 

Dang! I thought I got all the instances of gender. I'll issue another PR with the changes to `p_interaction.cpp` shortly-ish

bcahue:
GitHub's terms of service strictly forbid any non-human animal on the platform.

PR #894, opened 2023-09-04
[BUGFIX] Fix monsters not activating online after a certain amount of players

bcahue:
Found a vanilla check running online preventing players from being "seen" by monsters unless they shoot.


bcahue:
Closing as this was not the issue

Issue #893, opened 2023-09-02
[CHANGE] Do not null teleporter destinations where `sv_skill 0`


rakohus:
**What is the existing behavior? Please describe.**
`sv_skill 0` is no things mode. Per loopfz in Discord, it precludes teleporter destinations.

**What is the desired behavior? Why are you requesting the change?**
Leave teleporter destinations in addition to player starts.

rakohus:
Also add bridge things, water thing, pushers/pullers to whitelist

rakohus:
Addressed by @loopfz's PR in #896. Thanks!

PR #896, opened 2023-09-07
Change g_coopthingfilter to g_thingfilter, and allow the value "3"

loopfz:
This is my proposition to address #893 

I generalized `g_coopthingfilter` and changed it to `g_thingfilter`.
Value "3" removes all pickupable things from the map, leaving teleporters, etc. intact

Context: I'm trying to set up a TLMS server with loadout (using `g_spawninv`) and no items on the floor

### DIFF
--- a/common/c_cvarlist.cpp
+++ b/common/c_cvarlist.cpp
@@ -245,12 +245,13 @@ CVAR(g_preroundreset, "0", "After preround is over, reset the map one last time.
 CVAR(g_postroundtime, "3", "Amount of time after a round before the next round/endgame",
      CVARTYPE_INT, CVAR_SERVERARCHIVE | CVAR_NOENABLEDISABLE)
 
-CVAR_RANGE(g_coopthingfilter, "0", "Removes cooperative things of the map. Values are:\n" \
-	"// 0 - All Coop things are retained (default).\n" \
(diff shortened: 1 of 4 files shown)


rakohus:
I've tested this and it works, but I think we could possibly go even further. This is a nice paradigm shift from the `sv_skill` approach. Maybe filter options to ADD the multiplayer items to single player as well, in the event that an offline player wants them.

loopfz:
Interesting idea!
There are indeed many filter possibilities that could make sense, it's just a matter of deciding how much clutter you're willing to tolerate :)

An example I considered was offsetting the initial spawn of pickupable objects, because I found attack & defense CTF rounds becoming repetitive when there were powerful items in mid e.g. blue armor in Red Shift, since they respawn at each round start it becomes a race to pick it up. I didn't consider it too seriously though so it may not make sense; and I didnt want to clutter this PR too much, just mentioning it as an example.

Btw, I left keys out of this "3" filter, thinking that fringe uses such as g_spawninv berserk + g_thingfilter 3 could make for some weird DIY tyson challenge or something, and I didn't want to make progression impossible.

rakohus:
I agree with leaving keys along with teleporter destinations. Good change. Pending code review, it will be merged

bcahue (approved):
Looks good here

Issue #898, opened 2023-09-11
[FEATURE] WAD list for coop servers


loopfz:
**Is your feature request related to a problem? Please describe.**
I find the maplist feature perfect for DM/CTF gametypes, to easily cherry-pick a few maps from different wads, but I don't think it's very convenient for coop servers:

- if you want to feature several wads within the same coop server, the maplist ends up being very long quickly
- it forces a pistol start on every map. I don't think it's such a big deal because it's the preference of many players, just mentioning it for completeness sake
- you can't take secret exits to go to map31, since they play in order no matter what, you'll end up on map16

**Describe the solution you'd like to see**
My ideal solution would allow players to easily browse & callvote through a long list of wads using a command such as `wadlist`:

- `1. doom2 a_aliens_1_2.wad`
- `2. doom2 valiant.wad`
- `3. doom2 overboard.wad`
- `4. doom2 eviternity.wad`

Then call a vote: `callvote wad 2` to start playing Valiant.
Once passed, the maplist stays empty so natural progression within the wad is used.

**Describe alternatives you've considered**
Alternatively, I think a new cvar could be introduced to inhibit the maplist behavior on mapchange but still use it for votes.
Then I could populate my coop server maplist with only MAP01's.

`rcon sv_maplistIgnoreNextMap`:
`"sv_maplistIgnoreNextMap" is enabled`

`maplist`:

- `1. doom2 a_aliens_1_2.wad MAP01`
- `2. doom2 valiant.wad MAP01`
- `3. doom2 overboard.wad MAP01`
- `4. doom2 eviternity.wad MAP01`

Call a vote: `callvote map 2` to start playing Valiant.
Once passed, even though the maplist isn't empty, the cvar inhibits its use when selecting the next map by forcing this code path: https://github.com/odamex/odamex/blob/stable/server/src/sv_level.cpp#L257

This is functionally equivalent to the solution described earlier and seems a lot less intrusive, so it seems like a more realistic approach. Less maintenance if/when you want to evolve the maplist feature as well.

**Additional context**
Both solutions require a way to warp within the wad once the vote passes. Is there already a solution that I missed?
Otherwise, something like `callvote idclev map07` could work.

What do you think?
I'd be happy to contribute the change if you like the idea.

Issue #695, opened 2022-07-03
[FEATURE] `addwad` command


rakohus:
**Is your feature request related to a problem? Please describe.**
We currently have `addmap` which works great for non-coop modes. However, there is currently no simple way to load a server with multiple wads for coop player. You can add via `addmap`, but then secret exits aren't considered.

**Describe the solution you'd like to see**
Create the `addwad` command. You add a wad and Odamex/srv will automatically look at the maps in the wad and make an internal list. The list would observe the proper next map and, once reaching the final map in the current wad, move onto the next pwad. It also could populate the `maplist` list so that players could vote and jump between coop wads easier.

**Additional context**
This feels like a significant feature so input is appreciated.

PR #899, opened 2023-09-11
Implement sv_mapliststayonwad

loopfz:
Implements #898 

Running on a test server, my maplist config file now looks like:
```
addmap map01 doom2 aaliens_v1_2.wad
addmap map01 doom2 eviternity.wad
addmap map01 doom2 valiant.wad
addmap map01 doom2 vanguard.wad
addmap map01 doom2 overboard_fix.wad

gotomap 1
```

`gotomap 1` is a small quirk introduced by this PR: if not included while `sv_maplistignoremapchange` is enabled, the server will use doom2 MAP01 when booting instead of the first map from the maplist.

I ran a few maps on my test server and jumped between wads, and it felt *really* nice to use.
Also worth noting: `callvote nextmap` will jump to the next map within the wad, making it possible if slightly cumbersome to warp to a given map without another vote command, for now.


loopfz:
This implements #695 as well

loopfz:
I reworked this, I was unsatisfied after discussing its limitations with @rakohus 

I think the way it works now is pretty nice:
- renamed cvar to `sv_mapliststayonwad`, which seems easier to understand
- players progress naturally through the maps of the wad, then when the end is reached, the next maplist entry is loaded
- added an optional `lastmap=xxx` parameter to the `wad`, `addmap`, and `insertmap` commands, to artificially limit them for wads that are less than a full megawad and don't have mapinfo

It's used like this:
```
addmap map01 doom2 eviternity.wad
addmap map01 doom2 lunatic.wad lastmap=map05
addmap map01 doom2 vanguard.wad lastmap=map13
addmap map01 doom2 valiant.wad
```

It only works in last position of the command args.

fuzztooth:
Bumping - if you're able to update the conflict we will include this in an upcoming release.

cuantar:
This is just over a year old; I've been using it on my server for almost that long.  It needed a change to apply against `odamex-10.6.0` that I've attached here, in case someone finds it useful.
[odamex-PR-899-maplistignoremapchange.patch.txt](https://github.com/user-attachments/files/16989205/odamex-PR-899-maplistignoremapchange.patch.txt)

electricbrass:
I think the `maplist` command should also display the lastmap.

Something like this:
image
9011940444f17dca39020955ed6d4f81f2f1ad10

PR #900, opened 2023-09-12
Bugfix: Items picked up by a player arent added to his local item count

loopfz:
Reported by `Zigmun` on discord:

> On coop servers, item count doesn't work. Lan and online, it stays at 0.

Actually it is a little more complicated; items picked up by *other players* are correctly added to the item count, but not items picked up by the player itself.

Items picked up by other players are correctly processed here: https://github.com/odamex/odamex/blob/stable/client/src/cl_parse.cpp#L926

Changing `P_GiveSpecial` (https://github.com/odamex/odamex/blob/stable/common/p_interaction.cpp#L1231) to also increment the `found_items`  count on the client side seems to fix the issue; items picked up by the player are added to the count, and the count stays sync'ed between players.


loopfz:
I forgot to mention; the server side item count is correct, which means that a player joining mid game gets the correct count via level state

bcahue (approved):
Looks good to me.

Issue #912, opened 2023-09-29
[FEATURE] Let players keep their inventory after dying in cooperative modes


agiskell:
**Is your feature request related to a problem? Please describe.**
Odamex already allows players to retain keys(share them too!) on death, but not inventory, which is pretty surprising considering the other two MP-focused ports, ZDaemon and Zandronum, let the player keep their inventory after dying.

**Describe the solution you'd like to see**
Implement a cvar named something like sv_keepinv where 1 - player keeps their inventory on respawn and 0 - current behavior, where player loses their inventory.

rakohus:
Maybe a better approach here would be `sv_keepweapons`, to avoid overlap with `sv_keepkeys`

PR #913, opened 2023-10-01
Implement sv_keepweapons: keep weapons and ammo on death

loopfz:
Implements #912


rakohus:
Tested and it works great. Just needs someone to peek at the code.

cuantar:
I don't like that this patch overrides pistol start at the beginning of a level, as implemented.  Ideally it should differentiate between a dead player and one starting a a level.  It works as expected, though!

cmbernard333:
This will be awesome for multiplayer for really intense WADS.

rakohus:
> I don't like that this patch overrides pistol start at the beginning of a level, as implemented. Ideally it should differentiate between a dead player and one starting a a level. It works as expected, though!

Agreed. @loopfz this is a good patch, but I recommend that you make it so that if `g_resetinvonexit = 1`, then `sv_keepweapons` doesn't carry the inventory into the next map. Other than that, this seems good to be merged in.

fuzztooth:
Bumping - with the small update requested by Ralphis and cuantar we can merge this in.

electricbrass:
Another issue is if a player becomes a spectator and rejoins, they will appear to keep their weapons when `sv_keepweapons` is enabled. This only happens on the client however, and as soon as they fire a weapon, it will be corrected.

cuantar:
This patch doesn't apply against 11.0.0.

PR #911, opened 2023-09-21
Implement sv_curpwad to allow scripting based on wad names

loopfz:
This is meant to simplify coop server administration to allow customizing compatibility cvars, g_thingfilter (to remove too generous coop things in some wads), or sv_doubleammo (e.g. to compensate for a wad not designed for coop play) differently for each wad featured on the server.

Example use in a `sv_startmapscript` file :
```
alias "boomphys" "set co_allowdropoff 1; set co_boomphys 1; set co_novileghosts 1; set co_removesoullimit 1"
if sv_curpwad eq EVITERNITY.WAD boomphys
if sv_curpwad eq OVERBOARD.WAD boomphys
// etc.
```

PR #914, opened 2023-10-01
co_deadplayertarget: enemies keep chasing a dead player to avoid rare map bugs

loopfz:
I discovered a weird/broken behavior on Lunatic map05.

The map functions this way:

- you start in a secluded area, which is in the same sound propagation zone as *all* the monster closets of the map
- you have to shoot a switch to reach the next area, thus waking up *all* enemies
- you then teleport to the main arena. This sets free a voodoo doll on a conveyor belt which progressively releases monster closets to spawn the enemy waves in the main arena

The weird behavior was: enemy waves would randomly stop in the middle of the map.

I suspected something might be wrong with the voodoo dolls, but after checking the doll progression and all the closet states, it was fine.

Then I realized that the remaining monsters were actually idle in their closet, not awake. And since there are no conveyor belts in the closets, they just sit there. See the demo: http://doom.koholint.io/odamex-config/misc/odamex_lunatic_map05_bug.odd

 I wondered if there might be a sound propagation issue, so I experimented with `P_RecursiveSound` a bit, backporting boom code for testing purposes. This was a dead end, everything was fine.

It turns out the root cause is simpler: it was tied with player deaths.
When the player who fired at the switch initially, died, enemies dropped their target, which means those that didnt teleport in yet were stuck forever unless a respawning player happened to fire in the starting area (because the main arena is in a different sound propagation zone).

Different map design leading to different sound propagation zones, or a linedef to close the door in the initial area after teleporting would fix this specific map, but this seems like a general enough (if slightly odd) issue to fix it in the server itself instead of the map.

To fix it, I pick a new target among the players for awaken enemies when a player being targeted is either dead or spectates. This is a bit arbitrary, e.g. it puts more emphasis on the players at the start of the playerlist since it uses the first eligible one, and it also doesnt account for position at all which may lead to slightly odd behavior if the new target happens to be *far away* from the enemy whereas other players were right there.

I'm interested in hearing your thoughts about this one, it took a while to track down for me and the fix is a bit experimental.

### DIFF
--- a/common/c_cvarlist.cpp
+++ b/common/c_cvarlist.cpp
@@ -323,6 +323,9 @@ CVAR_RANGE(sv_countdown, "5",
 	CVAR(			co_blockmapfix, "0", "Fix the blockmap collision bug",
 					CVARTYPE_BOOL, CVAR_ARCHIVE | CVAR_SERVERINFO | CVAR_LATCH)
 
+	CVAR(                   co_deadplayertarget, "0", "Enemies keep chasing a dead player",
+					CVARTYPE_BOOL, CVAR_ARCHIVE | CVAR_SERVERINFO)
+
 
 	// Boom-compatibility changes
 	//------------------------------
--- a/common/c_cvarlist.cpp
+++ b/common/c_cvarlist.cpp
@@ -120,6 +120,10 @@ CVAR_RANGE(			sv_monstershealth, "1.0", "Amount to multiply monster health by",
 					CVARTYPE_FLOAT, CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_LATCH | CVAR_NOENABLEDISABLE,
 					0.0f, 100.0f)
 
+CVAR(                           sv_deadplayertarget, "0", "Enemies keep chasing a dead player",
+					CVARTYPE_BOOL, CVAR_SERVERARCHIVE | CVAR_SERVERINFO)
+
+
 CVAR_RANGE(			sv_skill,"3", "Sets the skill level, values are:\n" \
 					"// 0 - No things mode\n" \
 					"// 1 - I'm Too Young To Die\n" \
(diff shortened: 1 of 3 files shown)


loopfz:
Also, I realize this could (should?) probably be governed by a `co_` cvar, but I wanted to get the discussion going about the merits of changing this behavior or not and discuss the logic itself before going any deeper.

loopfz:
I tested in Woof coop with `-complevel 9` and, as expected, reproduced the issue, so I hid this under a new compatibility cvar in the `Fixes to Vanilla` category.

loopfz:
After sleeping on it I tried to make the logic less weird:

- if co_deadplayertarget is enabled, enemies will keep the player corpse as their target
- corpse garbage collection is prevented by forcing deadtic to 0 while it is still referenced as a target
- enemies targeting a corpse look for new targets via sound+sight, while still moving towards the corpse until then
- they do not shoot the corpse

As soon as a corpse isnt targeted anymore, it is released after deadtic reaches REMOVECORPSESTIC.

I'm still debating how worthwhile this fix is. It's definitely niche, but I guess it could help wads that were not designed for coop play smoother in coop (like Lunatic, which was only tested for single player).

fuzztooth:
Thanks.  We will check this out and respond with feedback.

bcahue (changes requested):
I'm not a fan of the monsters being active while the server is empty if everyone leaves (and if sv_freezetime isn't enabled), is there any chance this could be refactored into activating if a player has spawned in, and online only? (as in, making it an sv_ flag)

loopfz:
@bcahue what do you think now?

I've updated the koholint servers to run this latest version of this PR, along with #899 and #911, rebased to stable tip for further testing.

Btw I've had time to reflect on this change since it has been running on my coop servers for the past year (and used frequently by my playgroup in coop). We've never encountered a weird side effect, but it has given me peace of mind when playing maps that clearly weren't designed for multiplayer, so I think this is pretty useful overall.

PR #915, opened 2023-10-03
Fix wad downloads when filename uses mixed case

### LINKED ISSUE #843: [CHANGE] Add specific case sensitive searching to wad downloading
Currently, when comparing client wads to server wads, the downloading behavior searches for all lower case or all upper case. 

If a server is running a wad with a specific case, search for that specifically. e.g. `ExampleWad.wad`, then the client should also do the curl search for `ExampleWad.wad`.

Right now, I have taken to making all wads I upload lowercase, even if the original wad as released did not. This is to prevent issues with the clients finding the wad. 

This is, in my opinion, a critical and required update to our current behavior.

loopfz:
Fixes #843

This makes it so filenames advertised to the client use exactly the same case as the files present on the server.

I've tested a linux server with linux and windows clients, and it seems to work fine.
Tested via maplist progression, `callvote map`, `rcon wad`, with upper, lower, and mixed case filenames.
Everything downloaded correctly through the server advertised in `sv_downloadsites`.
Also tested locally in the client via the `wad` command.

nginx access log serving `Lunatic.wad` for reference:
`[03/Oct/2023:16:32:53 +0000] "GET /wads/Lunatic.wad HTTP/1.1" 200 18160632 "-" "Odamex/10.4.0"`

I think this works fine, but I'd appreciate someone more experimented with the codebase making sure that the changes in common/m_resfile.cpp don't have weird side effects.

### DIFF
--- a/common/m_resfile.cpp
+++ b/common/m_resfile.cpp
@@ -71,7 +71,7 @@ bool OResFile::make(OResFile& out, const std::string& file)
 
 	out.m_fullpath = fullpath;
 	out.m_md5 = hash;
-	out.m_basename = StdStringToUpper(basename);
+	out.m_basename = basename;
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


rakohus:
Tested this for function and it seems to work well. Pending code review.

seanmleonard (approved):
Looks good! This behavior will need to change in the `resource_files` branch but I'll take ownership of that.

PR #928, opened 2023-12-06
Fix auto-SIMD for i386+sse2, ppc+altivec

matoro:
Only Darwin ever used `-faltivec`, gcc uses `-maltivec` since at least 2007: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=30254#c1

Similarly, recent versions of gcc define `__APPLE_ALTIVEC__` for compatibility, so just unconditionally include `altivec.h`.

Canonical way to enable specific extensions is e.g. `-msse2`, use that instead of optimizing for a specific CPU.

Finally, cmake uses `target_compile_options` to add arbitrary flags. `target_compile_definitions` assumes you want a preprocessor definition, which was adding `-D-faltivec` to the command line which gcc doesn't like; this is what originally prompted this change.


chewi:
I don't know what's happened here, but this fix hasn't been applied.

HyperEye (approved):

PR #929, opened 2023-12-19
Fix for demos that play only audio and no video 

enaiel:
While trying to create a video out of a 4 player co-op game, I found that 2 netdemos would only play audio and no video.
Debugging the issue, I discovered the reason was that these demos were getting stuck in GS_CONNECTED state and never received the message to go to GS_LEVEL state. 
This PR will skip the GS_CONNECTED state for netdemos and go straight to GS_LEVEL state, allowing the demos to play video.
This change might have unintended consequences.

Attached a sample demo file exhibiting this issue.

[Odamex_COOP_20231210_220540_doomboy_NERVE.WAD_MAP08.odd.zip](https://github.com/odamex/odamex/files/13717075/Odamex_COOP_20231210_220540_doomboy_NERVE.WAD_MAP08.odd.zip)


bcahue (changes requested):
How about when the demo begins in a state like GS_INTERMISSION?

electricbrass:
The underlying reason for this issue is netdemos with broken headers. At the end of recording a demo, there is a cleanup step that rewrites the header with the proper values. If this is skipped, such as by a crash, the values in the header will be wrong. The `map_index_size` will be zero (indicating that zero maps were visited in the demo), which results in an early return from `NetDemo::PrevMap` when loading the first map in the demo. It is later in this function that the map is fully loaded and the game state is changed.

electricbrass:
Superceded by #1692

PR #930, opened 2023-12-19
A couple of improvements for the "Spy" command

enaiel:
The undocumented "Spy" command allows the user to spy on a named player. The command sets the "spyplayername" but only spies on the named player if they are already in the game. This PR changes this behavior to start spying on the named player as soon as they spawn.

This PR also introduces a client variable "hud_hidespyname" to enable hiding of the name of player being spied on.

Both these features are useful when creating a video from a netdemo using another player's POV, hiding their name, and switching perspective whenever they join and re-join the game.


rakohus:
I have tested this patch and it works very well - except for one edge case. If you are spying a player in coop, and then you join, it will stay spying on that player. It should probably back out in this case. Otherwise, it's good to go.

fuzztooth:
Bumping - if you can update this per Ralphis's suggestion, we'll include it.

PR #931, opened 2023-12-19
A couple of fixes for vanilla "oldintermission" screen

enaiel:
This PR introduces a couple of fixes for vanilla `oldintermission` screen (shown if variable `wi_oldintermission` is set):

* Draw `displayplayer` face instead of `consoleplayer`. Needed when spying on another player.
* Fix incorrect player background color to correctly reflect the player's color.

This PR does not fix the broken stats in the vanilla `oldintermission` screen.


client/src/wi_stuff.cpp
```diff
@@ -1135,10 +1135,11 @@ void WI_drawNetgameStats()
 
 		unsigned int x = NG_STATSX;
 		// [RH] Only use one graphic for the face backgrounds
-		if (demoplayback)
+		//enaiel: Fix incorrect player background when showing old intermission
```
seanmleonard, line 1138: If your change removes old code, please just delete the old code instead of commenting it out. The old code is preserved in git history so if it's needed later, it can be found.

rakohus:
Tested and it seems to do as said. Waiting on changes re: Sean's advice on PR.

client/src/wi_stuff.cpp
bcahue: I removed this in preparation to merge

bcahue (approved):
Looks good now

PR #932, opened 2023-12-28
Use R_BuildClassicPlayerTranslation from r_draw.cpp to use Vanilla palette indexing for select colour presets instead of RGB values.

Acts19quiz:
This is another much more extensive commit to address the desire for the colour presets to match the palette indexing from Vanilla DOOM as opposed to the RGBed versions currently provided by Odamex. https://github.com/odamex/odamex/issues/322

Here are some areas of note.

* OVERVIEW: Currently in 10.4.0, colorpreset isn't broadcast to the server and it doesn't keep track of it. Only the client needs to know its own player's colorpreset since it just sets the RGB values that get sent anyway. This is not the case anymore. In order for Vanilla presets to use palette indexing, the server must know if a player is using a Vanilla preset in order for it to know not to use the RGB value. Thus, server.proto and other various net files have been modified to keep track of it.
* `void R_BuildPlayerTranslation` in r_draw.cpp now accepts three arguments instead of two. If it detects that a player is using a colorpreset that has a Vanilla definition and isn't being encumbered by a circumstance that overrides a player-chosen color (a team gametype, r-forceenemycolor, or r-forceteamcolor), it instead hands off to `void R_BuildClassicPlayerTranslation`.
  * This required changing all calls to `R_BuildPlayerTranslation` in the code to accept three arguments, meaning in many instances it had to be modified to be aware of and query the relevant player's userinfo.
  * The reason this was done as opposed to leaving `R_BuildPlayerTranslation` alone and just having all calls to `R_BuildPlayerTranslation` alternate between it and `R_BuildClassicPlayerTranslation`: Doing the latter would require several more if/else statements scattered throughout the code, as well as having the `(colorpreset > -1 && colorpreset < 6)` line duplicated several times throughout. The former approach is (in my opinion) the cleaner and more modular approach in case the Vanilla presets are modified or expanded in the future.
* g_gametype.cpp now has a new gametype check, `int G_IsTeamColor`. This determines if the gametype and user CVARs need to override a player's colorpreset, and if so, return an out-of-range number. Otherwise, it returns the colorpreset given to it.
* Related to the above, some .cpp files now have `#include "g_gametype.h"` so that they can call that function.
* The player colour RGB values are still changed even underneath the Vanilla preset being used. There are some reasons for this. The biggest two are that team and enemy colours set by gametype and r- CVARs need an RGB value to shade. Also, the scoreboard uses colour squares next to player names. Instead of using a player sprite translation, it feeds a RGB value directly into `argb_t CL_GetPlayerColor(player_t *player)`. I left well enough alone there.
* I made an educated guess on the intended workings of `R_BuildClassicPlayerTranslation` in r_draw.cpp. It looks like its only intended use was for Vanilla demo playback in cl_game.cpp, and that the first player (index 0) passed to it is given no translation (i.e., green). Assuming that the unspecified `(color == 0)` is implied green, I made it explicit.
* `static void CL_RebuildAllPlayerTranslations()` in cl_main.cpp had static removed because it needs to be referenced in cl_parse.cpp in order to keep everything straight.

To test this out, I set up a local server with `sv_usemaster 0` and a few running instances connected to it. I tested several combinations of gametype (DM, COOP, TEAM DM, CTF) and combinations of different clients alternating r-forceenemycolor and r-forceteamcolor on and off. I did this until I worked out all of the bugs and inconsistencies. After doing that, I published this first draft. It could still be subjected to more testing, of course, and there may be revisions to it desired by those here.

It's certainly possible that I've left out some references, saves, or loads of colorpreset that need to be added. However, from what I've been able to test so far, it seems to work.
Odamex_COOP_20231228_165504


Acts19quiz:
I just noticed that d_main.cpp has an extraneous comment from a series of them I used to keep track of the changes I made while working on this. That will be removed upon either approval or alteration suggestions.

common/g_gametype.cpp
```diff
@@ -338,6 +338,17 @@ bool G_IsTeamGame()
 	return sv_gametype == GM_TEAMDM || sv_gametype == GM_CTF;
 }
 
+/**
+ * @brief Check if team colors are being enforced due to gametype or CVARs.
+ */
+int G_IsTeamColor(int colorpreset, bool forceteamcolor, bool forceenemycolor)
```
seanmleonard: The function name implies a boolean return type, rather than a color present value. Maybe this function should be changed to test whether team colors should override the user's color selection (`G_ShouldUseTeamColor`).

client/src/r_draw.cpp
```diff
@@ -454,44 +472,51 @@ void R_CopyTranslationRGB (int fromplayer, int toplayer)
 
 // [RH] Create a player's translation table based on
 //		a given mid-range color.
-void R_BuildPlayerTranslation(int player, argb_t dest_color)
+void R_BuildPlayerTranslation(int player, argb_t dest_color, int colorpreset)
```
seanmleonard, line 483: A default value for `colorpreset` would simplify this function's use.

client/src/r_draw.cpp
```diff
@@ -454,44 +472,51 @@ void R_CopyTranslationRGB (int fromplayer, int toplayer)
...
-		s = 0.0f;
-	float sdelta = 0.014375f;
-
-	v += 0.1f;
-	if (v > 1.0f)
-		v = 1.0f;
-	float vdelta = -0.05882f;
-
-	for (int i = 0x70; i < 0x80; i++)
+	if (colorpreset > -1 && colorpreset < 6)
```
seanmleonard: Can we use an enum value to specify that a preset is invalid (eg, the player is not using a preset)? Often times, the last element in the enum is used to represent the invalid option. If `colorpreset` is changed to be an unsigned int, you can simply have the check `if (colorpreset < NO_COLORPRESET)`. That would remove two magic numbers from this line of code.

client/src/r_draw.cpp
seanmleonard: The default value could be the enum `NO_COLORPRESET` (or similar phrasing) mentioned below

client/src/r_draw.cpp
seanmleonard: Additionally, the logic to check if this is a team game (which would override the color preset) could go into this check. That would simplify all of the calls to `R_BuildPlayerTranslation()` because they wouldn't have to use the `G_IsTeamColor()` function to provide the value for `colorpreset`.

Acts19quiz: Regarding the first post: A distinction here is that there are actually two types of colour presets here. Currently, there are just colour presets, but the idea here is to convert _some_ of them to use the Vanilla palette indexing rather than RGB values. https://github.com/Acts19quiz/odamex_Acts19quiz/blob/808bc8738de81ee247e6e5a8229e2f1e8e2f42a1/client/src/d_netinfo.cpp#L123-L148 COLOR_GREEN (0) through COLOR_ORANGE (5) are Vanilla presets, while everything else are non-Vanilla presets that, while being colour presets, we don't want to be using Vanilla indexing. COLOR_PURPLE, for example, is a preset we still want RGB translation on. As well as COLOR_CUSTOM, obviously.

https://github.com/Acts19quiz/odamex_Acts19quiz/blob/808bc8738de81ee247e6e5a8229e2f1e8e2f42a1/common/d_netinf.h#L39-L55 For reference, this is the enum list that I assume you're referring to. There are 12 preset values ranging from 0-11, with 1 additional total value at the end. These are all presets that players can choose from in PLAYER SETUP. We want presets 0-5 to have Vanilla translation, and 6-11 to use RGB translation. Notably, a preset is _always_ active, even if it's COLOR_CUSTOM.

Given what you're suggesting in the first post, would Odamex still be able to make this distinction (i.e., it's in the 0-5 range)?

Acts19quiz: Regarding the second post: I just moved the r-forceteamcolor and r-forceenemycolor checks to a single G_IsTeamColor call in R_BuildPlayerTranslation. This introduced a bug where activating r-forceteamcolor (in a coop game) removes the Vanilla indexing for the player using it as well as the other players. R_BuildPlayerTranslation says, 'Oh, Player A has r-forceteamcolor on! Better stop *all* Vanilla translations!' without discriminating what sprites to stop the Vanilla translation for (Player B and C, and only from A's POV). If Player A has r-forceteamcolor on in COOP but is using Vanilla "red", we want to turn off Vanilla translation for B and C from A's POV, but still have A use Vanilla translation. The following calls to R_BuildPlayerTranslation need to individually collect the relevant player's r- values:

https://github.com/Acts19quiz/odamex_Acts19quiz/blob/808bc8738de81ee247e6e5a8229e2f1e8e2f42a1/client/src/cl_main.cpp#L239
https://github.com/Acts19quiz/odamex_Acts19quiz/blob/808bc8738de81ee247e6e5a8229e2f1e8e2f42a1/client/src/cl_main.cpp#L1443

r_draw.cpp can't check for r- because it just can't know. What I *could* do, since only two calls to R_BuildPlayerTranslation need to check for r- values, is have those two lines make additional checks. Something like this (from cl_main.cpp#L1443):

```
	else
	{
		R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), G_IsTeamColor(player.userinfo.colorpreset, r_forceteamcolor, r_forceenemycolor));
	}
```

to

```
	else
	{
		if ((r_forceteamcolor && G_IsCoopGame()) || (r_forceenemycolor && G_IsFFAGame()))
			R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), NULLVALUE);
		else
			R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), player.userinfo.colorpreset);
	}
```

Then r_draw.cpp can check for CTF or TEAMDM since those are the two circumstances where we actually want to turn Vanilla translations off all the time. At the expense of two extra if/else checks, all R_BuildPlayerTranslation calls are simplified and need to only pass colorpreset, not three different values to G_IsTeamColor. This _should_ also make G_IsTeamColor obsolete and allow for its removal. Would this (Proposal A) be a satisfactory alteration?

Proposal B would be to keep G_IsTeamColor, but make it a true boolean (as suggested above) and revamp it so that it doesn't receive or return a colorpreset value, and is used in a fashion similar to the second example (along with the teamgame check being moved to r_draw.cpp).
```
	else
	{
		if (G_IsTeamColor(r_forceenemycolor, r_forceteamcolor))
			R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), NULLVALUE);
		else
			R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), player.userinfo.colorpreset);
	}
```

Proposal C would keep G_IsTeamColor as an int, rename it to an appropriate suggestion, and move its gametype check to r_draw.cpp so that only the two cl_main.cpp calls to R_BuildPlayerTranslation have the long-form format.

seanmleonard: > Given what you're suggesting in the first post, would Odamex still be able to make this distinction (i.e., it's in the 0-5 range)?

I think you can accomplish this by sub-dividing the enum a bit
enum colorpreset_t	// Acts 19 quiz the order must match m_menu.cpp.
```
{
	COLOR_GREEN,
	COLOR_INDIGO,
	COLOR_BROWN,
	COLOR_RED,
	COLOR_BLUE,
	COLOR_ORANGE,
	INDEXED_COLOR_PRESET_RANGE,
	
	COLOR_GOLD,
	COLOR_JUNGLEGREEN,
	COLOR_PURPLE,
	COLOR_WHITE,
	COLOR_BLACK,
	COLOR_CUSTOM,

	ARGB_COLOR_PRESET_RANGE
}
```

So you can use these conditionals now:
`if (colorpreset < INDEXED_COLOR_PRESET_RANGE) ...`
or
`else if (colorpreset < ARGB_COLOR_PRESET_RANGE)...`

seanmleonard: It looks like the parameters that are passed to `R_BuildPlayerTranslation()` can all be obtained from a `player_t` object. I propose changing it to have a single `player_t` parameter and extracting the player ID, color preset ID, and ARGB translation value inside `R_BuildPlayerTranslation()`. Then `R_BuildPlayerTranslation()` can own all of the logic for determining how to translate the colors for that player.

`R_BuildPlayerTranslation` should be able to look at the color preset, the ARGB translation, the game mode and the force team/enemy color CVARs and decide how to translate the color for the given player, right?

Acts19quiz: 1: I tired this change out, and it's... problematic. I inserted the sub-divider into d_netinf.h, but this causes a crash when messing around in PLAYER SETUP. As my note there observes, the list must match `static const char *colorpresets` in m_menu.cpp (and vice versa). So I have to put the sub-divider in that list of options, as well. And doing that makes it want to have it as a selectible option in PLAYER SETUP, which we obviously don't want (and if I don't do that, `static void M_ChangeColorPreset (int choice)` goes haywire--and crashes). Maybe I did it wrong, but whereas I can truncate the list of options to ignore the last one (NUMCOLOR), this is right in the middle of options.

With enum usage akin to what you're suggesting, I was able to eliminate the `colorpreset > -1` check, but with an end-of-enum-list definition being the total (12) when we want 6, the best I can think of is something like `#define MAXVANILLA 6` in r_draw.cpp. That's probably not quite what you want, but it's the closest thing I can come up with on the spot.

2: If I understand correctly, what you're suggesting is effectively "Proposal D", transforming R_BuildPlayerTranslation from receiving `int player[.id]` to `player_t player`.

Acts19quiz:
What I've done with version 2 of this is Proposal B. https://github.com/Acts19quiz/odamex_Acts19quiz/blob/808bc8738de81ee247e6e5a8229e2f1e8e2f42a1/client/src/cl_parse.cpp#L963-L966 This benefits from calling G_IsTeamColor, eliminating Proposal A (it's also nice to keep the logic checks contained in one function to help keep things tidy), but to do this, it can't handle colorpreset anymore, eliminating Proposal C. The two big changes here are eliminating as many 'magic numbers' from R_Build as possible, and greatly simplifying G_IsTeamColor to eliminate handling of colopreset and be a true bool.

Acts19quiz:
Just to note here, a re-do of everything on an Odamex 12.1.0 base (from 10.4.0) is pending. However, a complication has arisen where I can't get `sv_showplayerpowerups 1` to work. This is important since I need to properly test the Vanilla skin indexing against the new tinting to make sure both work together. A triage is underway in Discord and pending outcome.

Acts19quiz:
Closing with prejudice, refer to #1594

PR #944, opened 2024-03-24
Update network connection error graphic

rakohus:
Update network connection error graphic to include text so that it's more clear what's going on.


rakohus:
Gotta fix the graphic to be a palettized gif. No idea, working on it.

PR #957, opened 2024-05-05
Fix Mac Builds by turning off system LIBPNG and ZLIB for FLTK builds

rakohus:
Per @HyperEye's advice, turn off system LIBPNG and ZLIB use for FLTK to fix Mac builds.


bcahue (approved):
LGTM

PR #958, opened 2024-05-05
Add a space in spectating string between stats

rakohus:
Currently, when someone spectates, it will show a string as `Ralphis became a spectator.(1 FRAGS, 0 DEATHS)`. This commit adds a space after the sentence and the stats.


bcahue (approved):
LGTM

PR #960, opened 2024-05-12
Fix a minor typo in `g_horde.cpp`

electricbrass:
This is a very small change. It just changes "hode" to "horde" in a warning.


rakohus (approved):
Impeccably done! :) Thank you for the fix

PR #963, opened 2024-05-23
Don't use localview pitch when playing demos

### LINKED ISSUE #962: [BUG] In netdemos, moving the mouse tries to change the POV of the recording player's POV.
**Describe the bug**
When a netdemo has been recorded where freelook is enabled on the server as well as the client, trying to move the mouse up and down while the netdemo is playing will try to briefly move the recording player's POV as if it is affecting a real game. A split second later, the POV will snap back to the proper place.

10.3.0 gb28ee-8578 (and probably earlier)
...
10.5.0

**To Reproduce**
1. Go into (or start) an online game with freelook enabled both on server and client.
2. After recording the netdemo, play it back.
3. While it's playing, move the mouse up and down in the window.

**Expected behavior**
For mouse movement, while watching a netdemo, to not affect the POV of the player that recorded it.

[Odamex_HORDE_20240519_124404_HORDAMEX.WAD_MAP19.zip](https://github.com/odamex/odamex/files/15370270/Odamex_HORDE_20240519_124404_HORDAMEX.WAD_MAP19.zip)
This netdemo is recorded with Odamex 10.5.0, and requires DOOM2.WAD and HORDAMEX.WAD.

**Additional context**
If there are other players in the game, this wont' affect the players that didn't record the netdemo if you're spectating them instead.

ceski-1:
Fixes https://github.com/odamex/odamex/issues/962

### DIFF
--- a/client/src/r_interp.cpp
+++ b/client/src/r_interp.cpp
@@ -28,8 +28,6 @@
 #include "m_fixed.h"
 #include "r_state.h"
 #include "p_local.h"
-#include "cl_demo.h"
-
 
 typedef std::pair<fixed_t, unsigned int> fixed_uint_pair;
 
@@ -37,8 +35,6 @@ static std::vector<fixed_uint_pair> prev_ceilingheight;
 static std::vector<fixed_uint_pair> saved_ceilingheight;
 static std::vector<fixed_uint_pair> prev_floorheight;
 static std::vector<fixed_uint_pair> saved_floorheight;
-
-extern NetDemo netdemo;
 
 //
 // R_InterpolationTicker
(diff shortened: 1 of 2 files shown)


bcahue:
This seems like it is lifting and shifting where the check is done lower in the stack and properly applying it to both the interpolation system and the actual camera system. I will run this build for a little while and if it passes testing (I have concerns about spectating, spectating in netdemo, spectating in vanilla demo with multiple players, and general gameplay after spectating) I will accept this

ceski-1:
Made a small adjustment since I missed `localview.skipangle` vs `localview.skippitch`.

bcahue (approved):
Looks good to me

PR #965, opened 2024-06-04
Escape URIs so CURL can properly find WADs with spaces

bcahue:
[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) lists the need to percent-encode reserved characters in URIs. These encompass many different characters but the most common use case is spaces. CURL requires a percent-encoded URI to properly function, which makes WADs with spaces not able to be searched, or the wrong WAD downloaded (the first word of the wad filename with no file extension, if exists, will be downloaded).

This uses `curl_easy_escape` to percent-encode the filename so the behavior is correct. This has been tested on WADs with spaces and WADs without spaces, and WADs with underscores.

### DIFF
--- a/client/src/cl_download.cpp
+++ b/client/src/cl_download.cpp
@@ -277,11 +277,15 @@ static void TickCheck()
 			::dlstate.checkfilename = StdStringToLower(::dlstate.checkfilename);
 		}
 
-		// Now we have the full URL.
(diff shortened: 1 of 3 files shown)


leximayfield (approved):
LGTM.  Good catch.

seanmleonard (approved):
LGTM

PR #967, opened 2024-06-27
Fix WADs with spaces in their filename not compatible with callvote map

bcahue:
Because `callvote map` is handled differently than other map advancement functions, the C_QuoteString function was missing for changing map based on maplist index, such as `callvote map` or `callvote randmap`, but not `callvote nextmap`, since that advances to the next map, and handles the WADs properly, which is why `callvote nextmap` is not affected. This caused `callvote map` to give the following error message when trying to callvote to a wad with spaces:

image


leximayfield (approved):
Nice catch, LGTM.

PR #968, opened 2024-06-27
Make mapinfo parser less strict, allow BTSX_E1A.WAD/BTSX_E1B.WAD to play

bcahue:
Our mapinfo parser makes assumptions based on mapinfo version. For old mapinfo, our parser assumed equals was not allowed, which it is in other ports. This change may be rolled to all of our lower mapinfo MI functions, but it was most noticeable with the MIType_CompatFlag function. This function will now work if there's an equals or not. I changed doEquals to better describe what the intention of that boolean is -- a signal that this block came from a new style mapinfo that requires curly braces, while the old one is the old "hexen" style mapinfo that ZDoom extended before their new format.

I also changed MIType_Sky to read (but not implement) sky scrolling, again ignoring the mapinfo "version" and working if there's an equals or not.

With all these changes, BTSX E1 is now playable in Odamex.

### DIFF
--- a/common/g_mapinfo.cpp
+++ b/common/g_mapinfo.cpp
@@ -969,17 +969,12 @@ void MIType_Sky(OScanner& os, bool doEquals, void* data, unsigned int flags,
 
 	*static_cast<OLumpName*>(data) = os.getToken();
 
-	// Scroll speed
-	if (doEquals)
-	{
-		os.scan();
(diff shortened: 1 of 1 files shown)


rakohus:
Functionally, seems good

leximayfield (approved):
LGTM

Issue #810, opened 2022-12-07
[BUG] MAPINFO enforces 8 char limit on all types of parameters in a 'next' token


bcahue:
**Describe the bug**
Odamex will throw an error if it encounters a MAPINFO next token that isn't "EndSequence" and is higher than 8 characters. MAPINFO allows this to send players into an intermission after an EndSequence, and that intermission may have more than 8 characters.

**Build that the bug occurred in**
Odamex 10.3.0 gb28e-1

**To Reproduce**
Load a wad with a MAPINFO lump only that uses LANGUAGE lump string tokens, such as drlzoros.wad (https://www.doomworld.com/idgames/themes/xmas/drlzoros)

**Expected behavior**
Odamex to accept intermission tokens in a next token that's larger than 8 characters.

**Screenshots, NetDemos, & Crash Dumps**
N/A

**Additional context**
N/A

electricbrass:
Adding on to this: It seems to only happen if loading as a command line argument or from the in-game console. Loading from the GUI does not throw this error.

bcahue:
Fixed with #970

PR #970, opened 2024-07-01
More mapinfo fixes

bcahue:
This patch fixes bug #810 by stubbing out every possible identifier for map functions like `next`, and handling a map if any of the known identifiers aren't found.

This also will allow a float for scrolling sky, which is allowed for MAPINFO, as well as integers.

### DIFF
--- a/common/g_mapinfo.cpp
+++ b/common/g_mapinfo.cpp
@@ -789,7 +789,7 @@ void MIType_MapName(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
 	{
 			// todo
 			if (newStyleMapInfo)
-			MustGetStringName(os, ",");
+				MustGetStringName(os, ",");
 
 			os.mustScan();
 	}
@@ -797,7 +797,7 @@ void MIType_MapName(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
 	{
 			// todo
 			if (newStyleMapInfo)
-			MustGetStringName(os, ",");
+				MustGetStringName(os, ",");
 
 			os.mustScan();
 	}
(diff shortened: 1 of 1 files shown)


common/g_mapinfo.cpp
```diff
@@ -785,31 +785,78 @@ void MIType_MapName(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
...
 
-	if (IsIdentifier(os))
+	if (os.compareTokenNoCase("EndPic"))
 	{
-		if (os.compareTokenNoCase("EndPic"))
-		{
 			// todo
 			if (newStyleMapInfo)
-				MustGetStringName(os, ",");
+			MustGetStringName(os, ",");
```
leximayfield: This is a very suspect unbraced block spacing change.

common/g_mapinfo.cpp
```diff
@@ -785,31 +785,78 @@ void MIType_MapName(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
...
-		}
-		else if (os.compareTokenNoCase("EndSequence"))
-		{
+	}
+	else if (os.compareTokenNoCase("EndSequence"))
+	{
 			// todo
 			if (newStyleMapInfo)
-				MustGetStringName(os, ",");
+			MustGetStringName(os, ",");
```
leximayfield: This is a very suspect unbraced block spacing change.

leximayfield (changes requested):
There are some unbraced `if`'s that have spacing that looks very bad and misleading.  Could you either fix the spacing or add `{` and `}` to them?

common/g_mapinfo.cpp
bcahue: Fixed, thanks for double checking that, that was not intentional

common/g_mapinfo.cpp
bcahue: Fixed, thanks for double checking that, that was not intentional

leximayfield (commented):
LGTM

leximayfield (approved):

Issue #920, opened 2023-10-21
[FEATURE] Allow MAPINFO "sky2" and "doublesky" commands


thefatwilly:
**Is your feature request related to a problem? Please describe.**
Odamex to allow the mapinfo variables of "sky2" and "doublesky" to give mappers that want to have loads of variety in sky textures to be able to save some space in a wad by utilising the ability to 'graft' together sky textures in the MAPINFO lump. I.E. large map compilations and various CTF projects (like corectf) could seriously benefit from this and make downloading chunky wad files a little less tedious.

rakohus:
@thefatwilly Could you provide some example wads that use these features?

thefatwilly:
This is apparently a rare zdoom feature to be used, so I will have to generate a wad to show. Otherwise most mappers simply use the skybox feature.

rakohus:
![image](https://github.com/odamex/odamex/assets/22926877/271a9644-b0dc-45b7-b921-ecfb25b51ec1)
 Odamex is not currently a huge fan of transparency on the sky, producing a tutti fruitty effect and not rendering the second sky. Afterglow was able to provide me with a test wad. See attached test wad.

[ikdoublesky-test.zip](https://github.com/user-attachments/files/16047521/ikdoublesky-test.zip)

rakohus:
https://github.com/odamex/odamex/assets/22926877/8ec091c1-f6ed-4081-9bd1-8fea163aabdf

This is how it looks in GZ on the software renderer

rakohus:
Possibly relevant here:

GooberMan: Hexen skies don't use normal column based transparency, they treat palette index 0 as transparent
GooberMan: That'd be where the tutti fruitti comes from

bcahue:
I have this almost solved by seeing how it was done in the ZDoom 1.22 source, the last bit is, as GooberMan mentioned, is making a composite sky by comparing the posts in a column and only allowing sky2 to show thru if the sky1 color isn't 0 (cyan).

Here's what I have so far, but it makes the sky all 1 color before overflowing the buffer with skypost because of my laughable understanding of making this work with tallposts:

```c++
// create composite of both skies
byte composite[MAXWIDTH]; // column replacement
byte* skypost;
byte* skypost2;
byte* dest;
int count;
int top;
int bottom;

top = dcol.texturefrac >> FRACBITS;
bottom = (dcol.texturemid + (dcol.yh - centery) * dcol.iscale) >> FRACBITS;
count = bottom - top + 1;

[...]

PR #972, opened 2024-07-05
Restore doublesky and sky2

bcahue:
(Depends on changes from #971)

This patch restores doublesky and sky2 functionality. Using the test wad in #920 (thanks Afterglow!) I was able to recreate this functionality while working with our `tallpost_t` paradigm of detecting transparency and removing it to compress the column. Every time the sky visplane function is called, which can sometimes return thousands of columns at once, we needed to be able to walk the first sky texture and replace the pixel with the 2nd texture if the 1st wasn't in a valid range of topdelta + length compared to current pixel. We also check that data for '\0' (transparency in non-converted patches) and replace the pixel with sky2 data if present to work with both forms of "transparency."

Before:
344521955-271a9644-b0dc-45b7-b921-ecfb25b51ec1

After:
https://github.com/odamex/odamex/assets/2531014/4e6a0180-4d7e-42a9-91d1-eac42ff42ae9

And here's a test wad I used to test sky2:

[sky2test.zip](https://github.com/user-attachments/files/16112577/sky2test.zip)

### DIFF
--- a/client/src/cl_level.cpp
+++ b/client/src/cl_level.cpp
@@ -545,10 +545,19 @@ void G_DoLoadLevel (int position)
 	// [ML] 5/11/06 - remove sky2 remenants
 	// [SL] 2012-03-19 - Add sky2 back
 	sky1texture = R_TextureNumForName (level.skypic.c_str());
+	sky1scrolldelta = level.sky1ScrollDelta;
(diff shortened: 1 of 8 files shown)


fuzztooth (approved):
Let's be bold

Issue #973, opened 2024-07-06
[BUG] Stuck in infinite loop in function I_BuildPixelFormatFromSDLPixelFormatEnum because of low resolution


Rudicito:
**Describe the bug**
When running the game at small and unusual screen resolutions like 320x200 or 400x300 (not the in game resolution), the window doesn't appear after selecting the WAD file.

When looking with gdb, it gets stuck in a loop within the function `I_BuildPixelFormatFromSDLPixelFormatEnum`, 
in `client/sdl/i_video_sdl20.cpp`, line 773

```
static void I_BuildPixelFormatFromSDLPixelFormatEnum(uint32_t sdl_fmt, PixelFormat* format)
{
	uint32_t amask, rmask, gmask, bmask;
	int bpp;
	SDL_PixelFormatEnumToMasks(sdl_fmt, &bpp, &rmask, &gmask, &bmask, &amask);

	uint32_t rshift = 0, rloss = 8;
	for (uint32_t n = rmask; (n & 1) == 0; n >>= 1, rshift++) { }
	for (uint32_t n = rmask >> rshift; (n & 1) == 1; n >>= 1, rloss--) { }

	uint32_t gshift = 0, gloss = 8;
	for (uint32_t n = gmask; (n & 1) == 0; n >>= 1, gshift++) { }
	for (uint32_t n = gmask >> gshift; (n & 1) == 1; n >>= 1, gloss--) { }

	uint32_t bshift = 0, bloss = 8;
	for (uint32_t n = bmask; (n & 1) == 0; n >>= 1, bshift++) { }
	for (uint32_t n = bmask >> bshift; (n & 1) == 1; n >>= 1, bloss--) { }

	// handle SDL not reporting correct value for amask
	uint8_t ashift = bpp == 32 ?  48 - rshift - gshift - bshift : 0;
	uint8_t aloss = bpp == 32 ? 0 : 8;

	// Create the PixelFormat specification
	*format = PixelFormat(bpp, 8 - aloss, 8 - rloss, 8 - gloss, 8 - bloss, ashift, rshift, gshift, bshift);
}
```

Precisely:

`for (uint32_t n = rmask; (n & 1) == 0; n >>= 1, rshift++) { }`

The problem here is that `rmask = 0`, so this loop is endless, `rshift` increment to infinity.

Looking at the code of this function, I think that means the `sdl_fmt` variable is in fault here.

[...]

PR #977, opened 2024-07-08
Changed game window initialization from 640x480 to 320x200

Rudicito:
Partially solved #973 
(I remade a pull request, sorry, I messed up with git)


bcahue:
This looks good to me, I still need to test whether going below 320x200 causes that code to fire regarding 320x200 being the highest resolution, but you are correct in that the best way to fix this is to error if the starting window is too small for SDL.

bcahue (approved):
Looks good.

PR #982, opened 2024-07-26
Fix Static Init Damage

### LINKED ISSUE #536: [BUG] Damage static inits broken in Odamex 10

bcahue:
Fixes #536, this was a bug in how fast damage over 50 would be applied. Static init damage over 50 should be applied per tic instead of gametic & 31.


rakohus:
Tested and confirmed this concludes the necessary fixes to close #536

leximayfield (approved):
LGTM

Issue #924, opened 2023-11-19
[BUG] Support wider `STBAR` status bar lumps as DSDA / GZ do (1x1.wad)


Sbzro12345:
**Describe the bug**
Status bar when loading 1x1.wad looks visually glitched out.

**Build that the bug occurred in**
v10.4.0 Release version. Commit ID 46e0e1d7e6fcb8f0326884ba3331af92a3399bbc of stable branch.

**To Reproduce**
Load 1x1.wad with Odamex

**Expected behavior**
In DSDA-Doom, this is what the status bar looks like.
![image](https://github.com/odamex/odamex/assets/72307706/c61dec3c-7105-4e60-961f-328259f832de)
![image](https://github.com/odamex/odamex/assets/72307706/e678406b-5cc0-4ca4-b8b4-1d36b8c3ef8e)

**Screenshots, NetDemos, & Crash Dumps**
What shows in Odamex v10.4.0
![Odamex_COOP_20231119_032310](https://github.com/odamex/odamex/assets/72307706/6e85f019-a0ea-48e7-bd24-81def2b0691c)

rakohus:
This is caused by a `STBAR` lump that does not conform to the expected dimensions for a Doom wad. This must be a situation where DSDA supports an ultrawide status bar where Odamex, as of now, does not. The `STBAR` in 1x1.wad is 428x32, while the standard Doom `STBAR` is 320x32. 

Therefore, this is not a bug, but is actually a feature request to support wide status bars. I will amend the title of this ticket and make the name more appropriate. Thank you for reporting.

rakohus:
Related to bug #327.

bcahue:
Considered fixed with #983

Issue #671, opened 2022-05-19
[BUG] When using the secret exit in Registered/Ultimate DOOM's E2M5, the Intermission map resets to base E2M1 status.


Acts19quiz:
**Describe the bug**
When using the secret exit in Registered/Ultimate DOOM's E2M5, the E2M8 tower on the Intermission map resets to base completion status, and the E2M9 tower doesn't appear like it's supposed to.

**Build that the bug occurred in**
v0.9.2 (ge0d79-6829)
...
v10.1.0 (g0a84-8300)

**To Reproduce**
Use the secret exit in Registered/Ultimate DOOM's E2M5.

**Expected behavior**
For the main E2M8 under construction tower to have its intended status of completion, and for the E2M9 hidden tower to phase onto the map with an animation sequence.

**Screenshots, NetDemos, & Crash Dumps**
[Screenshots of E2M9 Intermission.](https://github.com/odamex/odamex/files/8732608/Intermission_E2M9_screenshots.zip)

**Additional context**
This happens on both Registered and Ultimate DOOM. It happens regardless of screen resolution or Widescreen status.

Issue #669, opened 2022-05-19
[BUG] Turning on "WIDESCREEN" improperly scales the position of Intermission graphics.


Acts19quiz:
**Describe the bug**
Turning on "WIDESCREEN" (OPTIONS>SET VIDEO MODE) improperly scales the position of Intermission graphics, specifically WISPLAT and WIURH#. They will move from their intended original positions outward towards the left and right edges based on their proximity to the center. Close to the center, they will barely move. The more outward left or right the graphics, the farther they will move. The splotches will not be near the buildings and landmarks representing their respective levels. In the worst cases, they will bleed past the edges of the episode background map graphic into the black boarders.

**Build that the bug occurred in**
v0.9.2 (ge0d79-6829)
...
v10.1.0 (g0a84-8300)

**To Reproduce**
Turn on WIDESCREEN on any of Registered/Ultimate DOOM's first three episode. You can do it before hitting the exit, or even during the Intermission itself. Any WISPLAT or WIURH# graphics will readjust position.

**Expected behavior**
Intermission graphics (WISPLAT and WIURH#) should keep their positions on the WIMAP# graphics, not adjust on a ratio relative to the screen ratio and distance from the center.

**Screenshots, NetDemos, & Crash Dumps**
[Screenshots with Widescreen enabled and disabled (resolutions are different, but this has no impact on the result).](https://github.com/odamex/odamex/files/8721696/Widescreen_Intermission_screencaps.zip)

**Additional context**
Screen resolution seems to have no bearing. The issue seems specific to whether "WIDESCREEN" is enabled or not. The map animations also don't seem to be affected.

bcahue:
Fixed as part of the fixes packaged with #983

Issue #668, opened 2022-05-19
[BUG] Graphic WIURH1 (right arrow "YOU ARE HERE") doesn't display.


Acts19quiz:
**Describe the bug**
The right arrow version of the "YOU ARE HERE" Registered/Ultimate DOOM graphic (WIURH1) doesn't display in the intermission screens that it should. Instead, only a red splotch is displayed.

**Build that the bug occurred in**
v0.8.#
...
v10.1.0 (g0a84-8300)

**To Reproduce**
Through normal gameplay, enter any map that usually uses WIURH1 (E3M4, E3M6 and E3M9). -warp and idclev skip the intermission screen and bug.

**Expected behavior**
The WIURH1 graphic to display when entering the three named maps.

**Screenshots, NetDemos, & Crash Dumps**
[Screenshots of Odamex and Chocolate Doom on an E3 Intermission.](https://github.com/odamex/odamex/files/8721647/WIURH1_screencaps.zip)

**Additional context**
Using either Registered or Ultimate DOOM makes no difference. Left arrow "YOU ARE HERE" (WIURH0) displays correctly.

PR #983, opened 2024-08-02
[FEATURE] Widescreen asset support

bcahue:
Again, apologies for the long PR but this is the summer of Odamex. On the menu _du jour_ is widescreen asset support, as well as scaled screenblocks and finale flats to enhance the widescreen experience. This PR uses (abuses?) WindowSurfaces in order to achieve widescreen drawing to crop the edges of the patch -- something our patch drawer cannot currently do.

We now support the following widescreen assets:
TITLEPIC
INTERPIC
DOOM 1 Episodes 1-3 WIMAPs
BOSSBACK
HELP/1
STBAR
CREDIT
ENDPIC
VICTORY2
PFUB1/2 aka BunnyScroll rewritten to use canvases

Also included is a cornucopia of fixes, which I will list here:

#327 - Added widescreen asset support, but did not touch the patch drawers.
#275 - Widescreen as well as transparent help is now supported
#924 - Related to #327, now the status bar in 1x1.wad works.
#669 - This is solved by drawing splats on their own surface, and then to the screen, to get proper scaling and not be overwritten by map animations
#671 - Solved due to an error in calculating current map
#668 - Solved due to an error in calculating how big the surface is before blitting to the screen

Also, as mentioned above, also added is both scaled screenblocks, as well as support for high definition flats in screenblocks. The screenblocks flat is drawn to a 320x200 surface, then scaled while preserving aspect ratio to the destination viewport. In order to keep aspect ratio, this may be a little bigger than the viewport in width or height -- this part gets cropped off.

This makes wads like PUSS25_LUNACY.wad display a better hud:

image

Scaled screenblocks in DOOM/DOOM2:

image

image

Test wads:
[odawidewads.zip](https://github.com/user-attachments/files/16465304/odawidewads.zip)

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -131,6 +131,11 @@ endif()
 # Client target
 if(TARGET SDL2::SDL2 OR TARGET SDL::SDL)
 
+  if (CMAKE_COMPILER_IS_GNUCXX)
+    message(STATUS "GNU Detected, forcing GCC to link math.h")
+    target_compile_options(odamex PRIVATE -lm)
+  endif()
+
   if (NOT GCONSOLE)
     # Textscreen
     set(TEXTSCREEN_LIBRARY "textscreen")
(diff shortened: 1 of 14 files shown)


fuzztooth (approved):

fuzztooth:
LET'S BE BOLD!

PR #984, opened 2024-08-02
[FEATURE] Monster respawning time-domain tuning 

cuantar:
Here is my monster respawning time-domain tuning patch.  My server at doom.catenarygrove.net has had this patch applied for some time, in various forms, as I tested and improved it.

What does it do, you wonder?  Well.  We get four new `CVARS`, complementing the existing `BOOL` `monstersrespawn`:
  - `BOOL` `monstersrespawn_tuning`: A gate toggle that is off by default.  When it is off, the previous Odamex behavior takes place.
  - `WORD` `monstersrespawn_time`: The time after death, in seconds, at which monsters will try to respawn.
  - `WORD` `monstersrespawn_period`: The time to wait, again in seconds, after a failed respawn attempt before trying once more to come back and kill you.
  - `FLOAT` `monstersrespawn_chance`: The raw probability that any given respawn check succeeds.  Like any well-defined probability, it is a number between 0 and 1.
  
Default values are set to attempt to replicate vanilla Nightmare behavior.

Note:  Cyberdemons and (especially) Archeviles DO NOT RESPAWN AT ALL, by popular request.  Most WADs are a pain in the ass when they do.

Also note: `monstersrespawn` must be `1` in order for any of this other code to be active, regardless of the state of `monstersrespawn_tuning`.


cuantar:
I will rebase this and drop the first commit.
(Hm --- I can't seem to do this)

bcahue (commented):
This modifies a part of the code that affects vanilla compatibility, so I'll need to test this build using our OdaTests package to run them against our standard suite of vanilla demos.

cuantar:
> This modifies a part of the code that affects vanilla compatibility, so I'll need to test this build using our OdaTests package to run them against our standard suite of vanilla demos.

The functionality is turned off by default, so be sure to turn it on with `sv_monstersrespawn 1` and `sv_monstersrespawn_tuning 1` .

cuantar:
lol, apologies for the strange branch name.  My _github-fu_ can only go up!

cuantar:
I've cleaned up this PR so it's nicer to deal with, and rebased the patch onto 10.6.0.

cuantar:
Rebased onto 11.0.0 --- at least, that is my intention.  Seems it didn't happen yet because I neglected to push the branch.

cuantar:
Rebased on `stable` again, after Odamex-11.1.0 release.

cuantar:
Rebased on `stable` for `odamex-11.1.1`.

PR #990, opened 2024-08-05
Adjust code to be gender-neutral

VoxelTek:
Changing code comments to replace referring to the player as "he" or "him" with "they" or "them". Additionally, renamed the "Cyborg" gender to simply "other".

In summary:
- `GENDER_NEUTER` was renamed to `GENDER_OTHER`
- "Cyborg" gender option changed to "Other"
- Code comments referring to the player as "he" have been made gender-neutral
- Gender-neutral pronouns changed from it/its to they/them


VoxelTek:
image

cuantar:
Ahoy there!  Nobody else has commented on this, but I saw it and wanted to add my thoughts.  

I think that, in the interest of technical clarity, _in the comments in the source code in particular_ it would be more appropriate to change all of the personal pronouns into language like "the player" than to substitute "they".

I've always liked the Cyborg option --- despite not identifying as non-binary, personally --- so I propose that instead of changing it, we keep it and add a new option that looks exactly as you've suggested in your PR.  You certainly are right that non-binary people are not Cyborgs; but Cyborg might indeed be something different, too.

What do you think?

VoxelTek:
[@cuantar ](https://github.com/odamex/odamex/pull/990?notification_referrer_id=NT_kwDOAzFLm7QxMTgwNjY0MTA4Njo1MzU2MjI2Nw#issuecomment-2287389825)
> I think that, in the interest of technical clarity, _in the comments in the source code in particular_ it would be more appropriate to change all of the personal pronouns into language like "the player" than to substitute "they".

That seems perfectly okay, and I'd be all for that. Only issue is that often, in the comments I changed, "the player" was already used earlier in the sentence. For example:
> `// GhostlyDeath -- Sometimes if the player spectates while they are falling down they squat`

Changing "they" to "the player" would leave the comment looking like this:

`// GhostlyDeath -- Sometimes if the player spectates while the player is falling down the player squats`
And naturally, this...doesn't sound too good. I'm all for simply referring to the player as "the player", but I feel that in cases where it isn't possible, "they" works fine as a substitute.


> I've always liked the Cyborg option --- despite not identifying as non-binary, personally --- so I propose that instead of changing it, we keep it and add a new option that looks exactly as you've suggested in your PR. You certainly are right that non-binary people are not Cyborgs; but Cyborg might indeed be something different, too.

Sounds good to me! From what I saw in the code, this shouldn't be too difficult to implement. I presume that the cyborg option should also revert to "it/its" pronouns?

In the meantime, I'll get started on trying to implement this. I'll also try and change "they" to "the player" whenever appropriate.

Thanks for your feedback!

roselorr:
friend of vox who has been seeing what's going on here - i agree that neutral and cyborg should be allowed as distinct options, and i'd like to add that it'd be most ideal to have a system in which players can input their own pronouns, though i understand that's kind of outside the scope of this PR and would require a more fundamental gender system rewrite.

cuantar:
Custom pronouns would be a good thing to have.  I wonder how hard adding that'd be, really.  

@VoxelTek I see that you're being careful with language already, in the interest of clarity.  Thanks!  I agree that in the cases where a pronoun is necessary, it's fine (and clear enough) to use "they".  Cool.  :slightly_smiling_face:

VoxelTek:
> Custom pronouns would be a good thing to have. I wonder how hard adding that'd be, really.

On that note, actually, it seems like it shouldn't be too hard, given a good understanding of the code. I, however, do not have that, and as such am currently trying to figure out how a bunch of the pronoun code works. It seems like it'd be simple enough to add additional genders and new pronouns to go along with them **in the code**, but I believe many more changes would be required to be able to do this from in-game, or to have pronouns independent of selected gender.

> I agree that in the cases where a pronoun is necessary, it's fine (and clear enough) to use "they". Cool. 🙂

Yeah, unfortunately due to the nature of the English language, pronouns are sometimes necessary, and I feel "they" would be the best fit in this case.

VoxelTek:
Cyborg gender has been re-added! Additionally, during this process, I realised I had made a small mistake when first modifying the code, as I didn't properly understand that the lengths of each pronoun were stored and used.

VoxelTek:
I can confirm, both client and server work perfectly fine with the custom build. I have been unable to test pronoun functionality, but I have no reason to believe it would not work. It is worth noting that there may be issues if a client were to connect to an outdated server, however, I believe the code has adequate protection for this situation.

leximayfield:
I appreciate this PR, however I think it covers a lot of unrelated concerns and could be split up.

- ZDoom calls the four genders "Male", "Female", "Neutral", and "Object", in that order.  I think it would be wise to follow ZDoom's guidance on this, even if it breaks protocol - we have a `protobreak` branch just for those sorts of changes. 🙂  It is also unclear how ZDoom handles sound effects for neutral vs object - a perusal of [their SNDINFO lump](https://github.com/rheit/zdoom/blob/master/wadsrc/static/filter/game-doomchex/sndinfo.txt) only seems to show "male", "female" and "other."
- Changing the text in the language lump sounds like a good idea at first, but the grammatically suspect way it's phrased is actually [a holdover from ZDoom itself.](https://github.com/rheit/zdoom/blob/d44976175256f3db8ec61cca40f1267cca68967d/wadsrc/static/language.enu#L651)  Not sure if changing it would break anything expecting the original string, and in any case this change is worthy of its own PR.
- Changing comments is tricky due to the way it complicate merges, and I tend to avoid doing so in code that I'm not directly touching for other reasons..  Believe me, there is actually more than a decade of `[AM]`'s that I would love to change to `[LM]`'s, but I don't know if it's such a great idea or not.

cuantar:
My _git-fu_ is still growing (particularly regarding merging, as I've mostly hacked alone where things were simpler) and I'm curious about your third point above:

> Changing comments is tricky due to the way it complicate merges

It makes sense that there would be _some_ effect from changing many lines in many files.  But, if in general nobody else is touching most of the codebase, then the chance of a collision with someone else's PR is small.  Will you elucidate further, please?

VoxelTek:
> Changing the text in the language lump sounds like a good idea at first, but the grammatically suspect way it's phrased is actually [a holdover from ZDoom itself.](https://github.com/rheit/zdoom/blob/d44976175256f3db8ec61cca40f1267cca68967d/wadsrc/static/language.enu#L651)  Not sure if changing it would break anything expecting the original string, and in any case this change is worthy of its own PR.

Yeah, thinking about it now it would probably be smarter to move it to a separate PR, it's just a small grammatical thing I noticed and thought I should fix. I'll revert the change and maybe make another pull request.

> ZDoom calls the four genders "Male", "Female", "Neutral", and "Object", in that order. I think it would be wise to follow ZDoom's guidance on this, even if it breaks protocol - we have a protobreak branch just for those sorts of changes. 🙂 It is also unclear how ZDoom handles sound effects for neutral vs object - a perusal of [their SNDINFO lump](https://github.com/rheit/zdoom/blob/master/wadsrc/static/filter/game-doomchex/sndinfo.txt) only seems to show "male", "female" and "other."

On the note of that, gendered sound doesn't appear to be a thing. From what I can understand, it seems someone ("joek"?) made a change a long long time ago (`client/src/s_sound.cpp:774`), and simply replaced all the genders in regards to sound with just "male". I presume this means that, no matter the gender selected, the sound effects played will be those classified as "male". 

In terms of the gender options, I mean, I'm open to following in the footsteps of ZDoom? As long as there's a neutral option available, I don't particularly care what it's called, as long as there's an option that isn't just "cyborg".

> Changing comments is tricky due to the way it complicate merges, and I tend to avoid doing so in code that I'm not directly touching for other reasons.. 

Hmm, I really don't like the idea of leaving comments in the code that refer to the player as "he" or "him". I don't see how this could complicate merges, but at the same time, I've never really dealt with this sort of situation before.

It's also worth noting I added a couple comments to other parts of the code. For example, in the above mentioned issue with all sound effects being "male", I added the following descriptive comment:
```
// I have no idea what this is here for, but I can only suspect
// that it's some sort of bodge due to not having audio for the other gender options?
// Someone smarter than me should fix this
```

This was mostly cause some of the things I had to change were unintuitive and undescribed, and it took me a while to work them out myself. Not to mention, a lot of these changes happened before the start of commits in this GitHub repo, so I couldn't even backtrack and read the relevant commit message. However, this naturally isn't particularly necessary, and my comments can probably just be removed.




Anyway, I'll make some changes to my PR. Lemme know what I should change, and I'll be happy to.

VoxelTek:
@AlexMax 
> ZDoom calls the four genders "Male", "Female", "Neutral", and "Object", in that order. I think it would be wise to follow ZDoom's guidance on this, even if it breaks protocol 

I've made a [branch](https://github.com/VoxelTek/odamex/tree/zdoom-genders) on my fork, where I've changed the gender options to "male", "female", "other" and "object". Not exactly how ZDoom does it, but a lot closer.

> It is also unclear how ZDoom handles sound effects for neutral vs object - a perusal of [their SNDINFO lump](https://github.com/rheit/zdoom/blob/master/wadsrc/static/filter/game-doomchex/sndinfo.txt) only seems to show "male", "female" and "other."

I mentioned it in my last comment, but it seems Odamex uses the male SFX for all players, regardless of selected gender. _Maybe_ there's no audio data for other options???? I have no clue, and someone else should probably look into this more.

> Changing the text in the language lump sounds like a good idea at first, but the grammatically suspect way it's phrased is actually [a holdover from ZDoom itself.](https://github.com/rheit/zdoom/blob/d44976175256f3db8ec61cca40f1267cca68967d/wadsrc/static/language.enu#L651) Not sure if changing it would break anything expecting the original string, and in any case this change is worthy of its own PR.

I've undone said change, and I might make it a separate pull request in future.

> Changing comments is tricky due to the way it complicate merges, and I tend to avoid doing so in code that I'm not directly touching for other reasons.. Believe me, there is actually more than a decade of `[AM]`'s that I would love to change to `[LM]`'s, but I don't know if it's such a great idea or not.

I have also now undone all the changes to existing comments, and I'll probably make a separate PR for those in future, however, I have left my added comments in. Mostly just so that other people can maybe understand what the code does better. If these are also problematic, I'll remove them.

rakohus:
> 
> I mentioned it in my last comment, but it seems Odamex uses the male SFX for all players, regardless of selected gender. _Maybe_ there's no audio data for other options???? I have no clue, and someone else should probably look into this more.
> 

This code _is_ currently broken in Odamex, but other ZDoom ports dating from the late 90s up through current GZDoom _do_ observe the genders when applying that SNDINFO, so it's something we need to preserve so that it may be implemented at some point. odamex.wad does not have any defaults for the non-default Doom sounds.

rakohus:
Addressed by #1308

Issue #947, opened 2024-03-31
[BUG] MAPINFO skill property "name" looks for a lump entered after "name" rather than present a string.


Acts19quiz:
**Describe the bug**
In [MAPINFO (old format)](https://zdoom.org/wiki/MAPINFO_(old_format)), specifically when defining new skills, the `picname` property is supposed to be the name of the skill as a lump (e.g., `PicName "M_JKILL"`). The `name` property is just raw text (e.g., `name "My New Skill"`). However, when using 'name' property, Odamex will erroneously look for a lump with the text entered after `name`.  Selecting NEW GAME in the menu forcefully exits the game to the console with the following error message:

> ERROR: W_GetNumForName: &lt;string&gt; not found!
(checked in: ODAMEX.WAD, &lt;iwad&gt;, &lt;pwad&gt;)

**Build that the bug occurred in**
10.4.0 G46E0E1-8728

**To Reproduce**
Make a MAPINFO with a new or redefined skill, using the `name` property instead of `picname`.

**Expected behavior**
For Odamex to display text strings in the menu matching what's given in MAPINFO for skill names in place of `picname` lumps.

**Screenshots, NetDemos, & Crash Dumps**
This is what the below test WAD looks like in ZDoom:
![MAPINFO_skill](https://github.com/odamex/odamex/assets/63941237/7784a36b-d072-4bb7-a92c-2801d92aa3ed)

**Additional context**
A demo WAD:
[mapinfo_skill.zip](https://github.com/odamex/odamex/files/14817062/mapinfo_skill.zip)

bcahue:
Odamex currently doesn't support text as an entry for `episode` in UMAPINFO, or `name` for `skill`. It's stubbed out for episode:

```c++
void SetupEpisodeList()
{
	for (int i = 0; i < episodenum; ++i)
	{
		if (EpisodeInfos[i].fulltext)
		{
			// Not implemented
	 	}
	 	else
	 	{
			strncpy(EpisodeMenu[i].name, EpisodeInfos[i].name.c_str(), 8);
	 	}
  
	 	EpisodeMenu[i].alphaKey = EpisodeInfos[i].key;
  	}
}
```

but puts the `name` untranslated into the variable meant for a `picname`:

```c++
void SetupSkillList()
{
	NewDef.lastOn = defaultskillmenu;

	int i = 0;
	for (; i < skillnum; ++i)
	{
		if (SkillInfos[i].pic_name.empty())
		{
[...]

PR #996, opened 2024-08-18
Draw MAPINFO "name" field for skills and episodes as text instead of searching for a patch

electricbrass:
Addresses #947

image
Previously, the `name` field was stored to the same variable as `picname`. This would result in an error as no lump matching the name could be found. Now if the `name` field is present, and `picname` is not, the episode name is drawn using BIGFONT.

image
Support for `name` in episode definitions has been implemented as well.


fuzztooth (dismissed):
This seems fine to me in general, though I'm a bit confused as to why the max_episodes needs to be looped through twice in g_umapinfo.

Need another c++-oriented dev to confirm.

electricbrass:
There might be a better way to do it, but it's because if *any* episode name doesn't have a graphic, *all* episodes should use the text name instead, according to the UMAPINFO spec.

electricbrass:
So I checked and despite that being in the umapinfo spec, Woof seems to be the only port to actually have that behavior, and it both seems very inconsequential to not have it and was causing some new issues, so I changed it to not use 2 loops.

fuzztooth (approved):

Issue #814, opened 2022-12-12
[FEATURE] Show items, kills, and secrets on HUD


electricbrass:
When playing in DSDA-doom it's very nice to have the option to keep the standard status bar while being able to see kills, items, and secrets without having to open the map to check. Something similar in Odamex would be nice to have for singleplayer and coop.

![image](https://user-images.githubusercontent.com/84749759/207114830-a482f981-be07-423a-a96d-6611d392b4fe.png)
(The extended hud in FDWL)

rakohus:
That's a very popular thing to see during streams, so in my opinion, might as well rip it off wholesale. We have the font.

PR #998, opened 2024-08-21
[FEATURE] Level Stats Extended HUD

electricbrass:
Addresses #814

Odamex finally gets the extended HUD that has become so popular in recent years. Enabled by a new "Extended HUD" option in the HUD menu, it can use SMALLFONT or DIGFONT, and be arranged vertically or horizontally. Also, the numbers turn yellow when all monsters are killed/all items are collected/all secrets are found.

image
image
image
image

When using the fullscreen HUD, the level stats are always arranged vertically.

image-160
image-136

This also moves around a few parts of the automap screen when in horde mode to prevent them from being covered by the horde HUD.

| OLD | NEW |
|------|-------|
|image|image|


bcahue (approved):
Looks good, been playing with it for a week now and it seems solid.

PR #1003, opened 2024-08-24
[BUGFIX] Fix instances where negative column numbers may cause out of range memory errors

bcahue:
The fix https://github.com/odamex/odamex/commit/ee23581081e1d1917abbb941546fde23893571ea for non-power of 2 textures used a modulus calculation that famously doesn't use floor division. As such, negative numbers tend to freak it out. This fixes that.

### DIFF
--- a/common/r_data.cpp
+++ b/common/r_data.cpp
@@ -478,7 +478,7 @@ tallpost_t* R_GetTextureColumn(int texnum, int colnum)
 	if (mask + 1 == width)
 		colnum &= mask;
 	else
-		colnum %= width;
+		colnum -= width * floor((float)colnum / (float)width);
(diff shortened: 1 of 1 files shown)


common/r_data.cpp
```diff
@@ -478,7 +480,7 @@ tallpost_t* R_GetTextureColumn(int texnum, int colnum)
 	if (mask + 1 == width)
 		colnum &= mask;
 	else
-		colnum %= width;
+		colnum -= width * std::floor((float)colnum / (float)width);
```
leximayfield, line 483: I wonder if there is a faster way to do this calculation.

leximayfield (approved):
LGTM!

PR #1008, opened 2024-08-26
Add colon and space to map label

electricbrass:
Does what it says on the tin. Matches behavior of DSDA-Doom


bcahue (changes requested):
Seems to be a few errant styling changes here.

electricbrass:
I turned on the trim trailing whitespace setting in my editor because of the style guidelines, and when I asked in the discord what to do in cases like this, where it ends up changing a file with existing extraneous whitespace, I was told to go ahead and commit it, as long as its a file I actually changed.

bcahue (approved):
Ah I see, approved then

fuzztooth (approved):

PR #1011, opened 2024-08-28
Various minor control changes and fixes

electricbrass:
Many other ports and games have caps lock by default bound to toggle running, and I believe Odamex should as well. Having it as a control in the key bindings menu, I think it should also have a nicer message than `"CL_RUN" IS ENABLED/DISABLED`, so with this change it displays `ALWAYS RUN ON/OFF`.

There are also a couple other minor changes here:
`+back` is now by default bound to `downarrow` instead of `backarrow`, as the latter is not a valid key.
`+right` and `+left` no longer do nothing when auto SR50 is enabled and `+strafe` is held.

### DIFF
--- a/client/src/c_bind.cpp
+++ b/client/src/c_bind.cpp
@@ -61,6 +61,7 @@ OBinding DefaultBindings[] =
 	{"leftalt", "+strafe"},
 	{"leftshift", "+speed"},
 	{"rightshift", "+speed"},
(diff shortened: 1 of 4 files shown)


fuzztooth (commented):
Looks good to me.  Thank you for catching these.

fuzztooth (approved):

PR #1013, opened 2024-08-29
[FEATURE] Allow spawning multiplayer things in singleplayer

electricbrass:
This adds a new command line parameter, `-coop-things`, that causes multiplayer things to spawn in singleplayer. A corresponding option is added to the boot window.

To enable multiplayer things from the in-game console, set `g_thingfilter` to -1.

image


electricbrass:
I've also added support for `SpawnMulti` in custom skill definitions, such as Eviternity II's Annihilate Me difficulty.

rakohus:
I might change the verbage to "Spawn Multiplayer Things", but otherwise, good

fuzztooth (approved):

PR #1016, opened 2024-09-03
[FEATURE] Build and release a Flatpak file as part of the build pipeline

bcahue:
Flatpak is increasingly becoming the desktop software deployment method of choice for Linux on the desktop. I went ahead and made changes to the build process to facilitate the complete building of Odamex as a Flatpak. This involved provisions for Odamex to build a local copy of wxWidgets and deutex and use it for the dependencies. The Flatpak file has branching information so the user can utilize the "switching" of Flatpak branches using this.

This also includes 2 cmake variables for maintainers to specify the initial directories for Odamex and its data files, since these directories may not be apparent with `cwd` since it can run in a sandbox, like they do in Flatpak.

With this, desktop linux users can enjoy grabbing and playing same nightly builds with ease as Windows users do, as the Flatpak file is essentially a binary that works on many distros.

This doesn't provide an ARM64 solution yet, because GitHub won't provide linux arm64 to open source projects until the end of the year (or so they say).

Also, Flatpak places the binaries contained within in a sandbox, runs all binaries out of `/app/bin/` (however that directory is only viewable from within the sandbox applications) and forwards `/home` and `/usr` to `/run/host` (ditto on viewability) however Flatpak users should be aware of this. Of course, if no Odalaunch config file is found, Odalaunch will run the version of Odamex that's in the Flatpak out of the box when launching Odamex.

Finally, we can upload this to Flathub with a proper manifest file (and signed with GPG key) and distribute releases that way, making compiling Odamex on most linux distros an academic thing rather than a necessity.


rautamiekka:
Honestly, surprised this is even considered an option, since Linux builds are, evidently, already seen as completely unreasonable.

bcahue:
Closing this as its superseded by #1230

PR #1017, opened 2024-09-03
Introduce g_horde_cooldown CVAR for a server-side check to determine if a Horde wave was played too recently.

Acts19quiz:
The following is a new CVAR that enables server ops (and single-player gamers) to prevent Horde waves from being played too recently. This is to address occasional complaints of, 'Didn't we just fight this wave last round?'.

Specifically: If a wave is duplicated within a certain amount of waves (set by a new CVAR, `g_horde_cooldown`, with values from 0-10), it will be silently skipped over and a new one chosen.

When a wave is determined to be valid, it's stored in an array of strings. Then when a new wave is chosen, it's compared to what's in the array. Wave names are chosen rather than wave index numbers because two entirely different waves can have the same indices with WAD switching in play. E.g., wave 7 in OdaHorde might be different than Hordamex.

There is some special offline/online handling, like with the `hordewave [name]` command, where I don't want clients in an online game making their own lists independent from the server.

The core of the function (`HordeWaveSelector`) is a loop in a loop because if a high `g_horde_cooldown` value is used with a mod containing a low amount of waves for a given skill, I don't want it endlessly looping. Just pick an appropriate wave and get on with it.

New important functions:

- `g_horde_cooldown` (CVAR) determines how many waves back to look to determine if a wave has been played too recently.
- `hordecooldownclear` is a server command to clear the Horde cooldown list.
- `hordecooldownprint` is a server command to print the Horde cooldown list.

Values:

- `g_horde_cooldown` (CVAR) ranges from 0 to 10, and defaults to 0.
- `HORDECOOLDOWN_SIZE` is set to 10, and is maximum size of g_horde_cooldown to determine the array size.
- `HordeWaveSelector` will make 20 attempts to pick a new wave before it gives up and accepts a duplicate.

All values and names are open to suggestion and change.


bcahue (approved):
Looks good.

common/p_horde.cpp
```diff
@@ -1070,3 +1150,59 @@ BEGIN_COMMAND(hordeinfo)
...
+{
+	if (!G_IsHordeMode())
+	{
+		Printf("Can't clear the horde cooldown list outside of horde mode.\n");
+		return;
+	}
+
+	if (network_game && clientside)
+	{
+		Printf(serveronlycmd);
```
bcahue: Needs a string literal like "%s\n" or it won't compile on linux

common/p_horde.cpp
```diff
@@ -1070,3 +1150,59 @@ BEGIN_COMMAND(hordeinfo)
...
+{
+	if (!G_IsHordeMode())
+	{
+		Printf("Can't obtain horde cooldown list outside of horde mode.\n");
+		return;
+	}
+
+	if (network_game && clientside)
+	{
+		Printf(serveronlycmd);
```
bcahue: Needs a string literal like "%s\n" or it won't compile on linux

bcahue (changes requested):

common/p_horde.cpp
Acts19quiz: Done. The compulsory new line is placed at the end of the string variable.

Acts19quiz:
Nuking, please refer to #1579

Issue #991, opened 2024-08-08
[BUG] Generic Doors in the Doom map format don't activate with projectile triggers


Acts19quiz:
**Describe the bug**
Generic Doors in the Doom map format that are set to activate with projectiles don't work.

**Build that the bug occurred in**
10.5.0

**To Reproduce**
1. Make a map in the Doom format (using Boom extensions).
2. Create a generic door set to open with gunfire/projectiles.
3. Run the map and shoot at the door.

**Expected behavior**
For projectiles (either hitscans or missiles) to open doors and linedefs set to activate wiht them.

**Screenshots, NetDemos, & Crash Dumps**
[Here is an example WAD set to MAP01.](https://github.com/user-attachments/files/16538307/generic_door_lift_doomformat.zip)

**Additional context**
The door works in ZDaemon and ZDoom.

electricbrass:
This actually original Boom behavior and matches non-ZDoom based ports. Generalized linedef types with triggers other than using them require tags. However, many of the checks to prevent activation without tags are already commented out in Odamex's code. Considering how many have been remove, I would suspect there's no harm in removing this one as well.

I noticed as well that both the door and the lift, despite being repeatable, can only be activated once by shooting.

PR #1018, opened 2024-09-03
[BUGFIX] Fix Boom shootable line specials not being repeatable

electricbrass:
Additionally removes the check for the line to have a tag to address #991. Different ports vary with respect to this behavior, with Odamex itself being inconsistent between the different trigger types, so I'm open to reverting this if desired or keeping the change.

### DIFF
--- a/common/p_boomfspec.cpp
+++ b/common/p_boomfspec.cpp
@@ -442,7 +442,7 @@ bool P_CrossCompatibleSpecialLine(line_t* line, int side, AActor* thing,
 		{
 			return true;
 			//line->special = 0;
-		}						
+		}
 		break;
 
 	case 40:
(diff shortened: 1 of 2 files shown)


common/p_boomfspec.cpp
```diff
@@ -3306,8 +3306,8 @@ bool P_ShootCompatibleSpecialLine(AActor* thing, line_t* line)
 			if (line->flags & ML_SECRET) // they can't open secret doors either
 				return false;
 		}
-		if (!line->id) // e6y //jff 3/2/98 all gun generalized types require tag
-			return false;
+		// if (!line->id) // e6y //jff 3/2/98 all gun generalized types require tag
```
fuzztooth: Can these just be removed rather than commented out?

electricbrass: I commented it out to match all the other commented out ones. Wasn't sure if there was a reason for them being like that.

electricbrass: Throughout this file there's a lot of nearly identical checks, some commented out, some not. So it seems like a larger cleanup of this is needed.

electricbrass: I've removed it and the other commented out checks, as well as the remaining uncommented checks in the same function.

fuzztooth (approved):

PR #1021, opened 2024-09-07
[BUGFIX] Properly escape text displayed to confirm a skill and set SpawnFilter flags correctly

electricbrass:
When parsing custom skills using `SpawnFilter`, the skill spawn flags were not being set correctly, resulting in incorrect behavior for skill dependent spawning on any skill using a `SpawnFilter` of 2 or greater.

Additionally, text displayed when confirming a custom skill, unlike the default text for Nightmare, was not having it's escape sequences replaced. This has been fixed.

Before:
image

After:
image

### DIFF
--- a/client/src/m_menu.cpp
+++ b/client/src/m_menu.cpp
@@ -1044,7 +1044,7 @@ void M_DrawEpisode()
 	{
 		y -= (LINEHEIGHT * (episodenum - 4));
 	}
-		
+
 	screen->DrawPatchClean(W_CachePatch("M_EPISOD"), 54, y);
 }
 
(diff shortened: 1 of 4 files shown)


electricbrass:
Escaping the text appears to introduce a bug that I am absolutely baffled by. Try loading Eviternity II's Nightmare difficulty. The text to confirm the difficulty will flash onscreen briefly, then disappear. The game will be unresponsive for a short time, then clear the menus. Sometimes, instead of just disappearing, it will be replaced by the text AVDN. I'm at a complete loss for what could be causing this. Checking with a debugger shows that the exact same string is being passed to M_StartMessage as when selecting Nightmare in Doom 2, other than case. And I did try changing the case to match. It still happened. But, copying and pasting the string from the debugger and directly passing it led to it working fine. Hopefully someone can figure out what the issue is here.

fuzztooth (approved):

PR #1022, opened 2024-09-08
Add last player hype messages for Survival Co-op, Survival Horde, LMS, and CTF with lives.

Acts19quiz:
This adds hype messages for the last surviving player in game modes with lives.

In Survival Co-op or Survival Horde, if only one player remains, the surviving player gets a unique centre-set message hyping the player up with a well wish message. All other players (and the server) get a message announcing that player as the last player.

In LMS and CTF with lives, when a team loses all players but one, a general hype message is broadcast to all players of that team.

The idea of adding a new custom sound effect has also been tossed around.

https://www.youtube.com/watch?v=1eN7oq8oekw

The Majora's Mask sound effect (or something similar to it, as we probably couldn't use licensed sound effects) has been suggested.

https://www.youtube.com/watch?v=iP8LpPPSkok&t=8

A generic note similar to this could also be used.


Acts19quiz:
To add as an additional conversation piece:

The reason SV-MidPrint is duplicated at CL-MidPrint in `g-gametype.cpp` is because `g-gametype.cpp` can't see defines in `sv-main.cpp` or `sv-main.h`, but the identifier "player-t", if added to `doomtype.h`, confuses everything that references `doomtype.h`.

However, is rewriting SV-MidPrint to accept a Player ID as a basic int rather than a whole "player-t" a possible option to eliminate the need for duplicating the server function in `g-gametype.cpp`?

EDIT: I made the change. Changing the input value of the function from a "player-t" to a regular "int" (specifically, a player id) allows for it to be declared in `doomtype.h`, thus eliminating the duplicate function in `g-gametype.cpp`.

I went through and adjusted all other calls to SV-MidPrint to submit a player's id rather than the whole player-t. This seemed to have no ill effect on existing functionality, though as far as I could tell, the only pre-existing usage of SV-MidPrint was to display the MotD in online games.

Acts19quiz:
I really appreciate the auto tabs that Visual Studio added to my new lines. They're so helpful.

PR #1026, opened 2024-09-13
[BUGFIX] Set texture height for masked segments to 512 instead of 256

electricbrass:
This fixes issues with large textures such as the portal at the end of Legacy of Rust map01 and the large banners from OTEX.

Before:

image
image

After:

image
image


bcahue (dismissed):
Looks good, assuming this doesn't introduce any side effects. But other 2s midtex's look good so I say send it.

electricbrass:
After some more testing, I found that some grass in valiant renders incorrectly. I went back to the original approach, but used 512 instead of 256, and there seem to be no issues (as long as even larger textures over 512 texels tall are not used).

fuzztooth (approved):

PR #1027, opened 2024-09-13
[BUGFIX] Fix UMAPINFO par times being 35x too high

electricbrass:
Yet another small MAPINFO fix. Par times in UMAPINFO were being multiplied by the tic rate unnecessarily, resulting in incorrect par times that often would just be SUCKS.


bcahue (approved):
Looks good

Issue #752, opened 2022-08-18
[FEATURE] Maintain offsets on mirrored sprites


electricbrass:
**Is your feature request related to a problem? Please describe.**
A bug dating back to the original doom is that when sprites are mirrored, they lose their offsets. This makes some monster animations seem a bit jerky if they are not centered within the sprite.

**Describe the solution you'd like to see**
The minor sprite fixing project fixes this with adding padding to the affect sprites, so that the axis the sprites should be rotated around is actually in the center of the sprite. This could also be fixed at a code level by maintaining the offsets when sprites are mirrored. This improves the look of some monsters.

![7yGsvLY](https://user-images.githubusercontent.com/84749759/185488628-db12d48c-151c-4844-a277-6ca093ca3cec.gif)

**Describe alternatives you've considered**
None, either we do it or we don't

**Additional context**
Here is what was done in ZDaemon:
https://forums.zdaemon.org/viewtopic.php?t=17103&sid=679352c68240bd0ba0383b9cda40c419

PR #1030, opened 2024-09-13
[FEATURE/VANILLA BUGFIX] Maintain sprite offsets when flipped

electricbrass:
Addresses #752

This fixes a minor visual bug from vanilla Doom that has been fixed by the minor sprite fixing project, as well as some other ports. Mirrored sprites lose their offset, resulting in jerky animation for some enemies such as the spider mastermind. This fixes that.

If preserving the vanilla behavior is desired, this could be made optional, though I'm not sure it's really worth adding another setting for this.

Before:

https://github.com/user-attachments/assets/e43cf56e-bd75-4095-a95e-544bea8e3bac

After:

https://github.com/user-attachments/assets/26b9b397-ad91-4558-a8d0-fa35c4c1caa5


bcahue (approved):
Looks good here

Issue #712, opened 2022-07-20
[CHANGE] Expand Horde's spawner warning message to include big boss spawner (5303)


Xenaero:
**What is the existing behavior? Please describe.**
Currently, when there's not a criteria of spawners met to start a Horde wave, the following shows:

> WARNING: This map is missing Horde Monster, Horde Supply Cache, or Horde Powerup spawns. At least one of each must be present.

In fact, you can start a Horde wave without a valid boss spawner being present at all, be it the normal or thin one.

**What is the desired behavior? Why are you requesting the change?**
This warning message should be expanded to include at least one instance of the normal (big) Boss Spawner - Type 5303. It may also be wise to not initiate gameplay until a boss spawner is placed.

There have been instances with map authors where a boss spawner was mistagged to not show up for MP and wasn't active in the map during a multiplayer session, as well as an instance of only a thin boss spawner being used. In the former, the boss never spawned. In the latter, the boss would spawn if it fit the dimensions, but if it was a big boy boss, it would not be able to spawn and brick the wave.

**If applicable, describe alternatives you've considered**
None. At the minimum, there needs to be a boss spawner present or the wave will never have the opportunity to advance.

**Additional context**
Wasn't sure if this was better as a bug or a change ticket. I've attached a demonstration map with a thin boss spawner and a spiderdemon-only boss wave to illustrate the issue this can pose.

[horde-thinboss-only.zip](https://github.com/odamex/odamex/files/9147257/horde-thinboss-only.zip)

rakohus:
Agreed that it should require the normal boss spawner.

To expand on the dilemma listed above, if a boss doesn't spawn because whatever wad breaks it, _it might be better_ for the mode to eventually move on to the next wave instead of hanging in a boss-less state.

PR #1031, opened 2024-09-13
[CHANGE] Improve missing spawns warning for horde mode

electricbrass:
Addresses #712

The warning for missing spawners is now displayed if there is no large boss spawner or regular monster spawner, and it now reads:

> WARNING: This map is missing Horde Monster, Horde Boss, Horde Supply Cache, or Horde Powerup spawns. At least one of each must be present.

As an alternative to this, we could instead restrict the waves that can be selected on a map by the spawner types present. e.g. Only waves with no ground monsters other than bosses in a map containing only flying spawners (and a boss spawner). Or only waves with thin bosses in a map with only thin boss spawners.

### DIFF
--- a/common/p_horde.cpp
+++ b/common/p_horde.cpp
@@ -610,7 +610,7 @@ void HordeState::tick()
 				alive += 1;
 		}
 		if (!alive)
-		{			
+		{
 			// Start the next wave.
 			nextWave();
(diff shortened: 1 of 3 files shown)


bcahue (approved):
Looks good

PR #1036, opened 2024-09-16
[CHANGE] Refuse to download commercial PWADs as well as IWADs

electricbrass:
Odamex already refuses to download commercial IWADs. This PR makes it so that it does the same for PWADs. The errors have also been made slightly more detailed, as well as directing the player towards DOOM + DOOM II on Steam and GOG, as the old versions of nerve.wad and masterlevels.wad are the only commercial PWADs recognized here that are not part of DOOM + DOOM II, and it is more likely that a server will be running the updated versions of these wads anyway.


bcahue (approved):
This looks good to me.

PR #1039, opened 2024-09-16
Clean up references to removed cvars + remove some unused cvars

electricbrass:
The sample configs set a number of cvars that have been removed since the configs were created. There were also a few cvars that go now completely unused but were still present in the codebase. I know there are plans for a larger cvar overhaul, but I wanted to see at least some minimal cleanup of this.

The 2 variables removed are `sv_waddownload` and `sv_waddownloadcap`.


rakohus (approved):

rakohus:
lgtm

Issue #443, opened 2021-08-02
[FEATURE] Show DOS Ending Screen: Only Modified


Some1NamedNate:
This proposed setting should show the ENDOOM screen **only if the ENDOOM lump is found from PWADs**. If no PWAD has an ENDOOM lump, the Client would close immediately, ignoring an IWAD's ENDOOM.

bcahue:
This is completed and merged to `stable`.

PR #1041, opened 2024-09-17
A few minor improvements

electricbrass:
There are 3 unrelated but all very small changes here:

- Added an option to show ENDOOM only if it's from a PWAD
- Enabling `vid_640x400` disables `vid_320x200` and vice versa
- The version of `pip-action` used was unmaintained and used a deprecated version of Node. This has been changed to an updated version from an actively maintained fork.

If desired, these can be split into separate PRs.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Checkout submodules
       run: git submodule update --init --recursive
     - name: Install Python packages
-      uses: BSFishy/pip-action@v1
+      uses: insightsengineering/pip-action@v2
       with:
         packages: |
           b2sdk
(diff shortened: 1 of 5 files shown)


fuzztooth:
What is the impetus or perceived benefit of having a "PWAD only" option for endoom screen?

electricbrass:
It's something a few other ports have and was requested in #443 (forgot to link that). The idea is that IWAD ENDOOM doesn't get shown every time you exit, while also letting you see the custom ENDOOMs included in PWADs.

Personally, I don't care for seeing the ENDOOM screen every time I exit, but I don't want to miss out on those added in, say, Eviternity.

fuzztooth (approved):

PR #1048, opened 2024-09-22
[FEATURE] ID24 Interlevel Lump Support

electricbrass:
All aspects of the intermission screen backgrounds (splats, you are here, animations, e2m8 tower being built) that were previously hardcoded are now definable by JSON lumps following the ID24 specification. To replace the previous hardcoded elements, odamex.wad contains 3 new lumps defining the Doom 1 intermissions.

With this implementation, currently only episode 1 of Legacy of Rust displays properly. This is due to the current version of id1.wad being based on a buggy implementation from D+DII that has been confirmed to be fixed for the next update.

This also lays some groundwork for support of ZDoom intermission scripts. Other than perhaps `IFTRAVELLING` and `IFNOTTRAVELLING`, all intermission script commands can be converted relatively easily to the same internal representation as ID24 interlevel lumps.


fuzztooth (approved):

fuzztooth (approved):

electricbrass:
The updated version of legacy of rust from today (10/3/24) does display correctly.

Issue #621, opened 2022-03-17
[BUG] A_MonsterMeleeAttack Range Parameter Crash


rakohus:
**Describe the bug**
When using the `A_MonsterProjectile` MBF21 actor pointer, hard client crash occurs due to the following line: `A_MonsterProjectile(thing MTF_EXTRA13, 0.0, 0.0, 16.0, 8.0)`, where `MTF_EXTRA13` is not defined.

**Build that the bug occurred in**
`horde-fixes` [feead5d](https://github.com/odamex/odamex/commit/feead5d4d26279d6de01a660f9c337ef136231f3)

**To Reproduce**
Load these wads. Go to the big imp on the far left, walk in front of him so he tries to attack, he will pump fake a few times but when he finally gets his melee attack off successfully, the client will bomb

[monsters.zip](https://github.com/odamex/odamex/files/8281184/monsters.zip)

**Expected behavior**
Not crash. Throw an error in console?

**Screenshots, NetDemos, & Crash Dumps**
Crash Dump: 
[odamex_gfeea_14156_20220317T032652.zip](https://github.com/odamex/odamex/files/8281182/odamex_gfeea_14156_20220317T032652.zip)

bcahue:
This is now fixed and merged to `stable`:

![image](https://github.com/user-attachments/assets/2762a08c-f205-40f8-9e79-6823d8b05ab6)

PR #1049, opened 2024-09-23
[BUGFIX] Throw error when attempting to spawn object in MBF21 codepointers

electricbrass:
Addresses #621

Now instead of crashing if the type is undefined, an error such as the following is shown.
image


fuzztooth (approved):
We like less crashing!

Issue #945, opened 2024-03-25
[BUG] Fails to build on Gentoo with clang-17


NHOrus:
As part of preparing for future mandatory errors in gcc-14, we are rebuilding software with increased amount of warnings.

https://bugs.gentoo.org/919287

This failure happens on Gentoo with Clang-17, and most likely would happen on incoming GCC-14 

Errors of compilation:
```
/var/tmp/portage/games-engines/odamex-10.4.0-r2/work/odamex-src-10.4.0/common/m_wdlstats.cpp:181:6: error: non-constant-expression cannot be narrowed from type 'size_type' (aka 'unsigned long') to 'int' in initializer list [-Wc++11-narrowing]
  181 |             ::wdlplayers.size() + 1,
      |             ^~~~~~~~~~~~~~~~~~~~~~~
/var/tmp/portage/games-engines/odamex-10.4.0-r2/work/odamex-src-10.4.0/common/m_wdlstats.cpp:181:6: note: insert an explicit cast to silence this issue
  181 |             ::wdlplayers.size() + 1,
      |             ^~~~~~~~~~~~~~~~~~~~~~~
      |             static_cast<int>(      )
/var/tmp/portage/games-engines/odamex-10.4.0-r2/work/odamex-src-10.4.0/common/m_wdlstats.cpp:213:35: error: non-constant-expression cannot be narrowed from type 'size_type' (aka 'unsigned long') to 'int' in initializer list [-Wc++11-narrowing]
  213 |         WDLPlayerSpawn wdlplayerspawn = {::wdlplayerspawns.size() + 1, mthing->x, mthing->y,
      |                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
/var/tmp/portage/games-engines/odamex-10.4.0-r2/work/odamex-src-10.4.0/common/m_wdlstats.cpp:213:35: note: insert an explicit cast to silence this issue
  213 |         WDLPlayerSpawn wdlplayerspawn = {::wdlplayerspawns.size() + 1, mthing->x, mthing->y,
      |                                          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                          static_cast<int>(           )
/var/tmp/portage/games-engines/odamex-10.4.0-r2/work/odamex-src-10.4.0/common/m_wdlstats.cpp:280:31: error: non-constant-expression cannot be narrowed from type 'size_type' (aka 'unsigned long') to 'int' in initializer list [-Wc++11-narrowing]
[...]

Issue #826, opened 2023-01-06
[BUG] Odamex doesn't build on FreeBSD


Bill-Paul:
**Describe the bug**
I've tried compiling both Odamex 10.2.0 and 10.3.0 on FreeBSD/amd64 13.1-RELEASE using the released source tarballs, but neither one builds out of the box. In the end I was able to coerce it into building, but not without having to make some code changes. Note that FreeBSD 13.1 uses LLVM/Clang 13.0 as its native compiler and that's what I tried to use. I think these problems are not actually FreeBSD-specific, but just cases of Clang being a bit more picky (though in some cases I'd say it's justified). Note that several of them indicate that the Odamex code is not 64-bit clean (or at least not as clean as it could be).

Clang also issues many warnings, most of which seem harmless, but they're for things that probably should be fixed. (I'll include a build log to show you what I mean.)

The problems I ran into are the same in both the 10.2.0 and 10.3.0 releases. I was able to build 10.3.0 using the same patches that I made for 10.2.0.

I'll try to list the most salient issues here:

1) Configuration. I decided to use cmake to build. FreeBSD 13.1 includes cmake 3.24. I used this command to configure:

```
% mkdir build
% cd build
% cmake	\
	-DCMAKE_BUILD_TYPE=Release				\
	-DCMAKE_C_COMPILER_ID=Clang				\
	-DCMAKE_CXX_COMPILER_ID=Clang				\
	-DCMAKE_CXX_FLAGS="-isystem /usr/local/include"		\
	-DCMAKE_CXX_LINK_FLAGS=-L/usr/local/lib			\
	-DUSE_INTERNAL_PNG=True					\
	--install-prefix=/usr/local/odamex ..
```

The reason for using the -isystem command has to do with the Google protobufs library. Odamex includes its own version of protobufs in the source tarball, however depending on what packages you have installed, FreeBSD may already have the Google protobufs library and header files present under /usr/local. Unfortunately they're different versions, and they conflict rather badly. Using the -isystem flag here ensures that the Odamex build favors the header files included with the Odamex source instead of the ones packaged with FreeBSD.

The -L/usr/local/lib flag is needed so that the link command can find libX11. It's in /usr/local/lib on FreeBSD, unlike Linux which has it in /usr/lib.

This is not necessarily something that needs to be fixed, but I'm writing it down here anyway in case some other lunatic besides me wants to try building on FreeBSD too.

[...]

PR #1056, opened 2024-09-24
[BUGFIX]Squash easy warnings to reduce default warning level noise when compiling

bcahue:
The huge build logs are once again a thing of the past! This PR performs a host of changes in order to reduce the amount of warnings across Clang/GCC/MSVC.

I converted all instances of sprintf to snprintf and vsprintf to vsnprintf as MacOS is deprecating sprintf. I also added casting to all instances of an unsigned variable being stuffed into an signed one, as Clang 17 will now error on that scenario.

The centerpiece of this is `info.cpp` is now constructed properly, and doesn't put out thousands of warnings anymore.

This reduces the total warnings from thousands to dozens, and ensures the warnings that remain are actual issues that should be addressed rather than low hanging fruit noise.

Addresses #826 and #945


fuzztooth (approved):
Ain't no one gon' review all that.

Issue #756, opened 2022-08-21
[BUG] Documentation Corrections needed for Debian/Linux compiling


MarkJeronimus:
In the README.md, in the Compilation Instructions section, it states that Linux requires the packages `libsdl2-dev`, `libsdl2-mixer-dev` and `libwxgtk3.0-dev`.

My Debian-flavored OS' package manager doesn't have `libwxgtk3.0-dev`. I tried compiling it anyway, and found I needed the following packages instead:
- `build-essential`
- `libcurl4-gnutls-dev`
- `libpng-dev`
- `libportmidi-dev`
- `libsdl2-dev`
- `libsdl2-mixer-dev`
- `libwxgtk3.0-gtk3-dev`
- `zlib1g-dev`

With these packages, Odamex could compile. I don't know if the list above is exhaustive. I do different kinds of software development on my machine, so it might be that other required packages happened to be installed already.

When you follow the "this page"-named link to the wiki and go to the "Required libraries", it refers only to packages `libsdl2-dev` and `libsdl2-mixer-dev`.

I think both need to be fixed.

bcahue:
This has been fixed and now our README points to our up to date documentation which should describe how to compile on modern flavors of linux.

PR #1058, opened 2024-09-24
[BUGFIX] Point readme link to up to date documentation

bcahue:
The current README.md points to the odamex.net wiki, which is extremely old and outdated. This changes the link to the documentation for compiling Odamex that is up to date.

Addresses #756


electricbrass:
The txt readme looks like it could use the same update.

fuzztooth (approved):

PR #1069, opened 2024-10-04
Recognize 10/3/2024 versions of Doom + Doom II wads

electricbrass:
The update to Doom + Doom II included changes to many of the wad files, resulting in Odamex no longer being able to identify the updated IWADs. This PR adds the new versions to the recognized wad list.

I have 2 concerns about the naming of the wads. First is that I do not know the best way to indicate version of the Doom + Doom II wads. I have labelled them as v1.0 and v1.1 but this is purely my own numbering.

Additionally, prior to Doom + Doom II, the names for the most commonly detected IWADs fit within the window:
image
But the new versions have long names that go off the end (I assume this is true of some of the other wads versions, but they aren't ones that Odamex appears to automatically search for):
image

I feel like it would make for a cleaner experience if the names were shortened (D+DII?) or the size of the window were increased.

Changes to both of these can be made in this PR based on feedback.

### DIFF
--- a/common/w_ident.cpp
+++ b/common/w_ident.cpp
@@ -81,13 +81,22 @@ static const identData_t identdata[] = {
         100,                                // weight
     },
     {
-        DOOM2_PREFIX " (DOOM + DOOM II)",   // idName
-        "DOOM2.WAD",                        // filename
-        "09B8A6AE",                         // crc32Sum
-        "9AA3CBF65B961D0BDAC98EC403B832E1", // md5Sum
-        DOOM2_PREFIX " v1.9",               // groupName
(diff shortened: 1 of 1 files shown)


electricbrass:
Legacy of Rust and the other id1* wads have version numbers in their WADINFO files. I have updated those used here to reflect that, and removed the version numbers from the other wads so as to not have inconsistent usage of official and unofficial version numbers. The changes to the IWADs other than TNT: Evilution are almost entirely cosmetic and would seem to not need to be distinguished between, but do contain one important difference if Odamex adds support for GAMECONF in the future. The older versions have the "executable" field set to "doom1.9" while the new versions have it set to `null`.

fuzztooth (approved):

PR #1070, opened 2024-10-05
Fix endianness check for *BSD

FozzeY:
The macro needs to be defined for Open/NetBSD.
Additionally, the file should be included as `<endian.h>` in userspace https://github.com/openbsd/src/blob/master/sys/sys/endian.h#L29.


bcahue (approved):
Looks good, thanks!

PR #1071, opened 2024-10-06
Add sv_allowfov cvar

gera2342:
So that servers can allow players to change their fov without setting `sv_allowcheats 1`.


rakohus (approved):
Seems simple enough

PR #1072, opened 2024-10-07
[BUGFIX] Fix 7 second delay after initializing animations with speed of 1

electricbrass:
Animated textures/flats with a speed of 1 had a 7.3 second delay from the time the wad was loaded. This was due to an overflow from decrementing the countdown to the next frame twice before the first time it's value was checked.


rakohus (approved):
lgtm

PR #1074, opened 2024-10-18
Add weapons list to the fullscreen HUD

gera2342:
Looks like this:

20241018_09h08m13s_grim

IDK whether "DIGIFONT" or "SMALLFONT" looks nicer or where it should be positioned. Also there may be more fitting name for the cvar than *hud_weapontext*.


gera2342:
I moved the text to the right which is easier on the eyes I think and changed *DIGIFONT* to *SMALLFONT* which would be consistent with the colors the WAD uses.

20241018_12h30m36s_grim

gera2342:
Forgot about not overlapping with score/spread, now fixed
20241018_12h54m28s_grim
Also horde
20241019_14h39m13s_grim

Am I missing any other modes?

gera2342:
Maybe it's better to use 1+ and 3+ for chainsaw and SSG like in horde instead of 8 and 9?

gera2342:
20241029_23h19m31s_grim
overlaps with flags in ctf
20241029_23h34m59s_grim
fixed

rakohus:
I think it looks pretty good right now! But, it could be made more useful in the following ways, listed in what I think is most important:

1) Chainsaw needs to be 1+, SSG needs to be 3+ and follow the Horde convention. 8/9 is confusing because actually pressing those keys on the keyboard don't do anything. Most players that don't understand the underlying workings of Doom won't ever equate "9" with the SSG.

2) It would be nice if the weapon that you're currently holding was highlighted in a different color.

3) Following from the logic above, it would be nice if weapons that you are out of ammo on turned a darker color like gray.

rakohus:
Also, another thought, I _personally_ don't care for the "W" but agree that there needs to be a way to designate what that list is for. I don't think "W" immediately conveys what the list is for, which is something that should be considered

electricbrass:
I think WPN like in the earlier screenshots works well. Or it could be ARMS like the status bar.

gera2342:
Now looks like this:
20241201_11h48m18s_grim
Should I squash this pr into one commit?

bcahue (approved):

PR #1076, opened 2024-10-24
[FEATURE] Add IPC communication to Odamex to allow communication with bootstrappers and shims; Add Discord support

bcahue:
The reason you'd want Odamex to communicate with an external bootstrapper (that would launch Odamex as a child process) or a shim (which Odamex launches as a child process) is so Odamex can communicate and interact with services that otherwise would require Odamex to link against a 3rd party library that is not GPLv2 compliant, like Discord SDK, Steamworks SDK, or Epic Games Services (EOS) SDK, to name a few.

The odashim executable introduced here links against the Discord SDK in order to provide rich presence support for Odamex. This allows Odamex to report to Discord stats in game as well as map played (both available as tooltips), game mode, and player's status (in game, spectating etc):

image

I stopped short of adding the invite to join button or join/spectate button, as Odamex has no way of knowing whether a server has become passworded or not. It could mark a flag if it sends one on connect but it may cause a bad state. Otherwise I could add a launch command of odamex://-connect ip to it. Also, invite to join requires interaction to accept, which is something we can't really provide with our current UI without a lot of changes.

The shim exe won't start unless Odamex starts it. The shim exe itself stops if it doesn't receive communication from the parent in 30 seconds. The client controls sending the shim information from the console command `cl_shim`. If cl_shim is 1 and the client detects no communication from the shim in 45 seconds, the client attempts to start the shim. If the attempts fail, a backoff is issued preventing further attempts.

Also, frustratingly, all this doesn't have Discord detect the game. This means that it doesn't show up as a game in the Activity log, nor show up for the user in the bottom left as a game when launching. This requires a ticket to Discord support once the app is Verified, which requires a Team to bet set up for it. Then, support will whitelist the odamex executable.


electricbrass:
My latest changes should fix #1082

fuzztooth:
Conflicts aside, is this waiting for anything else?

electricbrass:
It hasn't been tested on mac or linux yet

fuzztooth:
Now that we have automated mac builds, can we get this one updated, conflicts resolved and some quick testing done so we can get this in?

Issue #1078, opened 2024-11-01
[BUG] Odalaunch search bar broken when using wxGTK 3.2


gera2342:
When odalaunch is linked against wxGTK 3.2, now default in many distros, the search bar is broken. On X it is impossible to type anything into it, on wayland possible but what I type is not visible.

Screenshot on wayland: I typed the word *hell*.
![20241101_05h28m46s_grim](https://github.com/user-attachments/assets/4efbbc75-ef28-4f03-b323-173033109ad0)

It works as expected when linked against wxGTK 3.0.5.1
 e.g. `cmake -DwxWidgets_SELECT_OPTIONS="--version=3.0" ..`

bcahue:
This should be fixed with the merge of #1088

PR #1088, opened 2024-12-11
Explicitly set the size of the Odalaunch search bar + Ubuntu buildgen scripts for 24.04

electricbrass:
Addresses #1078

Over the next month, GitHub Actions runners will be updating to Ubuntu 24.04. Due to this, `libwxgtk3.0-gtk3-dev` is no longer available. This PR changes the build scripts to install the new `libwxgtk3.2-dev` package instead. For reasons I don't understand, automatic sizing of the search bar does not work with wxWidgets 3.2, so it is now set to a width of 50 dialog units.

### DIFF
--- a/ci/ubuntu-buildgen.sh
+++ b/ci/ubuntu-buildgen.sh
@@ -11,10 +11,10 @@ set -x
 sudo apt update
 if [[ -z ${USE_SDL12:-} ]]; then
     sudo apt install ninja-build libsdl2-dev libsdl2-mixer-dev \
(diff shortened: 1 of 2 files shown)


electricbrass:
It looks like `libpng-dev` is no longer installed by default, so I have also added it to the list of packages to be installed.

gera2342:
Doesn't fix search bar for me

gera2342:
Oh sry, it does fix :)
Just had to rebuild with `make clean`

bcahue (approved):
This Ubuntu migration seems to be underway, approved

Issue #1098, opened 2024-12-13
[BUG] Legacy of Rust map09 Platforms to Red Key Not Raising


Hekksy:
**Describe the bug**
In Legacy of Rust map09, the player uses two switches to raise two sets of platforms to get to the red key. These platforms only raise halfway, making the map incompletable due to being unable to get the red key.

**Build that the bug occurred in**
This bug is reproducable in Odamex 10.6 as well as the latest DSDHacked branch

**To Reproduce**
Go to id1.wad map09 and noclip to the red key area and use the switches in singleplayer or online.

**Expected behavior**
Both sets of platforms should raise to match the height of the platform housing the red key

**Screenshots, NetDemos, & Crash Dumps**
https://cdn.discordapp.com/attachments/250157428455243777/1317193388994596974/Odamex_COOP_20241213_121220_UDHekksy_id1.wad_MAP09.png?ex=675dcb70&is=675c79f0&hm=4c0c3ed73a6d10154e2b745362bc90da6959ba2c3cf7f7ded3b8bea8268e4315&
Odamex

https://cdn.discordapp.com/attachments/250157428455243777/1317193276373073970/image.png?ex=675dcb55&is=675c79d5&hm=d55fb01664d7ce761fc7778b0561113c9ac13cbd8f48e8c94471975ae2450287&
Kex Engine (as intended)

Hekksy:
Confirmed fixed in Stable with co_boomphys 1

Hekksy:
Confirmed fixed with co_boomphys 1 in Stable

PR #1099, opened 2024-12-13
[BUGFIX] Destination height for floor and ceiling lower/raise to shortest texture is incorrect

electricbrass:
Addresses #1098

`P_FindShortestTextureAround` and `P_FindShortestUpperAround` check for textures >= 0. Texture index 0 is a placeholder with a height of 64, meaning these functions will never return a height above 64 unless all sides of all lines checked have a set texture. This is a bug from vanilla that was fixed in Boom, so I have the fix only when `co_boomphys` is enabled.

### DIFF
--- a/common/p_spec.cpp
+++ b/common/p_spec.cpp
@@ -1259,11 +1259,11 @@ fixed_t P_FindShortestTextureAround (sector_t *sec)
 		if (sec->lines[i]->flags & ML_TWOSIDED)
 		{
 			side = &sides[(sec->lines[i])->sidenum[0]];
-			if (side->bottomtexture >= 0 && textureheight[side->bottomtexture] < minsize)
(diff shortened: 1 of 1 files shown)


electricbrass:
On further investigation, the bug actually originates with vanilla doom, and was fixed in Boom. I'll move the fix to be part of the `co_boomphys` compatibility setting.

electricbrass:
[textureraise.zip](https://github.com/user-attachments/files/18131673/textureraise.zip)
Here is a test map. With `co_boomphys 0` both floors should raise to height 64, and with `co_boomphys 1` both should raise to height 128. The left is vanilla line type 30, and the right is a Boom generalized floor.

rakohus (approved):

rakohus:
Tested with `co_boomphys` on and off, works as expected

PR #1102, opened 2024-12-21
Restore interpolation of skies

electricbrass:
The current implementation of interpolating scrolling skies only works for the old skies without SKYDEFS. This PR restores the interpolation to SKYDEFS defined skies, including for vertical scrolling.


bcahue (approved):
This looks good to me

PR #1104, opened 2025-01-05
Implement new options for horde mode

jpernst:
I'm a big fan of invasion style game modes, so I thought I'd add a few options to Odamex's horde mode to allow different style and pacing options. By default, the new options are set such that the behavior is identical to current horde mode.

New CVAR options:

g_horde_relaxhp (default 0):
Places a hard limit on the spawnable health of each wave. This is calculated the same way as goal HP. If this value is less than goal HP, then there is no limit and monsters will spawn continuously (the default horde behavior). Setting this value equal to goal HP achieves a more traditional invasion-like play style where a wave is fully cleared before a new wave begins.

g_horde_bosspace (default 1):
Allows for changing the pace of boss fights. Only waves that are a multiple of this value will have a boss, otherwise the wave ends immediately when the goal HP is reached and the next wave begins. A value of 0 means there are no bosses at all. The default value of 1 means every wave has a boss. Setting this equal to the total number of waves means only the final wave will have a boss.

These options are especially helpful for solo play or small groups to make things more manageable, but also allow for a different kind of difficulty by increasing the spawn rate, allowing for shorter but very intense waves with periods of calm in between. I'm new to this codebase so let me know if I've missed anything.

### DIFF
--- a/common/c_cvarlist.cpp
+++ b/common/c_cvarlist.cpp
@@ -273,6 +273,12 @@ CVAR(g_horde_maxtotalhp, "10.0", "Multiplier for maximum spawned health at a tim
 CVAR(g_horde_goalhp, "8.0", "Goal health multiplier for a given round", CVARTYPE_FLOAT,
      CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_NOENABLEDISABLE)
 
+CVAR(g_horde_relaxhp, "0", "Multiplier for limit of spawnable health per wave", CVARTYPE_FLOAT,
+     CVAR_SERVERARCHIVE | CVAR_SERVERINFO | CVAR_NOENABLEDISABLE)
+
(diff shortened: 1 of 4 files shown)


jpernst:
Added a couple more options:

`g_horde_bossrelax`: default Off. When a boss fight starts, the relaxhp counter is reset to allow a boss to have backup, even if the limit has already been reached. Useful when the relaxhp limit is low to ensure a boss fight is still interesting, especially in combination with the next option
`g_horde_bosswipe`: default Off. When a boss fight is won, kill all remaining monsters on the map, allowing the next wave to have a fresh start and give the players a breather. Very satisfying after a tough fight.

jpernst:
This was a speculative, experimental patch inspired by an event that has already passed, and its value is questionable besides, so I'll just close this out.

Issue #306, opened 2021-04-24
[FEATURE] OPL Music Emulation


odamex-system:
*Bug **585**
Posted By: **Ralphis (Ralph Vickers)**
Posted On: **Tue, 04 May 2010 11:10:48 GMT**
Last updated **Wed, 11 Sep 2019 19:32:28 GMT***

Chocolate Doom recently implemented OPL Music Emulation. While it isn't perfect, it could potentially help Odamex for two reasons:

1) The cool reason. OPL Music is cool, sounds ""classic"", and is a great feature to have for a Doom engine. It also fits within Odamex's scope of providing an online experience relatively close to vanilla.

2) The practical reason. With all of the midi issues sdl_mixer causes for Windows Vista/7 users, offering the OPL alternative is a great solution that can be done without any changes to SDL_Mixer. The more people upgrade to the newer Windows that lack proper midi mapper support, the more complaints Odamex and other SDL Doom engines are going to hear. Offering OPL music would potentially fix a lot of these complaints.

rakohus:
EE uses ADLMidi. According to CSG, this is the best option for doom engines right now. Lib is here: https://github.com/Wohlstand/libADLMIDI

bcahue:
In stable now.

PR #1105, opened 2025-01-08
Implement OPL MIDI playback using libADLMIDI

jpernst:
This adds a libADLMIDI music system option (Implements #306)

The following CVARS control the driver:
`snd_oplcore`: Quality of the OPL emulation core to use, options are 0 (Dosbox), 1 (Nuked 1.74), 2 (Nuked 1.8), Default 0.
`snd_oplpan`: Enable smooth stereo panning. Off/On. Default On.
`snd_oplchips`: Number of OPL chips to emulate, from 1-8. Default 6
`snd_oplbank`: Which instrument set to use, options are 0 (Doom), 1 (Doom II), and 2 (DMXOPL3). Default 1.

Corresponding menu items are added to the sound menu with more user-friendly explanations.
Additionally, since this driver relies on SDL_Mixer to output the audio (since libADLMIDI does not do that on its own), non-MIDI music formats are delegated to SDL_Mixer, allowing this driver to support all music types. For this reason I think this driver is a strong candidate for being the default option, as it requires no external configuration and will produce working music in all cases, but I'll leave that for the project to decide.


rakohus:
This is great! Have been testing it. Definitely performance issues when it's maxed out but it works super well and sounds awesome. Some observations and thoughts below:

1) If you go from Portmidi -> LibADLMidi, the volume seems correct. However if you go from SDL_Mixer -> LibADLMidi, the fm synth is much much quieter. Not sure what's up with that, but the phenomenon carries between levels so it won't correct until the driver is reinitialized. This is on Windows 10.

2) Quality of life: the name of the backend `LIBADLMIDI` should probably say something like `LIBADLMIDI (FM SYNTH)` or something similar. The average user won't know what LIBADLMIDI means.

3) Starting to think we might want to nest the FM Synth Options in another sub-menu right there after we pick some good defaults. The current defaults are fine, but DMXOPL3 sounds so much better than Doom(2) options that I would personally advocate for that to be the default. I welcome other thoughts here on menu / defaults. **EDIT**: Actually after playing with more maps, the Doom 2 instrument set is probably the best default. DMXOPL3 is really sweet though.

rakohus:
image

Actually had some additional thoughts that might make this feature more visible and make more sense in the context of the menu. 

4) Since libADLMIDI already relies on sdl_mixer as mp3/ogg/mod/etc do, it probably should not be a driver option via `MUSIC SYSTEM BACKEND` option. Instead, create a new menu option underneath that simply says `FM SYNTH MIDI` that is 0/1. If it's set to 1, it does as the other formats do and just playback via SDL_Mixer regardless of the `MUSIC SYSTEM BACKEND` selection. I believe that would be more user friendly and probably get it used more

jpernst:
I changed the name of the option to make it more clear that it's FM synth, and also attempted a possible fix for the volume issue. I'm not on windows and can't reproduce it myself, but hopefully it works.

electricbrass:
The volume issue still occurs for me, also on Windows 10.

jpernst:
Made a few minor cleanups, but it's unlikely to fix the volume issue. I suspect that has something to do with SDL_Mixer's windows implementation, so I can't test or debug it. If someone on windows can identify the problem or offer a workaround, I'll add it.

jpernst:
I've made one more attempt at a fix based on comments from the original sdl_mixer driver. The windows midi mapper volume can interfere with normal pcm volume on modern versions of windows, so I now reset it when this driver is loaded. I can't think of any other reason the volume would change.

electricbrass:
Debug builds appear to be failing for me on this PR.

image

electricbrass:
> 4\. Since libADLMIDI already relies on sdl_mixer as mp3/ogg/mod/etc do, it probably should not be a driver option via `MUSIC SYSTEM BACKEND` option. Instead, create a new menu option underneath that simply says `FM SYNTH MIDI` that is 0/1. If it's set to 1, it does as the other formats do and just playback via SDL_Mixer regardless of the `MUSIC SYSTEM BACKEND` selection. I believe that would be more user friendly and probably get it used more

I'm also noticing that currently `AdlMidiMusicSystem` is inheriting from `MusicSystem` (rather than `MidiMusicSystem`) and returns true from all the `IsXXXCapable()` functions despite (to my knowledge), not being capable of most of those.

jpernst:
Work on this has been picked up by #1196, so this PR should no longer be necessary; closing.

PR #1106, opened 2025-01-09
Implement basic scaling for console font

jpernst:
I have poor vision and the default 8x8 console font was giving me headaches trying to read it, so I took a quick pass to make it scalable. The `con_scaletext` CVAR (range 1-4) controls the factor. A menu option was also added in the messages submenu.


rakohus (approved):
Awesome, much needed for a long time. Thank you!

PR #1107, opened 2025-01-10
Add console auto scaling

electricbrass:
This changes `con_scaletext` CVAR so that when set to 0, the console is scaled automatically by the vertical resolution / 450.


bcahue (approved):
Looks good.

PR #1108, opened 2025-01-11
Rework maplist `lastmap`

electricbrass:
This is based on @loopfz's work in #899. `sv_mapliststayonwad` has been removed. Instead, it's behavior is triggered by the presence of the `lastmap` argument to the `wad`, `addmap`, and `insertmap` commands. This means that explicitly setting the lastmap is required when that behavior is desired. Additionally, `sv_loopepisode` has been removed, as it is incompatible with UMAPINFO's implementation of episodes, and this allows for more flexibility.

### DIFF
--- a/server/src/sv_level.cpp
+++ b/server/src/sv_level.cpp
@@ -204,9 +204,9 @@ std::string G_NextMap()
 {
 	std::string next = level.nextmap.c_str();
 
-	if (gamestate == GS_STARTUP || sv_gametype != GM_COOP || next.empty())
+	if (gamestate == GS_STARTUP || (sv_gametype != GM_COOP && forcedlastmap.empty()) || next.empty())
 	{
-		// if not coop, stay on same level
+		// if not coop, and lastmap is not specified, stay on same level
 		// [ML] 1/25/10: OR if next is empty
 		next = level.mapname.c_str();
 	}
--- a/server/src/sv_level.cpp
+++ b/server/src/sv_level.cpp
@@ -204,9 +204,10 @@ std::string G_NextMap()
 {
 	std::string next = level.nextmap.c_str();
 
-	if (gamestate == GS_STARTUP || next.empty())
+	if (gamestate == GS_STARTUP || sv_gametype != GM_COOP || next.empty())
 	{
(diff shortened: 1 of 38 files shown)


server/src/sv_level.cpp
```diff
@@ -177,23 +186,40 @@ BEGIN_COMMAND (wad) // denis - changes wads
...
+	return false;
+}
+
 // Returns the next map, assuming there is no maplist.
 std::string G_NextMap()
 {
 	std::string next = level.nextmap.c_str();
 
-	if (gamestate == GS_STARTUP || sv_gametype != GM_COOP || next.empty())
+	if (gamestate == GS_STARTUP || next.empty())
```
electricbrass: Not sure yet why this was changed in the original pr

server/src/sv_level.cpp
```diff
@@ -247,7 +271,7 @@ void G_ChangeMap()
 		else
 		{
 			size_t next_index;
-			if (!Maplist::instance().get_next_index(next_index))
+			if ((!forcedlastmap.empty() && !isLastMap()) || !Maplist::instance().get_next_index(next_index))
```
electricbrass, line 262: need to check the logic of `isLastMap`

server/src/sv_level.cpp
```diff
@@ -204,19 +230,17 @@ std::string G_NextMap()
 
 	// NES - exiting a Doom 1 episode moves to the next episode,
 	// rather than always going back to E1M1
-	if (iequals(next.substr(0, 7), "EndGame") ||
-	    (gamemode == retail_chex && iequals(level.nextmap.c_str(), "E1M6")))
+	if (level.nextmap == "" || level.mapname == forcedlastmap ||
+			iequals(next.substr(0, 7), "EndGame") ||
+			(gamemode == retail_chex && iequals(level.nextmap.c_str(), "E1M6")))
```
electricbrass, line 224: need to check the logic of this if condition and the following one

bcahue (dismissed):
Looks good, another message is being added so it will break protocol but if this is for 11 that is moot.

PR #1115, opened 2025-01-12
[BUGFIX] Flipped sprites during cast call being drawn offset and stretch

electricbrass:
Now they're centered again and not stretched horizontally.


bcahue (approved):
Looks good, tried it and it works great!

Issue #1119, opened 2025-01-12
[BUG] X offsets applied incorrectly to horizontally scaled textures


electricbrass:
**Describe the bug**
Offsets appear to be being applied *before* scaling, when it should be the other way around so that offsets are still in map units, rather than texels.

![image](https://github.com/user-attachments/assets/7d519656-c0e7-49e3-a35a-43dc6bffff59)
This is 4 linedefs with offsets set so that they should align to display the full texture, but we see the sections do not align properly.

Interestingly, if this section of `R_PrepWall` is modified from
```
fixed_t colfrac = segoffs + FLOAT2FIXED(uinvz / curscale);
texoffs[i] = colfrac;

if (toptexture)
{
	int colnum = R_TexScaleX(colfrac, toptexture) >> FRACBITS;
	topposts[i] = R_GetTextureColumn(toptexture, colnum);
}
if (midtexture)
{
	int colnum = R_TexScaleX(colfrac, midtexture) >> FRACBITS;
	midposts[i] = R_GetTextureColumn(midtexture, colnum);
}
if (bottomtexture)
{
	int colnum = R_TexScaleX(colfrac, bottomtexture) >> FRACBITS;
	bottomposts[i] = R_GetTextureColumn(bottomtexture, colnum);
}
```
to
```
fixed_t colfrac = FLOAT2FIXED(uinvz / curscale);
texoffs[i] = segoffs + colfrac;

if (toptexture)
{
	int colnum = (segoffs + R_TexScaleX(colfrac, toptexture)) >> FRACBITS;
	topposts[i] = R_GetTextureColumn(toptexture, colnum);
}
if (midtexture)
{
	int colnum = (segoffs + R_TexScaleX(colfrac, midtexture)) >> FRACBITS;
	midposts[i] = R_GetTextureColumn(midtexture, colnum);
}
if (bottomtexture)
{
	int colnum = (segoffs + R_TexScaleX(colfrac, bottomtexture)) >> FRACBITS;
	bottomposts[i] = R_GetTextureColumn(bottomtexture, colnum);
}
```
The scaled textures will render proper (at least on solid walls), except that they then exhibit behavior similar to #1085.

Here is how the wall should look:
![image](https://github.com/user-attachments/assets/75c4c57a-c015-4a26-a96c-5ae6a4970e81)

electricbrass:
I think this might also be an issue for vertical offsets on solid walls, but I can confirm it has been fixed for 2 sided midtextures.

Issue #439, opened 2021-07-29
[BUG] Scaled textures in TEXTURE1/2 lumps doesn't have their Y scale applied


Ch0wW:
**Describe the bug**
When playing a wad that has TEXTURE1/2 lump with scaled Y/X textures, we can observe that the texture is properly scaled on the X axis. However, the Y axis isn't scaled at all, potentially messing up the design of all levels that use it.

**Build that the bug occurred in**
Any release of Odamex, up to the latest development version.

**To Reproduce**
Simply play a wad that have Y-scaled textures defined in a TEXTURE1/2 lump.

**Expected behavior**
Textures should be properly scaled.

**Screenshots, NetDemos, & Crash Dumps**
![image](https://user-images.githubusercontent.com/2938006/127559184-44a1dabd-7361-4bfc-a6d6-a9182e17f5c2.png)

**Additional context**
I'm unsure if ZDoom 1.22 even had that issue, and if that was fixed in ZDoom 1.23b33.

Ch0wW:
Proper behaviour should look like this:
![image](https://user-images.githubusercontent.com/2938006/127559500-7e355330-8113-473d-8585-c7eb542f071a.png)

Ru5tK1ng:
After digging around, I believe this was fixed in 1.23b29. The following was in the changelog:

- Made some changes to the renderer to support drawing textures with different
  scales on the same wall and to also support obtaining textures scaling from
  some of the previously unused bytes in the maptexture_t structs stored in
  the wad.

bcahue:
In r_segs.cpp:

![image](https://github.com/user-attachments/assets/285e185b-905b-48e6-93db-67db951bfac6)

Seems we need to take y-scaling into account for this in order for this to work. It seems to propagate the scaling information correctly, just isn't accounting for y scale when rendering.

Issue #1085, opened 2024-11-28
[BUG] Midtex scrolling when it touches left edge of screen


electricbrass:
**Describe the bug**
In comatose.wad, some fence midtextures will scroll when you turn so that the left edge of the screen touches them. They move with the camera as long as they continue to touch the left edge.

**Build that the bug occurred in**
Any version after [20795f0](https://github.com/odamex/odamex/commit/20795f0477b2120d2abc0d7875a4f91d3dab6609)

**To Reproduce**
Load comatose.wad and walk up to the fences near the start of the map.

**Expected behavior**
The fence should not move when it touches the screen edge.

**Screenshots, NetDemos, & Crash Dumps**
[Video](https://cdn.discordapp.com/attachments/250157428455243777/1294893102171164693/Odamex_10.6.0_deepbsp-support_gcc4c8-9258_-_DOOM_2__Hell_on_Earth_2024-10-13_00-21-25.mp4?ex=6749f86d&is=6748a6ed&hm=a1e5e5dcff5f08dda052b83008fe22d0c414b979faa45787f604730b32753864&)

rakohus:
So this issue came in with the deepbsp support?

electricbrass:
I suspect it existed before then, but I don't know of any other maps where this happens to be able to check earlier versions.

electricbrass:
I believe this is related to when the seg offset is applied when calculating the horizontal texture offset.

electricbrass:
Changing this line so that `colfrac` equals `FLOAT2FIXED(uinvz / curscale) - segoffs` makes these lines render correctly. However, it does break every line *except* these.
https://github.com/electricbrass/odamex/blob/eabfd51625fb996431b319c5dc974457d5c95d85/client/src/r_segs.cpp#L762

PR #1120, opened 2025-01-12
[BUGFIX] Vertical scale of textures is not applied

electricbrass:
Addresses #439

Before:
image

After:
image

While implementing this, three related bugs were discovered. Firstly, when rendering scaled textures as 2 sided midtexes, and marked as lower unpegged, the texture height was being multiplied by the scale factor, instead of the inverse, resulting in the texture being drawn at an incorrect height. This has been fixed. Also for 2 sided midtexes, the vertical offset was being applied before the scaling, instead of after, resulting in the offset being in texels instead of map units. This is also fixed. The last bug is documented in #1119. Currently it is *not* fixed by this PR.

### DIFF
--- a/client/src/r_segs.cpp
+++ b/client/src/r_segs.cpp
@@ -638,7 +638,7 @@ void R_RenderMaskedSegRange(drawseg_t* ds, int x1, int x2)
 
 	// find texture positioning
 	if (curline->linedef->flags & ML_DONTPEGBOTTOM)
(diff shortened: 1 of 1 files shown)


electricbrass:
After spending a bit more time looking at the code I saw how to fix #1119.

electricbrass:
This now also includes a fix for #1085

bcahue (approved):
rw_bottomtexturemid doesn't seem to be set? I don't have an in-depth look into these changes but it passes the eye test.

electricbrass:
It's being set a few lines after rw_toptexturemid

PR #1080, opened 2024-11-24
Change allfearthesentinel.net to allfearthesentinel.com.

Acts19quiz:
Change all references to TSPG from static.allfearthesentinel.net/wads/ to [static.allfearthesentinel.com/wads/](https://static.allfearthesentinel.com/wads/). allfearthesentinel.net no longer resolves to an existing website, and queries for WADs hosted there will fail.

[TSPG changed domain name](https://zandronum.com/tracker/view.php?id=4099) over a year ago.

PR #1123, opened 2025-01-13
[BUGFIX] Fix update to cl_downloadsites when upgrading a config

electricbrass:
#1080 broke the functionality to update the value of `cl_downloadsites` when upgrading to a new Odamex version. This has been fixed, as well as properly added the ability for the TSPG url to be updated.


bcahue (approved):
Looks good

PR #1124, opened 2025-01-13
A few minor fixes and changes

electricbrass:
This PR is a collection of a few small things:
- Some more compiler warnings are squashed by using the `[[fallthrough]]` attribute everywhere I'm certain the fallthrough is intended, not comparing CVARs to bools, changing `SArray::iterator` and `OHashTable::iterator` to no longer inherit from the deprecated `std::iterator`, and fixing some of the simplest cases of implicit `size_t` to `int` casts.
- The `NORETURN` define has been replaced by the compiler-agnostic `[[noreturn]]` attribute that was introduced in C++11. 
- Some old comments containing outdated information have been removed or updated.
- Warnings and errors thrown by `OScanner` are now labelled `Parse Error` and `Parse Warning` instead of `Script Error` and `Script Warning`, for clarity.
- A temporary hack for compatibility between 10.0 and other 10.x versions has been removed.
- `override` is used wherever it is appropriate (except in the sdl files, there's so much there, I'm not touching that lmao).
- `StrFormat` has been removed now that we have access to `fmt::sprintf`
- A few `FIXME`s have been addressed

### DIFF
--- a/client/sdl/i_main.cpp
+++ b/client/sdl/i_main.cpp
@@ -49,7 +49,7 @@
 #include <stack>
 #include <iostream>
 
-#include "i_sdl.h" 
+#include "i_sdl.h"
 #include "i_crash.h"
 // [Russell] - Don't need SDLmain library
 #ifdef _WIN32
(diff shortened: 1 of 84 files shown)


bcahue (approved):
This looks good, not sure what legacyId was used for or if its accounted for now in 11, but a quick Horde test should show any potential issues.

electricbrass:
Horde wave IDs were changed between 10.0 and 10.1 so legacyId was the ID from 10.0. That was there to allow for compatibility between 10.0 and later 10.x versions. With the major version bump, that compatibility is no longer needed.

PR #1125, opened 2025-01-14
Use constexpr where possible 2: Electric Boogaloo

electricbrass:
This PR supersedes #658. It brings those changes up to date with the latest from protobreak and fixes a few issues here and there present in that PR.


bcahue (approved):
Looks good. From a quick test it seems working as well.

PR #1126, opened 2025-01-14
[BUGFIX] Crash when unimplented ZDoom line specials called in certain contexts

electricbrass:
`LS_NOTIMP` prints out the number of the unimplemented special, but this was being obtained from the line pointer passed to line special functions. If the action does not have an associated line, this pointer would be null and cause a crash. Now, the special is passed as an additional function argument in all locations where line specials functions are called.


electricbrass:
The additional argument has been removed and LS_NOTIMP no longer prints the special number.

bcahue (approved):
I had good intentions when I added that, but unfortunately didn't account for when LineSpecials isn't activated by a line. This is enough to save us a crash from no-op linedefs or script lines. Thanks!

PR #1128, opened 2025-01-15
[FEATURE] Add support for MAPINFO AirSupply property

electricbrass:
This allows specifying the amount of time, in seconds, before a player starts taking damage while underwater. It can also be set to 0 to allow players to be underwater indefinitely. The default value is 10, the same as ZDaemon and ZDoom 1.23b33. This differs from modern GZDoom and Zandronum which have a default value of 20, and also apply less damage when drowning.


bcahue (approved):
Looks good.

Issue #1054, opened 2024-09-24
[FEATURE] Support Skulltag Style Win/Lose Interpics & Music


rakohus:
**Is your feature request related to a problem? Please describe.**
Classic Skulltag had some built in lumps that would occur depending on the players' win/loss conditions. These could be useful in Odamex games, including survival/horde/etc. This would be a nice feature to further enhance Horde mode in particular.

**Describe the solution you'd like to see**
Support the following lumps:

**GRAPHICS**
- `winerpic`: Shown instead of interpic when a winning condition is met in game modes. This is the same lump name used by classic Skulltag, so it would expand support for some of those older wads.
- `loserpic`: Same as above, but for a losing condition.

If only one exists, just fall back to regular interpic.

**MUSIC**
- `d_stwin`: Play this music during intermission if winning condition met. This is the same lump name used by classic Skulltag, so it would expand support for some of those older wads.
- `d_stlose`: Same as above, but for a losing condition.
- `d_owin`: Same as d_stwin, but renamed for Odamex. If for some reason both d_stwin and d_owin are present, prioritize d_o*.
- `d_olose`: Same as above.

electricbrass:
> If only one exists, just fall back to regular interpic.

Do you mean that if, for example, there is a `winerpic`, but no `loserpic`, that `interpic` should always be used? Or that `interpic` should be used as the `loserpic`?

One of the few places I found information about this (Zandronum 1.1 changelog), it described the latter behavior.

electricbrass:
Also, what should be done in the case of a draw?

electricbrass:
And if a map has an enter/exitpic set, does that or the winer/loserpic take precedence?
Or do we add the ability to define per-map winer/loserpics in mapinfo?

electricbrass:
There's a submitted patch for zandronum adding new mapinfo properties
https://zandronum.com/tracker/view.php?id=4213&history=1

PR #1129, opened 2025-01-15
[FEATURE] Support Skulltag Style Win/Lose Interpics & Music

electricbrass:
Addresses  #1054

Adds support for 8 new lumps: `WINERPIC`, `LOSERPIC`, `D_STWIN`, `D_STLOSE`, `WINANIM`, `LOSEANIM`, `D_OWIN`, and `D_OLOSE`.

The first 4 are those used by Skulltag for custom intermission pics and music depending on whether the player viewing them won or lost. `D_OWIN` and `D_OLOSE` are the same as `D_STWIN` and `D_STLOSE`. `WINANIM` and `LOSEANIM` will be loaded as ID24 interlevel lumps, and take precedence over `WINERPIC` and `LOSERPIC`.

Also adds support for `InterMusic` in `MAPINFO`. This allows specifying custom music on intermission screens. Note this is not the same as `UMAPINFO`'s `intermusic`, which sets the music for text screens.

### DIFF
--- a/client/src/cl_level.cpp
+++ b/client/src/cl_level.cpp
@@ -412,6 +412,22 @@ void G_DoCompleted (void)
 		}
 	}
 
+	const WinInfo& win = levelstate.getWinInfo();
+	switch (win.type)
+	{
+		case WinInfo::WIN_EVERYBODY:
+			wminfo.winner = true;
+			break;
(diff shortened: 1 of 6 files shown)


bcahue (approved):
Looks good, thanks!

Issue #1127, opened 2025-01-15
[BUG] Crash when calling extended ACS functions


electricbrass:
**Describe the bug**
When calling one of the internal ACS functions from Eternity and GZDoom extended ACS, Odamex crashes.

**Build that the bug occurred in**
Various 11.0.0 and 10.6.0 dev builds, but it has likely existed for a very long time.

**To Reproduce**
Trying loading a map with a simple script such as:
```
#include "zcommon.acs"

script 1 OPEN
{
	GetAirSupply(1);
}
```

**Expected behavior**
Print an error message instead of crashing. Maybe even implement support for some of these functions.

**Additional context**
This is because Odamex assumes all ACS functions are also line specials, but the internal ACS functions have negative IDs, so it tries to access the LineSpecials array with a negative index.

electricbrass:
Actually, I was wrong about what was happening here. The issue I described would occur *if* execution ever actually reached `DLevelScript::ActivateLineSpecial()`, which it does not. In the case of the particular script I sent, here is what happens: First, a byte is pushed to the stack with `PCD_PUSHBYTE`. The next instruction is `CD_DIVGLOBALARRAY`. It tries to access `ACS_GlobalArrays[111]`, but there are only 64 global arrays, so it is treating some other memory as an `OHashTable`, leading to this infinite loop:
![image](https://github.com/user-attachments/assets/e9f20347-c4c0-445a-ac55-4e7d967dc8c5)
Or sometimes it just crashes like when I first noticed the problem.

electricbrass:
If the number of global arrays is increased, the next instruction in the script is to terminate the script. This definitely seems like the bytecode is not being interpreted correctly.

electricbrass:
Well, `240 is PCD_DIVGLOBALARRAY`, but 351 is `PCD_CALLFUNC`. Here's the compiled ACS. 240 and 111. 
![image](https://github.com/user-attachments/assets/7a240f4a-9833-4c1e-9e99-a9cac062fd52)
So I decided to disassemble it. I don't think this is quite correct, the MULTIPLY there is actually the function to call (16). If I change the function to another, such as 9, that number changes to match, so that's likely an issue with listACS. But this does confirm that this should be considering the next TWO bytes.
![image](https://github.com/user-attachments/assets/d7a46dd5-890d-42b4-af0e-4ee784edd0a0)
The 1 between 111 and 16 is the number of arguments the function takes.

electricbrass:
[...]

PR #1130, opened 2025-01-15
[BUGFIX] Incorrect ACS P-codes outside 0-239 range

electricbrass:
Addresses #1127

ACSe uses 2 bytes for instructions after the first 240. Odamex was reading only the first byte, resulting in executing the wrong instruction and undefined behavior.


bcahue (approved):
Looks good, tested with this wad
[scriptpcodetestwad.zip](https://github.com/user-attachments/files/18432835/scriptpcodetestwad.zip)

PR #1132, opened 2025-01-15
[FEATURE] Add search bar to PWAD select GUI

electricbrass:
image

I think it should all work but this code feels...fragile.

I've also bumped the FLTK version to a stable release instead of a prerelease version of 1.4.


bcahue (approved):
Looks good here

PR #1134, opened 2025-01-16
Minor menu fixes & improvements

electricbrass:
A typo in the player setup menu has been fixed: "Prefered" -> "Preferred"

The "Download from server" setting has been changed to say "Download from internet". The help string for `cl_serverdownload` has also been updated to reflect that wads are no longer downloaded from Odamex servers.

Messages display settings now always use On for show and Off for hide.
Before:
image
After:
image


rakohus:
LGTM

PR #1136, opened 2025-01-17
Convert SV_BroadcastPrintf and friends to use fmtlib templates

electricbrass:
No more varargs :)


bcahue (commented):
Since this changes how chat is done, we should test it before approving. Everything seems to be in order otherwise.

electricbrass:
I've spent some time testing with every gamemode in singleplayer and on a server, mixing and matching clients and servers with and without this change, and as far as I can tell, all messages are displayed properly.

PR #1142, opened 2025-01-18
Don't immediately exit the menu when changing player name

electricbrass:
This is an alternative to #886. It's still a bit of a hack, but that's difficult to avoid with the state of the menu. This solution at least gets closer to the root cause.


bcahue (approved):
Looks good.

Issue #1033, opened 2024-09-13
[CHANGE] Add subfolders for screenshots and downloaded wads


electricbrass:
**What is the existing behavior? Please describe.**
Screenshots, downloaded wads, and crash dumps currently all get placed together in a single folder.

**What is the desired behavior? Why are you requesting the change?**
Within the folder currently used (whatever that may be, depending on OS, install method, and user settings), place screenshots and downloaded wads into subfolders.

electricbrass:
I noticed that the placement of screenshots on Windows depends on whether it is a portable installation or not, while wad downloads do not. Also, the user directory is the last place in the wad downloads list, which means without `cl_waddownloaddir` being set, for users who have an IWAD folder, it will likely default to downloading there.

PR #1143, opened 2025-01-18
[FEATURE] Place downloaded wads, netdemos, and screenshots into subfolders

electricbrass:
Addresses #1033

To help improve organization of the Odamex directory, netdemos and screenshots are now placed into subfolders of the user directory. Downloads will be placed similarly if another location is not specified by `cl_waddownloaddir`.


bcahue (commented):
Requires testing, as it may have a chance to add subdirs to Program Files (steamapps directory) and I believe Odamex would fail in that instance

electricbrass:
It should be following essentially the same logic that is already being used to determine whether to write to My Documents or the Odamex binary directory, but I agree that it needs testing.

Issue #1109, opened 2025-01-11
[BUG] Background Fire Not Present in Player Setup Menu in Resolutions Higher Than 1920x1080


Hekksy:
**Describe the bug**
If your resolution is higher than 1920x1080, the "fire" graphic behind the player sprite is missing in the Player Setup menu. Tested with 1920x1440, 2560x1600, and 3840x2160

**Build that the bug occurred in**
Latest stable, head-merge-0-goe51-1 and 10.6.0 release

**To Reproduce**
Change to a resolution higher than 1080p and go to player setup to find the graphic missing

**Expected behavior**
The fire is behind the player sprite

**Screenshots, NetDemos, & Crash Dumps**
Not needed

**Additional context**

Hekksy:
Confirmed fixed at 2k and 4k resolutions in protobreak

PR #1144, opened 2025-01-19
[BUGFIX] Background Fire Not Present in Player Setup Menu in Resolutions Higher Than 1920x1080

electricbrass:
Addresses #1109

For resolutions up to 5K, the original `R_RenderFire` is still used. For higher resolution I've added an overload that uses `CleanXfac` directly instead of a template parameter, so that we don't run into this issue again in the future.


bcahue (approved):
Looks good.

PR #1145, opened 2025-01-19
[BUGFIX] Text screen backgrounds blank

electricbrass:
The improved GameInfo support unfortunately made the backgrounds of finale texts all black. This is now fixed.


bcahue (approved):
Looks good.

PR #1146, opened 2025-01-19
[BUGFIX] Crash after parsing GameInfo CreditPage field

electricbrass:
Fixed an out of bounds array access

also continuing my little cleanups of cpp files that only exist for a few variables


bcahue (approved):
Good find!

PR #630, opened 2022-03-21
Generalize gameinfo from hardcoded presets to serialized MAPINFOs

DeathEgg:
All data from gameinfo (except for maxSwitch, some flags, and titleString) has been serialized into MAPINFO files. This is done by adding support for the following MAPINFO tokens for gameinfo definitions. A lot of these weren't necessary for generalizing existing IWAD's, but instead for supporting Heretic and other IWAD's in the future.

* Border
* BorderFlat
* ChatSound
* CreditPage
* IntermissionCounter
* IntermissionMusic
* NoLoopFinaleMusic
* PageTime
* QuitSound
* FinaleFlat
* FinaleMusic
* InfoPage
* TelefogHeight
* TitleMusic
* TitlePage
* TitleTime
* DefKickback
* Endoom
* PauseSign
* GibFactor
* TextScreenX
* TextScreenY

I've also included the Heretic relevant MAPINFO's inside of odamex.wad. They aren't used yet, but they have everything Heretic needs for gameinfo filled out already.

PR #1147, opened 2025-01-20
[BUGFIX] Various issues introduced by #630 and #1135 

electricbrass:
Both of these PRs introduced a few bugs. This fixes all that I've found so far that haven't already been merged.

- Broken format strings for determining `MAPxx` map lump names
- Comparison of OLumpName to `'\0'` resulting in incorrect evaluation of if statement
- Broken format string for finding big font patches for the intermission screen
- Crash from an unhandled exception when trying to format a possibly NULL char*

Those listed so far are game breaking, completely preventing use of odasrv with MAPxx maps, immediate errors on certain intermissions, and a crash when controllers without names recognized by SDL are used (Thank you @rakohus for finding this last one).


bcahue (approved):
Looks good

PR #1153, opened 2025-01-26
Remove `BOOL` type

electricbrass:
Having 2 different boolean types makes the code less readable, especially when one of them is just an `int` and is even used as such in some places. Definitely doesn't help type safety either. This PR replaces all instances of `BOOL` with either `bool` or `int`.

This needs a lot of testing and a careful eye when reviewing the code, there are likely bugs introduced by some `BOOL` variables being changed to `bool`s when they really are being treated as `int`s.


electricbrass:
I've looked carefully over all the places changed, and have spent some time testing all gamemodes on various maps, and have encountered no issues so far.

PR #1155, opened 2025-01-26
[BUGFIX] FLTK error window when using certain command line arguments

electricbrass:
Currently, if any command line arguments are passed to Odamex, and none of them include any of those that skip the boot window (`-iwad`, `+connect`, `-connect`, `+map`, `-warp`, `+netplay`, `+playdemo`, `-file`, `-playdemo`, `-timedemo`), all arguments will be passed to FLTK, which will bring up the error window message below.
image
This PR introduces `-fltk`, which will pass its parameter as the command line arguments for the FLTK frontend.

Example -- start Odamex with the FLTK scheme set to Oxy:
`./odamex -fltk "-scheme oxy"`

### DIFF
--- a/client/gui/gui_boot.cpp
+++ b/client/gui/gui_boot.cpp
@@ -591,12 +591,20 @@ scannedWADs_t GUI_BootWindow()
 	// deforms the window.
 	Fl::keyboard_screen_scaling(0);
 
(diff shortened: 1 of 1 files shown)


bcahue (approved):
Looks good and solves the issue

PR #1162, opened 2025-01-28
[BUGFIX] Fix flying monsters getting stuck on players/each other

### LINKED ISSUE #698: [BUG] Odahorde flying enemies 'sticking' to players
**Describe the bug**
When running under a custom flying enemy exclusive to Odahorde, if that enemy were to lower its z axis it then clips into the player forcing them to become immobile until that enemy is killed.

10.1.0

**To Reproduce**
Playing odahorde with custom flying enemies

**Expected behavior**
Player to be tangible to enemies and unable to clip into sprites.

Demo attached

**Additional context**
[Odahorde mobs getting stuck in me.zip](https://github.com/odamex/odamex/files/9048913/Odahorde.mobs.getting.stuck.in.me.zip)

bcahue:
Resolves #698 and should also resolve flying monsters getting stuck on each other in Horde/high monster density maps.

Turns out there was a missing Z check on floating up/down. This caused floating monsters to have a chance to move within another actor, messing up the move functions and blocking any further moves.

(Marisa Heit comments preserved as this is taken from ZDoom 1.23)


bcahue:
[odaflyingmonstertest.zip](https://github.com/user-attachments/files/18568838/odaflyingmonstertest.zip)

Here's the test wad I used. Be sure to turn on:

co_realactorheight 1

According to the code co_zdoomphys1 has Z height implications but they don't seem to affect this fix.

The first caco demonstrates it won't move into you and block your movement, the 2nd one you have to activate both cacos at the same time. On 10.6 they get stuck fairly often, compared to this PR where they do not.

PR #1165, opened 2025-01-29
[BUGFIX] Crash when trying to free fire sky memory

electricbrass:
Fire skies of certain dimensions would corrupt the zone memory when updated, leading to a crash when deallocating that memory. This has been fixed.


bcahue (approved):
Looks good, nice catch!

PR #1166, opened 2025-01-29
[BUGFIX] Errors and crashes when displaying cast call if monster uses multiple sprites

electricbrass:
Currently, the cast call assumes that monsters only have a single sprite, but this is not necessarily true if they have been modified. This leads to out of bounds memory accesses, as the number of frames in a sprite may vary, which leads to crashes. This is fixed.


bcahue (approved):
Changing the BOOL type scares me but after a quick playthru, I don't see any adverse effects. Approved

PR #1178, opened 2025-02-14
Rewrite `m_fileio[_win32/posix].cpp` using `std::filesystem`

electricbrass:
All of the fileio code that doesn't deal with os-specific directories is now unified using C++17's `<filesystem>` library.


electricbrass:
Apparently the versions of GCC used by Ubuntu 18.04 and Fedora 28 don't fully support the filesystem library.

PR #1180, opened 2025-02-17
Bump SDL Mixer to 2.6.2

electricbrass:
Seems about time to move on to the next version on Windows. Most Mac and Linux users are already getting newer versions, and some important bugs are fixed.

Fixes 32in24-18 map11 music and other tracker music with position jumps.


electricbrass:
Turns out the Windows release of 2.6.3 was accidentally built without mod playback support. So we need to use either an older or a newer version. For now I have changed this PR to 2.6.2, but 2.8.1 seems like a good choice as well, but it is very recent.

bcahue (approved):

Issue #1183, opened 2025-02-22
[BUG] Eviternity Arachnotron Sounds


Hekksy:
**Describe the bug**
Arachnotrons from Eviternity 1 make no sound when firing their weapon.

**Build that the bug occurred in**
Odamex 10.6, also checked in dsdhacked and 11. Same results.

**To Reproduce**
Find an arachnotron. Warping to map26 and using noclip to the right of spawn works (there are two). You'll notice no sounds play as they fire their weapon.

**Expected behavior**
Play the new firing sound

**Additional context**
Found by a user in the Eviternity Discord.

Hekksy:
I have tested electricbrass' fix and can confirm it fixes the issue.

PR #1184, opened 2025-02-22
[BUGFIX] Eviternity arachnotron firing sound is silent

electricbrass:
Addresses #1183

`S_StartSound` was considering sfx_id 0 to be an invalid sound, but this is actually the id of `misc/unused`, which *is* used by dehacked patches. Now this sound is played properly.


bcahue (approved):
Looks good.

Issue #1160, opened 2025-01-27
[BUG] FOV not saved by client and singleplayer still requires sv_allowfov for FOV changes


electricbrass:
**Describe the bug**
There are 2 related issues here. One is that if a player changes their FOV, this will be reset upon wad switch or restarting the client. The other is that in singleplayer, sv_allowfov must still be enabled to change the FOV.

**Build that the bug occurred in**
Tested in 11.0.0-prerelease.9766 but definitely present for much longer. Likely as long as the `fov` command and `sv_allowfov` cvars have existed.

**Expected behavior**
A players set FOV should be saved, and always used in singleplayer or on a server with sv_allowfov enabled.

rakohus:
The second part of this is corrected in 11.1.

PR #1185, opened 2025-02-22
Allow changing fov in singleplayer without first setting sv_allowfov

electricbrass:
Partially addresses #1160


bcahue (approved):
Looks good.

PR #1186, opened 2025-02-23
Add unit tests to CI using Google Test

electricbrass:
This PR adds a new executable that can be built: `odagtest`. Currently it has some tests for `OHashTable`, `OLumpName`, `szp`, and fixed point math. The main Linux, Mac, and Windows ci workflows all run the tests and will fail if any tests fail.

~~At the moment, it is not linked against `odamex-common` because of the large amount of stubs this requires to not get linkage errors, but eventually this is probably the direction to go, because trying to add individual cpp files from common and figuring out all their dependencies makes for a very bad workflow and a mess CMakeLists.txt.~~

Some solution for exposing some client and server specific areas of code to the tests would be good to have as well.

The majority of changes to other files are replacing `extern` variables with `inline` to reduce the amount of duplication required in `t_stubs.cpp`.


electricbrass:
These tests are absolutely not comprehensive, but the main purpose of this PR is to have everything in place for adding more unit tests in the future.

electricbrass:
This also includes a fix for an off by one error when converting from `fixed_t` to `int` that was introduced with the 32 bit renderer.

fuzztooth:
This one is going to require some conflict resolution.

PR #1191, opened 2025-03-02
Make printf functions more uniform

electricbrass:
Currently, some of Odamex's printf-style functions use the `fmt::format` syntax and support custom formatters, while others use the `sprintf` syntax and do not. This PR updates almost all functions to the former, though the older `Printf` are unchanged is there are so many uses of it. The `PrintFmt` alternative introduced by Lexi is still present though and provides extra functionality such as formatting of `OLumpName` and `OString`.


bcahue (approved):
Looks to be correct

PR #1193, opened 2025-03-02
Dockerfile and docker-compose file for odasrv

cmbernard333:
## Summary

Added a dockerfile for running any version of odasrv based on branch or tag. 

## Implementation

Added new files under `docker` to allow a user to run any version of the odasrv binary by building a docker image. It does the following:

1. Pulls the specified branch (by default `stable`)
2. Setups up the build environment and compiles the specified branch
3. Creates a run environment for the newly created binary

### Compilation

When building a docker image for a specific branch, you can provide it a branch or tag via the `VERSION` file or via the `REPO_TAG` argument to the make process. For example:

```
$ cd docker
$ make REPO_TAG=protobreak
```

After the image is built, it is available under the name `docker.io/odamex-team/odamex-server:testing`. 

If a `TAG` variable is not specified, the version of the newly created docker image will always be `testing`.

### Patches

If manual patches need to be applied to the code base during compilation, they can be added to the `docker-files/patches` directory.

## Running a server via Docker Compose

A `docker-compose.yml` file is provided under `examples` with default configuration for running `odasrv`. Additionally, an example `.env` file is provided as `.env.example` (will need to be renamed to `.env`). The only variable required to run the docker compose file is `DOOM_BASE` which should specify a directory pointing to two other directories `IWADS` and `WADS` which contain the IWADS and PWADS for running the server.


docker/examples/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+        volumes:
+            - doomIWADS:/wads
+            - doomWADS:/other_wads
+            - ./configs:/configs:ro
+            - odahome:/home/doomguy/.odamex
+        command: >
+            -iwad /wads/DOOM2.WAD
+            -waddir /other_wads
+            +exec /configs/config.ini
+            -skill 4
```
electricbrass: `-skill` here is unnecessary, it should just be set in the config.

docker/examples/.env
```diff
@@ -0,0 +1,3 @@
+# replace with path containing /IWAD and /WADS
+DOOM_BASE=~/Documents/GZDoom
```
electricbrass: I'd probably make the example `DOOM_BASE` be something like `~/wads` or `~/Doom/wads`.

electricbrass:
A few other notes:
- Odamex configs aren't `ini` files, and the most common extension used for them is `cfg`, so I would use that instead of ini here.
- It would be useful to be able to set rcon and join passwords with `.env`.

And this happens when I try and actually use this:
image

docker/Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+
+# Clone the repository
+RUN true \
+    && test -n "$REPO_URL" && test -n "$REPO_TAG" \
+    && git clone --depth 1 --branch "$REPO_TAG" "$REPO_URL" odamex
+
+WORKDIR /build/odamex
+
+# Apply Manual Patches (make sure they are UTF-8 encoded)
+COPY docker-files/patches /patches
```
electricbrass, line 35: This fails if the patches directory does not exist.

docker/examples/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+            -heapsize 96
```
electricbrass: `-heapsize` doesn't actually do anything anymore.

cmbernard333:
> A few other notes:
> 
>     * Odamex configs aren't `ini` files, and the most common extension used for them is `cfg`, so I would use that instead of ini here.
> 
>     * It would be useful to be able to set rcon and join passwords with `.env`.
> 
> 
> And this happens when I try and actually use this: image

Aren't join password and rcon password set via config?

electricbrass:
> Aren't join password and rcon password set via config?

Yes, but someone might use this for a server where they have the config publicly visible. I suppose just adding an exec line to the config that references another file would work too though. So now that I think about it I'd leave that as is probably.

cmbernard333:
> A few other notes:
> 
>     * Odamex configs aren't `ini` files, and the most common extension used for them is `cfg`, so I would use that instead of ini here.
> 
>     * It would be useful to be able to set rcon and join passwords with `.env`.
> 
> 
> And this happens when I try and actually use this: image

So I built and ran this on my mac (as opposed to Linux where I wrote it) and I can't reproduce this error. How did you run it?

electricbrass:
> So I built and ran this on my mac (as opposed to Linux where I wrote it) and I can't reproduce this error. How did you run it?

I ran it on Ubuntu 24.04 under WSL2 on Windows 10. I just checkout out this branch and ran `make` in the `docker` directory and then ran `run.sh`.

electricbrass:
I'm trying it again but now I'm getting this:
image

docker/Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+    && apt-get install -y -qq --no-install-recommends \
+        ca-certificates \
+        git \
+        gcc g++ \
+        cmake \
+        make \
+        deutex \
+        libssl-dev \
+        libsdl2-dev libsdl2-mixer-dev \
+        libpng-dev libpng++-dev \
```
electricbrass: odasrv doesn't need a lot of these. SDL, SDL_mixer, libpng, and cURL are only used by the client.

docker/Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+#     && /usr/games/deutex -rgb 0 255 255 -doom2 bootstrap -build wadinfo.txt ../odamex.wad
+
+# Build Odamex
+WORKDIR /build/odamex/build
+RUN true \
+    && cmake -W no-dev \
+        -D CMAKE_BUILD_TYPE=Release \
+        -D CMAKE_CXX_COMPILER=g++ \
+        -D CMAKE_C_COMPILER=gcc \
+        .. \
```
electricbrass, line 64: If you add `-D BUILD_CLIENT=0 -D BUILD_LAUNCHER=0`, it will skip building some of their dependencies that are unneeded for the server.

electricbrass:
It seems to work for me with these changes:

```
diff --git a/docker/Dockerfile b/docker/Dockerfile
index 954dc6979..842fcb276 100644
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,9 +13,7 @@ RUN true \
         make \
         deutex \
         zlib1g-dev \
-        libgl1-mesa-dev \
-        libssl-dev \
-        libcurl4-openssl-dev libcurlpp-dev \
+        libzstd1 libzstd-dev \
         patch \
         > /dev/null

@@ -88,11 +86,6 @@ RUN true \
     && apt-get update -qq \
     && apt-get install -qq --no-install-recommends \
         tini \
-        libssl1.1 \
-        libsdl2-2.0-0 \
-        libsdl2-mixer-2.0-0 \
-        libpng16-16 \
-        libcurl4 \
         gosu \
         > /dev/null \
     && rm -rf /var/lib/apt/lists/*
```

docker/Dockerfile
```diff
@@ -0,0 +1,107 @@
...
+RUN mkdir /patches
+COPY docker-files/patches /patches
+RUN true \
+    && shopt -s nullglob \
+    && for p in /patches/*.patch; do patch -p1 < $p; done
+
+# Update submodules
+RUN true \
+    && git submodule init \
+    && git submodule update
```
electricbrass, line 45: This should be submodule update -r
I think that's whats causing the issue with zlib

electricbrass: This should probably be submodule update -r

cmbernard333: Done

cmbernard333:
> It seems to work for me with these changes:
> 
> ```
> diff --git a/docker/Dockerfile b/docker/Dockerfile
> index 954dc6979..842fcb276 100644
> --- a/docker/Dockerfile
> +++ b/docker/Dockerfile
> @@ -13,9 +13,7 @@ RUN true \
>          make \
>          deutex \
>          zlib1g-dev \
> -        libgl1-mesa-dev \
> -        libssl-dev \
> -        libcurl4-openssl-dev libcurlpp-dev \
> +        libzstd1 libzstd-dev \
>          patch \
>          > /dev/null
> 
> @@ -88,11 +86,6 @@ RUN true \
>      && apt-get update -qq \
>      && apt-get install -qq --no-install-recommends \
>          tini \
> -        libssl1.1 \
> -        libsdl2-2.0-0 \
> -        libsdl2-mixer-2.0-0 \
> -        libpng16-16 \
> -        libcurl4 \
>          gosu \
>          > /dev/null \
>      && rm -rf /var/lib/apt/lists/*
> ```

Done

bcahue (approved):

PR #1196, opened 2025-03-03
Fix linker errors with libADLMidi in Windows debug builds and rework sound menu

electricbrass:
This relies on  #1105 and fixes a few issues it had.

First, debug builds on Windows now actually work instead of failing while linking.

The sound menu, which was getting crowded, now has a couple submenus. All LibADLMidi options have been moved to the "OPL FM Synth Options" menu, and the PortMidi "MIDI reset" option has been moved to the "Advanced MIDI Options" menu, which also contains some other options that were not previously exposed in the menu.

Main sound options menu:
image

OPL FM menu:
image

Advanced MIDI menu:
image


bcahue (approved):
Looks good

PR #1202, opened 2025-03-10
[FEATURE] Extra life and Resurrect Teammate powerups for Horde

bcahue:
The following 2 cvars are introduced:

- `g_horde_extralife` (defaults to 0.0, limit is 4.0)
  - This cvar will add to the powerup recipe for waves with the chance listed in the cvar, a powerup that grants an extra life for the player that collects it.

- `g_horde_resurrect` (defaults to 0.0, limit is 4.0)
  - This cvar will add to the powerup recipe for waves with the chance listed in the cvar, a powerup that will resurrect a random teammate who is joined into the round but dead and out of lives.

These cvars will have no effect in a Horde game with no `g_lives` set.

I also added 2 sprites for this. They kinda suck but better than nothing:

Odamex_SOLO_20250309_230017

We should try to get better ones.


common/p_horde.cpp
```diff
@@ -714,7 +716,27 @@ void HordeState::tick()
 		const int offset = P_RandomInt(16) + 30;
 		m_nextPowerup = ::level.time + (offset * TICRATE);
 
-		const mobjtype_t pw = define.randomPowerup().mobj;
+		hordeDefine_t newdef = define;
```
electricbrass: If the define has no powerups in the HORDEDEF, the new powerups won't spawn either because this whole section is skipped. It also seems like this could be done without copying the define on every tick. They're unlikely to be very large most of the time but it is copying 4 vectors. Could probably just do it when the define is being selected, since `g_lives` is latched.

common/p_interaction.cpp
```diff
@@ -577,6 +577,85 @@ ItemEquipVal P_GivePower(player_t *player, int /*powertype_t*/ power)
...
+	auto it = std::find_if(players.begin(), players.end(),
+	                       [&](player_s& p) { return p.id == playerid; });
+
+	if (it == players.end())
+		return;
+
+	it->lives += 1;
+	it->playerstate = PST_REBORN;
+
+	SV_BroadcastPrintf("%s has brought %s back into the fight!\n",
```
electricbrass: Now that all the `Printf` functions are implemented with `{fmt}`, you don't need to call `c_str()`. It will work with a `std::string` (and it does take them by `const&` so there's no extra copies happening).

common/p_interaction.cpp
```diff
@@ -577,6 +577,85 @@ ItemEquipVal P_GivePower(player_t *player, int /*powertype_t*/ power)
...
+	if (!G_IsLivesGame())
+		return;
+
+	if (!player)
+		return;
+
+	// Grab all the players in game, make a list of their ids, then pick one at random
+	PlayersView ingame = PlayerQuery().notHasLives().execute().players;
+	std::vector<int> ingameplayers;
+	for (const auto& p : ingame)
```
electricbrass, line 596: If only constructing this vector to check it's size, why not just do `int ingameplayers = std::count_if(ingame.begin(), ingame.end(), [&](const player_t* p){ return p->id != player.id })`

common/p_interaction.cpp
```diff
@@ -577,6 +577,85 @@ ItemEquipVal P_GivePower(player_t *player, int /*powertype_t*/ power)
...
+	if (ingameplayers.empty())
+	{
+		SV_BroadcastPrintf("%s tried to resurrect someone, but everyone is alive!\n",
+		                   player->userinfo.netname.c_str());
+		return;
+	}
+
+	int playerid = M_RandomInt(ingameplayers.size());
+
+	auto it = std::find_if(players.begin(), players.end(),
```
electricbrass: There's already `idplayer`, for finding a player by id. And then just use `validplayer` for the check below this.

electricbrass (changes requested):

common/p_interaction.cpp
bcahue: It does more than check its size, its the list of eligible ids where the id is randomly selected.

electricbrass: Am I reading something wrong? It looks like only the size is being used.

bcahue: I forgot to add where it actually selects from the array, whoops. Its coming soon

common/p_horde.cpp
bcahue: I fixed that particular bug, but the reason I don't want to modify the defines directly is that it seems the server only loads the defines once, and by injecting them whenever needed, it saves us from having to track and add/remove the powerups depending on cvar changes, and also users might want to put these powerups in their defines and add a level of complexity there.

electricbrass: I agree, which is why I said at the time the wave is chosen, not when the defines are loaded.

bcahue: The define has to be modified or horde needs to be restructured to allow temp defines, either doesn't seem very attractive for such a small feature. Also, this runs only on the tic that powerups are spawned in, so there's not much worry about it running every tic

electricbrass: Maybe I need to take another look at the larger structure of how this all works. It's been a bit since I did. I thought it would mostly just be taking this and moving it to another spot.
If it works it works though. The edge case of empty powerup defines was the most important to fix.

electricbrass: Okay yea I see now why that wouldn't work without modifying the original define.

PR #1213, opened 2025-03-27
Remove all calls to `strcpy` and `strcat`

electricbrass:
These functions are deprecated by Apple and trigger warnings on a few different compilers. `M_StringCopy`, an implementation of `strlcopy`, has been taken from Chocolate Doom, and is used in most places as a safer alternative to `strcpy`.


bcahue (approved):
Bunch of changes here that make me nervous but our hands are tied with the deprecation looming. Plus we'll have a testing phase before 11.1. Approved.

PR #1214, opened 2025-03-28
[BUGFIX] Bottom of Doom episode 1 sky renders incorrectly.

electricbrass:
#1138 introduced an issue when rendering the episode 1 sky. As seen below, the bottom 8 rows of the texture are uninitialized in memory. It turns out that the patch used by the sky has a y offset of -8. In vanilla, negative y offsets on patches are ignored, and offsets were also ignored in single patch textures. Odamex apparently supports negative patch offsets, and now that all textures go through the compositing step, the offset in SKY1 becomes visible. To fix this, I took the same approach as Eternity and GZDoom. When loading textures, the bad offset is set to 0 if the texture being loaded is this one.

image

### DIFF
--- a/common/r_data.cpp
+++ b/common/r_data.cpp
@@ -253,6 +253,19 @@ void R_DrawColumnInCache(const tallpost_t *post, byte *cache,
 	}
 }
 
+// Some vanilla textures use patch offsets that were ignored by the vanilla executable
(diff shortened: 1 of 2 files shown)


bcahue (approved):
Looks good.

PR #1220, opened 2025-04-02
Revert "[BUGFIX] Bottom of Doom episode 1 sky renders incorrectly."

bcahue:
Reverts odamex/odamex#1214


electricbrass:
#1222 should fix the issue caused by #1214.

PR #1221, opened 2025-04-02
Revert "Remove all calls to `strcpy` and `strcat`"

bcahue:
Reverts odamex/odamex#1213


electricbrass:
#1222 should fix the issues introduced by #1213

PR #1230, opened 2025-04-06
[FEATURE] Build and release a Flatpak file as part of the build pipeline

electricbrass:
This takes #1016 and brings it up to the latest stable, plus changes based on work done by @HyperEye.

With this, playback of midi and mod music from the flatpak works correctly now.

Users can specify their own timidity config to use with `flatpak override net.odamex.Odamex --env=TIMIDITY_CFG=<path to cfg>`. Note that this does not support soundfonts, only GUS patches.


electricbrass:
Todo list:
- ARM64 builds
- release builds
- allow users to set their own soundfonts

electricbrass:
I think this is finally ready. I've spent a lot of time making sure everything is in line with Flathub's requirements and guidelines so that the submission process goes smoothly after this is merged.

To get around Flatpak's limitation of only being able to run a single command with `flatpak run`, I've created a small shell script that is run instead, which allows the user to specify which of the 3 executables to run. This is especially useful for using odasrv.

image

bcahue (approved):
I don't have a linux machine to test on currently but this all seems well put together.

Issue #1204, opened 2025-03-11
[BUG] Line endings in release source tarball are different from the repo


cuantar:
**Describe the bug**
I've been maintaining a patch (PR 897; monster respawning time-domain tuning) using _e.g._ `git diff upstream/stable > monstersrespawn.patch`.  When attempting to apply my patch to the release tarball for Odamex 11.0.0 (_e.g._ like `patch -Np1 -i monstersrespawn.patch` from inside the toplevel dir in my clone of the upstream git repository), I encounter the following:
```
patching file common/c_cvarlist.cpp
Hunk #1 FAILED at 112 (different line endings).
1 out of 1 hunk FAILED -- saving rejects to file common/c_cvarlist.cpp.rej
patching file common/p_mobj.cpp
Hunk #1 FAILED at 80 (different line endings).
Hunk #2 FAILED at 777 (different line endings).
2 out of 2 hunks FAILED -- saving rejects to file common/p_mobj.cpp.rej
```

**Build that the bug occurred in**
Odamex-11.0.0 release source tarballs.  I reproduced the issue with https://github.com/odamex/odamex/releases/download/11.0.0/odamex-src-11.0.0.tar.xz, specifically.

**To Reproduce**
Download my patch from here:  https://patch-diff.githubusercontent.com/raw/odamex/odamex/pull/984.patch
(Probably will observe the same result with a different PR's patch...)

Download the release tarball in the link above.

Extract the tarball.

Change to the directory created.

Attempt to apply the patch (`patch -Np1 -i ...`) and observe the result.

**Expected behavior**
The patch should apply cleanly.

cuantar:
I hypothesize that the process responsible for creating the release tarballs is changing the line endings from Unix to DOS CRLF...

bcahue:
While we can add to the release job to convert line endings to LF, we should also investigate adding

```
text eol=lf
```

to `.gitattributes` to force LF line endings.

rakohus:
Fixed in 11.1. Release imminent

PR #1237, opened 2025-04-11
Add `* text=auto eol=lf` to gitattributes

electricbrass:
Addresses #1204

Git's autocrlf on Windows is causing the files in the .tar.gz and .tar.xz files to have CRLF line endings. At this point, support for LF on Windows is widespread for all the text file types in the Odamex repo, so I set the gitattributes to set them to LF on checkout.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+* text=auto eol=lf
 *.cpp ident
 *.h ident
 *.rc ident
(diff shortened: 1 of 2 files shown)


bcahue:
Was this tested for the Windows-Release pipeline?

electricbrass:
Yes
https://github.com/electricbrass/odamex/releases/tag/11.1.0

bcahue (approved):

PR #1241, opened 2025-04-15
Initial ID24 Music Changer support

elf-alchemist:


elf-alchemist:
Closed because I messed up the rebase between stable/protobreak on accident (whoops, ahaha), will fix and re-up soon enough

PR #1247, opened 2025-04-16
Update WinGet package as part of release workflow

electricbrass:
There is an existing WinGet package for Odamex, but it lags behind in updates by quite a bit. This PR is to make this package something officially maintained and updated as part of releases. When a new release is made, a PR is automatically made to https://github.com/microsoft/winget-pkgs with an updated manifest for the new version. Here is an example of such a PR: https://github.com/microsoft/winget-pkgs/pull/248963 (note that once https://github.com/microsoft/winget-pkgs/pull/243548 is merged, the new manifests created by this will be based on that one, rather than the 10.6.0 manifest that is currently the latest. This one properly indicates that Odamex is available for both x86 and x64, and allows installing per user or for all users)

For this to work, a fork of https://github.com/microsoft/winget-pkgs will need to be made under the odamex organization, and this repository will need a GitHub Personal Access Token (classic) with permissions set to `public_repos` as a secret with the name `WINGET_TOKEN`.
If there is an odamex service account, the access token should probably be from that, rather than someones personal account.


electricbrass:
I'm going to look into how GitHub Apps work, and see if we can use that as a solution for the PAT.

kris6673:
Until you get this working, I have an automation set up to run every 4 hours that checks if there is a new version available, that's not been added to the winget repo :)
I'll remove it when this gets merged.

electricbrass:
This will not work with a GitHub App, but I did just remember that @odamex-system exists. So that should work for what's needed here.

PR #1253, opened 2025-04-19
[BUGFIX] MBF21 ammo per shot

### LINKED ISSUE #1251: [BUG] MBF21 Ammo Per Shot Issues
**Describe the bug**
Currently the implementation of the MBF21 "Ammo per shot" dehacked field is closer to the ZDoom 1.23b33 "Ammo use" field. The primary differences are that the default ammo per shot value for the chainsaw and fist is 1, not 0, and that certain vanilla codepointers should have behavior dependent on whether or not ammo per shot has been set explicitly or not. For example, A_FireShotgun2 should consume 2 ammo on any weapon without ammo per shot set, even if the default for that weapon is not 2, but if ammo per shot is set, it should consume the ammo that has been set. This likely conflicts with ZDoom behavior, as ammo use does not have an equivalent check.

Everything from the last 10 years (it was introduced from Eternity and ZDoom a while before MBF21).

**Expected behavior**
Who even knows tbh, cross port inconsistencies are a nightmare

**Additional context**
GZDoom exhibits the same behavior as Odamex currently. It likely also didn't consider the edge cases from using its existing implementation for MBF21.

electricbrass:
Should resolve #1251
The behavior for ammo per shot now matches DSDA, Woof and Eternity as far as I can tell, while preserving ZDoom compatibility with ammo use.


electricbrass:
Actually this doesn't address the behavior with vanilla codepointers yet.

electricbrass:
After some more research and help from Xaser, here is the solution I have come up with. Odamex now fully implements MBF21 ammo behavior, and uses it by default, rather than the ZDoom 1.23b33 behavior it had previously. When either "Ammo use" or "Min ammo" is present in a DeHackEd lump, ammo switches to 1.23b33 compatibility.

cmbernard333:
👍

cmbernard333:
Looks good. Very useful for dsdhacked support.

cmbernard333 (approved):

PR #1256, opened 2025-04-30
[FEATURE] MUSINFO support

electricbrass:
What it says on the tin. Supports (to my knowledge) all variants of music changer things
- 14100 - change to map default 
- 14101-14163 - change to MUSINFO track 1-63
- 14164 - same as above but apparently unsupported by some ports
- 14165 - change to the MUSINFO track specified by arg0, 0 changes to map default. arg1 specifies order for mod music

Brings Odamex closer to supporting all the older features required for full ID24
Based on initial work by DeathEgg

Test map:
[musinfo.zip](https://github.com/user-attachments/files/21848817/musinfo.zip)


client/sdl/i_music.cpp
```diff
@@ -288,7 +288,7 @@ static MusicSystemType I_SelectMusicSystem(byte *data, size_t length)
 	return MS_SDLMIXER;
 }
 
-void I_PlaySong(const OByteSpan data, const bool loop)
+void I_PlaySong(const OByteSpan data, const bool loop, const int order)
```
cmbernard333, line 291: `const int` needed here?

client/sdl/i_music.h
```diff
@@ -68,7 +68,7 @@ void I_ResumeSong();
 //  plays a song, and when the song is done,
 //  starts playing it again in an endless loop.
 // Horrible thing to do, considering.
-void I_PlaySong(const OByteSpan data, const bool loop);
+void I_PlaySong(const OByteSpan data, const bool loop, const int order);
```
cmbernard333, line 71: `const int` needed here?

common/g_level.h
```diff
@@ -169,6 +170,9 @@ struct level_pwad_info_t
 	float			aircontrol;
 	int				airsupply;
 
+	// MUSINFO
+	std::unordered_map<int, std::string> musinfo_map;
```
cmbernard333, line 209: Is this level number to track name?

electricbrass: Level number? The map is a mapping of the musinfo indexes to music lump names. So the music changer thing's arg0 is used to retrieve the music lump from the map.

cmbernard333: Shows how much I know about MUSINFO. Thanks for the clarification!

client/sdl/i_music.h
electricbrass: Doesn't need to be const but it matches style of the other arguments (and anyways I think using const for non pointer/reference type arguments should be more common).

electricbrass:
Apparently SDL3_mixer removes `Mix_ModMusicJumpToOrder()`, which means the `order` arg for Hexen format music changers won't function when we move to SDL3 without some sort of workaround or moving away from SDL mixer.

Issue #1133, opened 2025-01-16
[BUG] Skies on walls briefly drawn incorrectly when changing screen sizes.


electricbrass:
**Describe the bug**
Skies drawn on upper walls will be drawn at the wrong height briefly when changing screen sizes.

**Build that the bug occurred in**
10.6.0 and 11.0.0-prerelease.9710

**To Reproduce**
Go to the large outdoor area in Eviternity MAP15 and look at the sky opposite the castle. Press + or -.

**Expected behavior**
No flashy!!

**Screenshots, NetDemos, & Crash Dumps**
[video](https://cdn.discordapp.com/attachments/250157428455243777/1329301871877165157/Odamex_11.0.0-prerelease.9710_-_DOOM_2__Hell_on_Earth_2025-01-15_22-05-31.mp4?ex=6789d857&is=678886d7&hm=ba0dff5ad6de30ab59ba229f4e6bf587c64f37c06dfbf1e1ff1af86b3d028df6&)

PR #1264, opened 2025-05-28
[BUGFIX] Sky transfers ignoring vertical offsets on first rendered frame

electricbrass:
Addresses #1133

No more flashy :)


electricbrass:
This also fixes an issue where MBF sky transfers change the vertical offset of all non-SKYDEFS skies.

fabiangreffrath:
Hm, maybe you still need this to fulfill this line of the spec?

`A sky defined by a transfer line or via the level defaults always uses the default sky behavior. This is defined as: ... Applying scrolling values as defined by any transfer lines associated with the rendered sky's sector.`

PR #1277, opened 2025-06-17
Allow building against the system FLTK 1.4 library

chewi:
`USE_INTERNAL_FLTK` defaults to true and ignores `USE_INTERNAL_LIBS` because users are unlikely to have it installed.

Tested against FLTK 1.4.3 on Gentoo Linux.

This is a fresh attempt at #591, which was rejected, but that was trying to make it work with 1.3 before 1.4 was released. 1.4 has been out a good while now, and I'm sure you'll agree that this is a _lot_ simpler.


electricbrass (approved):
Looks good to me 👍

PR #1285, opened 2025-06-24
[BUGFIX] Fix HOMs when trying to do sky hack with flatmapped skies

electricbrass:
Before:
image

After:
image


electricbrass:
They now also behave correctly when firing at them (no bullet puffs, rockets disappearing, etc)

bcahue (approved):

Issue #1282, opened 2025-06-23
[BUG] `exec` fails to open configs with paths relative to the user directory


electricbrass:
**Describe the bug**
`exec` allows absolute paths or paths within the write directory (`~\Documents\My Games\Odamex`, `~/.odamex`, or the installed location of the odamex/odasrv binary, depending on OS and installation method). Absolute paths work fine, but trying to `exec` a config in the write directory will fail with `Could not open "<filename>"`. This is not an error that the file could not be found, which instead reads `Could not find "<filename>"`.

**Build that the bug occurred in**
11.0.0 and onwards, haven't checked earlier yet.

**To Reproduce**
In the console, type `exec odamex.cfg`, which should reload the default config. It will fail.

**Expected behavior**
The file should be opened successfully and `exec`ed.

**Additional context**
There's maybe also supposed to be some support for relative paths? But relative to what? I don't know.

Would help somewhat with #1274

electricbrass:
I'm actually not sure whether the bug is it failing to open the files, or the fact that it's considering them found.

Issue #1274, opened 2025-06-11
[FEATURE] `cfgdirs` cvar that specifies directories for the `exec` command to search.


electricbrass:
**Is your feature request related to a problem? Please describe.**
I have Odamex and Odasrv cfgs that exec other cfgs. I have to specify the full path, which is slightly annoying, and also makes them less portable. If I move the directory they're contained in, or want to use them on a computer where things are organized differently, I have to remember to change every path. It's also annoying for private servers where I might be changing settings on the fly and want to exec a config but can't remember the path to where it is.

**Describe the solution you'd like to see**
A `cfgdirs` cvar that contains a list of directories that are searched when using the `exec` command. For example if I have the directory `/home/mia/Games/Doom/Odamex/odasrv_cfgs/`, containing `coop.cfg` and `horde.cfg`, and `cfgdirs` were set to point to that directory, I could do `exec coop.cfg` and `exec horde.cfg` instead of specifying the full path.

**Additional context**
I also think that the default value for this in odasrv should allow cfgs in the `config-samples` directory to be `exec`ed with a path starting from the samples directory, if installed to the standard location. e.g. `exec config-samples\coop-doom.cfg` instead of `exec "C:\Program Files\Odamex\config-samples\coop-doom.cfg"`

It appears that currently only absolute paths or paths within the write directory can be used with `exec`.

electricbrass:
if there is any way to concatenate strings in the console, I would be able to get most of what I want from this with a few aliases

PR #1288, opened 2025-06-25
[BUGFIX+FEATURE] Fix `exec` search paths + add `-cfgdir` argument

electricbrass:
Addresses #1274 and #1282

`exec` now actually loads the configs that it finds in a users write directory, rather than finding them and complaining that it can't open them.

It now also searches paths in the directory specified by `cfgdir`. This path is left empty for the client, but for the server it points to the directory containing the config samples, allowing users to do `exec config-samples/coop-doom.cfg`, for example. There are 2 problems with default values currently:
1. I don't know what that path would be on macOS.
2. The Windows path is the path for a system-wide install. Portable installs work fine because that path is already searched by default, but this isn't a sensible default for per-user installs. I'll probably use `M_GetBinaryDir`, but that will require changing the approach slightly, as it is not constexpr.


electricbrass:
After more testing, this does not fully work yet for my purposes, as using `+cfgdir` as a command line argument sets the value of `cfgdir` *after* the config specified with `-config` is `exec`ed.

electricbrass:
The `cfgdir` cvar has been removed, and replaced with a `-cfgdir` argument. This removes the benefit of easy access to the config-samples directory, as well as persistent storage of it on the client, but I believe the primary use for this would be to set a directory for odasrv to search, which makes this form the more useful in my opinion.

PR #1314, opened 2025-07-14
[BUGFIX] Fix use of equality instead of assignment in sndinfo

electricbrass:
what it says on the tin


ceski-1:
Oops, sorry about that. You'll want to update the server version of `s_sound.cpp` as well.

PR #1320, opened 2025-07-17
Cleanup/overhaul of DSDHacked support

electricbrass:
This is an attempt to fix many of the remaining bugs with the DSDHacked implemenation as well as clean up the code a bit.

It includes fixes for the following:
- Legacy of Rust lamps playing "misc/teamchat" when broken
- Weapon sprites in certain index ranges not rendering
- Modified vanilla things only using their dehacked modifications when not directly placed in the map
- A few bad format strings
- Avatars not being spawned on clients in multiplayer
- CTF things not being spawned on clients in multiplayer
- Missing GPL headers on new files
- SNDINFO ambient sound things not spawning

As well as the following features:
- Support for ID24 `USER_` and `DEHTHING_` strings (this allows using DSDHacked things in HORDEF!)
- Support for ID24's `bossactionednum` UMAPINFO key


common/doom_obj_container.h
```diff
@@ -1,315 +1,199 @@
...
-	return this->lookup_table.end();
-}
 
-template<class ObjType, class IdxType, class FreeFunction>
-bool DoomObjectContainer<ObjType, IdxType, FreeFunction>::contains(IdxType idx) const
-{
-	return this->find(idx) != this->end();
-}
+	// TODO: do we ever need this?
+	void append(const DoomObjectContainerType& other)
```
cmbernard333: Shouldn't need it unless the new `insert` doesn't support appending another doom object container. Backup is the only use case I can think of something like this.

common/doom_obj_container.h
```diff
@@ -1,315 +1,199 @@
...
-{
-	return this->container.capacity();
-}
-
-template <class ObjType, class IdxType, class FreeFunction>
-void DoomObjectContainer<ObjType, IdxType, FreeFunction>::clear()
-{
-	for (auto& obj : this->container)
+	using LookupTable = std::unordered_map<IdxType, ObjType*>;
+	// TODO: maybe use the vector for iteration instead of the map for speed?
```
cmbernard333: Depends on the type of iteration. The in order table has _everything_ that has been added via initialization and dehacked. The map should always contain the working set so iterating over that is usually the most accurate of what objects are used in game.

common/doom_obj_container.h
electricbrass: insert doesn't support that but even for backup, we don't really ever want to be appending. I think it's probably safe to just remove this entirely.

Issue #675, opened 2022-05-25
[FEATURE] Change CTF Flag Actor Heights for QoL


rakohus:
**Is your feature request related to a problem? Please describe.**
Quality of life.

**Describe the solution you'd like to see**
Change CTF flag dimensions: 42 height

**Additional context**
Should approximately match our sprites in-game, taking offsets into account.

rakohus:
Removing from 10.2.0 milestone. We can kick the can on this for now.

rakohus:
Changed to 48 units tall in #1321

PR #1321, opened 2025-07-17
Use GZDoom's torch heights instead of Eternity's

electricbrass:
Since GZDoom is much more popular, wads are more often tested with these heights, and may have issues with the Eternity heights. This can be seen in Eviternity II's map01, where the torches in the flying section fail to teleport in when `co_realactorheight` is enabled, while working just fine in GZDoom and Helion (which also uses the GZDoom heights).


electricbrass:
Flag heights have also been changed to 48, which brings them in line with Zandronum and better matches the sprites

Addresses #675

Issue #1081, opened 2024-11-24
[BUG] Eviternity II/DSDHacked: MAP18 has unclipped textures.


Acts19quiz:
**Describe the bug**
A region in MAP18 of Eviternity II has texture clipping not present in DSDA-Doom.

**Build that the bug occurred in**
11.0.0 (dsdhacked-pr-tests, g8835-1)

**To Reproduce**
1. Observe the attached region of MAP18 of Eviternity II. Since I can't described where to go, I've provided an automap overlay for the DSDA-Doom screenshot.

**Expected behavior**
No texture clipping.

**Screenshots, NetDemos, & Crash Dumps**
![Odamex_COOP_20241124_003028](https://github.com/user-attachments/assets/d06017a1-3db8-48bd-be31-20c0db828093)
Screenshot 1 is Odamex.
![Eviternity_II_MAP18_texture_clip](https://github.com/user-attachments/assets/d21701e3-3e8e-465c-8be2-493707a923df)
Screenshot 2 is DSDA-Doom.

**Other**
The DSDA-Doom screenshot is from its Software renderer for an equal comparison of behaviour.

bcahue:
I wonder if #1138 fixed this

electricbrass:
#1138 should mostly not affect lower textures, and this is already a composite texture, so it isn't affected at all by it. It's a different issue.

electricbrass:
Turns out that actually is middle texture, on a line just barely in front of the one behind it. The light level on both sides is also the same. I think this is actually just the traditional midtex clipping issue seen with floors.

electricbrass:
The sectors do have different specials though, which I believe in some ports will cause this behavior.

electricbrass:
Yep that is exactly it. In Odamex, the entire brick texture is showing through the floor. On woof and dsda, only the right half is. The difference is that the left half has different sector specials in front and behind the texture.
[texclipfloor.zip](https://github.com/user-attachments/files/21847069/texclipfloor.zip)

electricbrass:
Chocolate Doom matches Odamex, but neither Woof or DSDA preserves this behavior in lower complevels. DSDA even clips the right part partially.

electricbrass:
So this is actually a behavior very specific to DSDA and Woof, and not present in any other port's software renderer that I can find.

drivingspectrum:
[...]

PR #1357, opened 2025-08-19
Vertically clip masked midtextures when surrounding sectors have differing specials

electricbrass:
Addresses #1081

In case there are wads taking advantage of the vanilla behavior (which seems likely, as there are multiple prominent examples which the same lack of clipping in situations without sector specials), `r_clipmaskedspecial` can be disabled to revert back to it.

This isn't an ideal solution, but there doesn't seem to be one at the moment. Yet another problem from Eviternity II only being tested on GZ and DSDA...


electricbrass:
Closing as #1359 solves the same issue

PR #1387, opened 2025-10-05
Remove TArray

electricbrass:
`TArray` is a resizable array class that offers very little benefit over `std::vector`. I've replaced all instances of it with `std::vector`, giving us access to iterators and some other nice functions and removing the maintenance burden of `TArray`.


electricbrass:
I finally learned how cherry-picking works. I've had this sitting around as part of another branch with more extensive changes for months now.

PR #1388, opened 2025-10-07
[FEATURE] Add support for ZDoom extended GL nodes

electricbrass:
This PR adds support for 6 new node types: XGLN, XGL2, XGL3, and their compressed variants. At the moment, there are very few additional maps this allows, but this is a prerequisite for UDMF support, with XGL2 and XGL3 removing the 16-bit linedef limit and allowing for nodes with fractional precision, respectively.


electricbrass:
This also fixes a bug when loading maps with any of the compressed node types (including the already supported ZNOD), where, under certain circumstances, a pointer from Z_Malloc was passed to M_Realloc, messing up the OZone tracking of that memory.

PR #1403, opened 2025-10-26
Improve voting UI; Add sounds

rakohus:
Add vote UI sounds and key prompts; introduce cl_vote_sounds CVAR; update SNDINFO; add new sounds for voting.


electricbrass:
Should the cvar be `cl_` or `snd_`? Both have precedent (`cl_chatsounds` and `snd_gamesfx`) but it seems like maybe it should be `snd` for better organization, unless this is to align with other possible future `cl_vote_` cvars.

electricbrass (approved):

PR #1446, opened 2025-11-22
Build Odamex as C++20

electricbrass:
Now that we've made the jump into the post-C++11 world, I think it makes more sense to continue keeping up to date rather than sitting on a particular C++ standard for another 20 years. Since it hasn't even been a year since the move to C++17, this is here more to start discussion on this topic than to make the move to 20 right now.


HyperEye:
I don't think we should be in a rush to sunset support for moderately old platforms. When the C++ standard is increased versions of different Linux distributions get excluded. In addition the target version for the MacOS build has to be increased. We don't need to wait 20 years to bump the standard but we should be thoughtful about it.

PR #1447, opened 2025-11-22
Build Odamex as C++20

electricbrass:
Now that we've made the jump into the post-C++11 world, I think it makes more sense to continue keeping up to date rather than sitting on a particular C++ standard for another 20 years. Since it hasn't even been a year since the move to C++17, this is here more to start discussion on this topic than to make the move to 20 right now.


HyperEye:
I'm duplicating my comment on the previous ticket here since this one is open and a better place for discussion.

I don't think we should be in a rush to sunset support for moderately old platforms. When the C++ standard is increased versions of different Linux distributions get excluded. In addition the target version for the MacOS build has to be increased. We don't need to wait 20 years to bump the standard but we should be thoughtful about it.

electricbrass:
I don't think Linux is much of an issue. The vast majority of systems that Odamex runs on currently are supported just as well by newer versions of Linux. There's very few reasons why anyone would be keeping a computer used for gaming on any version thats new enough for C++17 but too old for C++20. Additionally, the Flatpak provides an option for running Odamex on such systems, though of course this doesn't help anyone wanting to compile it themselves. This PR still works on Ubuntu 20.04, which reached its EOL earlier this year and was the already the minimum targeted Ubuntu version since Odamex 11.

I'll also mention that UZDoom appears to be moving to to C++20 very soon (currently planned for 5.1), and Rum & Raisin is also on 20. There are a lot of additions that make the compile-time tools that were introduced in 11-17 able to be used much more cleanly and other improvements that make it a lot easier to write cleaner, more robust, and sometimes more efficient code.

macOS sounds like the main concern, and I'll admit I know very little about it. I understand that Macs that are not even very old can't update to the newest macOS, and I don't have any sense for how much of the Mac user base this would affect. So I will absolutely defer to anyone else who knows more about macOS. 

I'd also be curious to know what the current situation is on Nintendo Switch and whether this would cause issues there.

Here's what my personal hope is: moving to C++20 for Odamex 13 sometime next year, and from then on matching the 3 year schedule that there's been since C++11 (but about 6 years behind, e.g. moving to C++23 around 2029). With consideration of course for any current or future delays to adoption of any of those standards or the features within (I definitely don't think we should start using modules any time soon).

electricbrass:
On the topic of platform support, are our current Mac builds x86 only? I know Rosetta 2 support is being dropped soon.

HyperEye:
The Mac builds are x86. At some point we will either have to start producing two Mac builds (arm and x86) or get fat builds working so that the release supports both natively. We have time to solve the problem as Rosetta 2 support will continue until at least late 2027.

common/g_level.h
```diff
@@ -94,7 +94,13 @@ constexpr static levelFlags_t LEVEL2_COMPAT_CROSSDROPOFF = BIT(18);
...
 struct acsdefered_s;
 class FBehavior;
-struct bossaction_t;
+
+struct bossaction_t
+{
+	int type      = -1;
+	short special = 0;
+	short tag     = 0;
+};
```
leximayfield, line 104: Why did this get adjusted?  It's not obvious to me why this was needed to be moved, and o top of that, it's now missing the `MT_NULL` constant.

libraries/tinylibs/include/nonstd/bit.hpp
leximayfield, line 1: To me, part of what doing a C++20 upgrade entails is switching over to standard library constructs unless there is a compelling reason not to do so.

Is there a good reason not to simply switch to `std::bit`?

leximayfield (changes requested):
I am weakly in favor of a move to C++20.  The standard is 5 years old and I think most of the teething issues have been put to rest.  I think that even Apple's compiler defaults to C++20 now, and they're known for being slow on the uptake in terms of C++ features.

However, I would you to to see if perhaps some of our tiny libraries could be removed.  I would like to keep `fmtlib` because it provides features above and beyond what the standard has as of C++20, `expected` is C++23, and `scope` never made it in afaik.  However, `bit` and `span` I think could be candidates for removal.

libraries/tinylibs/include/nonstd/bit.hpp
electricbrass: Currently the main things being used are the extensions that don't exist in `std::bit`.

electricbrass:
> However, `bit` and `span` I think could be candidates for removal.

Agreed on `span`, it should go. In this PR, I was just only making the minimal changes required to make it compile.

libraries/tinylibs/include/nonstd/bit.hpp
electricbrass: The update to the library here is because older versions of it would disable the extensions and just directly used `std::bit`. This version uses `std::bit` for the things that exist there, but keeps around the extensions.

common/g_level.h
electricbrass: I don't remember exactly the reason why, but the usages of the type between the forward declaration and the actual definition later in the file fail to compile in C++20. Replacing MT_NULL with -1 was probably an oversight though.

electricbrass: https://godbolt.org/z/Erbd75f9q
With clang it fails a static_assert when trying to do pointer arithmetic with a pointer to an incomplete type.

electricbrass:
The build failure in 7e5750b8de416e8be72b5da09a041cbbc5fb23f5 is a runtime exception with C++17, and because of that ended up making it into 12.0.0. Preventing more bugs like this is another reason I am in favor of C++20.

electricbrass:
Hm, looks like we'll probably need to bump the Ubuntu 20.04 and Fedora 32 runners to 22.04 and 36 to be able to remove span. And likely for some other features as well.

leximayfield:
> Hm, looks like we'll probably need to bump the Ubuntu 20.04 and Fedora 32 runners to 22.04 and 36 to be able to remove span. And likely for some other features as well.

Hrm, I chose 32 specifically because it was what Red Hat Enterprise Linux 9 was based on.  RHEL 10 is based on 40 which is quite recent.

Though I suppose this broaches the idea of "why not Docker."  I suppose it's a matter of what our hosts are running and how they're running it.

libraries/tinylibs/include/nonstd/bit.hpp
leximayfield: Makes sense.

common/g_level.h
leximayfield: Ai yi yi.  Okay, was just curious, as long as it properly ends up as `MT_NULL`.

electricbrass:
> Hrm, I chose 32 specifically because it was what Red Hat Enterprise Linux 9 was based on.

That would be 34 actually. It was 28 previously (matching RHEL 8), and Ubuntu 18.04, but C++17 required something a bit newer so I updated the Ubuntu builds to 20.04 and moved the Fedora version forward by the same amount of time, not realizing the reason for the choice of 28.

electricbrass:
20.04 -> 22.04 for Ubuntu would takes it from GCC 9 to 11, which would get the vast majority of the C++20 features that are missing from 9. `span` in particular made it in in 10.

32 -> 34 to bring Fedora back in line with RHEL releases would also give GCC 11.

Acts19quiz:
> I don't think Linux is much of an issue. The vast majority of systems that Odamex runs on currently are supported just as well by newer versions of Linux. There's very few reasons why anyone would be keeping a computer used for gaming on any version thats new enough for C++17 but too old for C++20. Additionally, the Flatpak provides an option for running Odamex on such systems, though of course this doesn't help anyone wanting to compile it themselves. This PR still works on Ubuntu 20.04, which reached its EOL earlier this year and was the already the minimum targeted Ubuntu version since Odamex 11.

On the topic of Ubuntu, does Odamex 12 run on Ubuntu 18/19? I ask because I have an old 32-bit Vista computer I converted to Ubuntu, Ubuntu 18/19 is as far as x86 went, and while I don't do everyday gameplay on it, there are rare circumstances where I'll test/start a server on it.

(of course I'm describing a mostly server situation)

No big loss or win either way, just some food for thought.

electricbrass:
> On the topic of Ubuntu, does Odamex 12 run on Ubuntu 18/19?

It will not build with the default GCC on that version, but it should work if you install a later GCC if there's any available in the repos or a PPA or wherever.

I think the flatpak should also work for that Ubuntu version, though neither Flathub nor us build a 32-bit version.

Issue #1439, opened 2025-11-17
[BUG] Severe performance loss with very large amounts of enemies (or friendlies) spawned


fuzztooth:
**Describe the bug**
Performance degradation in large monster counts.  

**Build that the bug occurred in**
protobreak branch

**To Reproduce**
Friendlies implementation for ID24, also nuts.wad has severe performance loss.

**Expected behavior**
Should work like stable

**Screenshots, NetDemos, & Crash Dumps**
The call to RoughMonsterCheck in P_LookForMonsters is what is doing it

**Additional context**
Stable can handle nuts just fine, protobreak can't. when trying it the client will completely lock up and need to be killed with task manager.  A profiler said the that monster targeting was taking up basically all the cpu time available.

electricbrass:
This was significantly improved in #1452 but performance has still taken a decent hit.

bcahue:
OKU2V31.wad has 20k monsters and the recent protobreak fixes has significantly improved performance, including able to play that WAD above 60fps with friendlies on the field, so closing this.

PR #1452, opened 2025-12-01
[BUGFIX] Serious performance regression in monster targeting

electricbrass:
Addresses #1439

The new monster targeting code had a search based on Hexen's RoughBlockCheck. That function iterates over all the things in a blockmap block. It looks like when MBF21 was originally implemented for Odamex, a similar check was used for A_FindTracer, but because Odamex removed the `bnext` pointer used in the Hexen code years ago, `snext` was naively substituted, and this carried over to the friendlies implementation. This instead meant that these functions iterate over all things in a *sector*, causing serious performance issues on maps with high monster counts and large sectors.

Standing at the beginning of nuts.wad, on 11.2.0 I get 65-70fps on an i5-9300H at 1080p
On protobreak prior to this PR, the map is completely unplayable. Odamex completely locks up and I have to kill it from task manager.
With this PR, I now get about 50fps.

### DIFF
--- a/common/actor.h
+++ b/common/actor.h
@@ -473,8 +473,6 @@ public:
 
 	DWORD			effects;			// [RH] see p_effect.h
 
-    // Interaction info, by BLOCKMAP.
-    // Links in blocks (if needed).
 	struct subsector_s		*subsector;
 
     // The closest interval over all contacted Sectors.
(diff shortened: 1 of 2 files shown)


fuzztooth:
Even if this does not restore the same performance, it feels like it's in the right direction as it mostly removes code.

PR #1453, opened 2025-12-02
[BUGFIX] Crash when rendering railgun particles

electricbrass:
Also made it so NewParticle is defined consistently across different compilers


fuzztooth:
+1 for removing compiler-specific versions of functions!

Issue #1460, opened 2025-12-03
[BUG] Music switching linedef can cause music to be incorrect for nextmap


Hekksy:
**Describe the bug**
If a wad has a music changing linedef and then switches to another wad (via exit or nextmap), the music in the next wad will be incorrect, in my example Doom 2 map02's music from map01 to another wad's map01.

**Build that the bug occurred in**
Protobreak

**To Reproduce**
I used Chaos Below (chsblw.wad) and Gaia Aura (gaiaaura.wad) to find this bug. Have a server config setup with just gaiaaura.wad ("wad gaiaaura.wad" in the cfg file) and start the server. Then in  the server console change to chsblw.wad via the wad command. Set sv_skills 0, sv_allowcheats 1, and reload the map via "map map01." Enable noclip and turn around and run through the big door and keep going straight until you see an ominous stairway. Going downstairs triggers a music change. With the change, you can now "callvote nextmap" and the wad will go to Gaia Aura, which normally has replaced music. The music will instead be for Doom 2 map02 (incorrect)

**Expected behavior**
Loading the next wad after a music changer should play the next wad's correct music.

**Screenshots, NetDemos, & Crash Dumps**
Attached netdemo for quick reproduction

[musicbug.zip](https://github.com/user-attachments/files/23918407/musicbug.zip)

Hekksy:
Confirmed fixed

PR #1461, opened 2025-12-03
[BUGFIX] Switching wads plays the last music lump name changed to by musinfo

electricbrass:
Addresses #1460


electricbrass:
This actually affected any time maps changed. The first map loaded after leaving one where musinfo had changed the music, the music in the new level would be wrong.

fuzztooth (approved):

PR #1464, opened 2025-12-04
[BUGFIX] Serverchat only printing blank lines on clients

electricbrass:
Unfortunately compile time format string checking doesn't catch when there's more arguments than format specifiers...


bcahue (approved):
Looks good.

PR #1466, opened 2025-12-04
Clean up friendly effects code a bit

electricbrass:
Now there's less risk of forgetting to check cl_showfriends, and its all in one place.
Also adds a sentience check, so that wads using friendly flag for tricks with decorations don't have pillars and candlesticks that love you.


bcahue (approved):
Looks good here.

PR #1473, opened 2025-12-05
Implement new UMAPINFO keys `jumping`, `crouching`, and `freeaim`.

electricbrass:
A frequent issue with projects using UMAPINFO, is that if they are designed with the intention that crouching, jumping, and freelook are unavailable or disabled, they end up having to redefine all their maps in ZDoom to disable these features or other similar workarounds. To remedy this, a proposal to add freeaim, jumping, and crouching keys to UMAPINFO has been made. The exact details of the specification haven't been worked out just yet, hence this being just a draft for now, but the current wording can be viewed at https://github.com/kraflab/umapinfo/pull/6

### DIFF
--- a/common/g_umapinfo.cpp
+++ b/common/g_umapinfo.cpp
@@ -445,6 +445,11 @@ bool ParseStandardUmapInfoProperty(OScanner& os, level_pwad_info_t* mape)
 			break;
 		}
 	}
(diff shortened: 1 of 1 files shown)


electricbrass:
It looks like the discussion has reached an agreement on this revision. Exact wording of the spec hasn't been determined yet, but shouldn't have any further effect on this PR, so I'm marking this as ready.

PR #1504, opened 2025-12-19
Odasrv: Win 11 power throttle fix; retire Windows 7 support for odasrv

CurrentResident:
As of sometime in 2022, Windows 11 began throttling processes' access to the high-resolution timer by default.  The throttle also makes a heuristic decision about whether or not to slow down the process's execution regardless of whether or not the process requested a high-res timer.

Please see [the Remarks section of the `timeBeginPeriod` API documentation](https://learn.microsoft.com/en-us/windows/win32/api/timeapi/nf-timeapi-timebeginperiod#remarks) and the [`SetProcessInformation` API docs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setprocessinformation)

When odasrv gets auto-throttled on my Windows 11 rig, it causes `kbhit()` and `sendto()` to sleep for excessive lengths of time:

Screenshot 2025-12-19 082857

(The following two images are from a different session from above)

Screenshot 2025-12-22 164052

Screenshot 2025-12-22 164105

The fix is to call `SetProcessInformation` to disable the throttling states.  Because this API didn't exist until Windows 8 (or Windows Server 2012), these become the new minimum Windows versions for running odasrv.  Please note that this is *only* for odasrv - The odamex client is unaffected.

With throttling disabled, the odasrv profile becomes:

Screenshot 2025-12-19 093343

(The following two images are from a different session from above)

Screenshot 2025-12-22 163541

Screenshot 2025-12-22 163517

Note the reduction in time spent running ConsoleInput and SendPackets.

### DIFF
--- a/server/CMakeLists.txt
+++ b/server/CMakeLists.txt
@@ -44,6 +44,11 @@ odamex_target_settings(odasrv)
 target_include_directories(odasrv PRIVATE src)
 if(WIN32)
   target_include_directories(odasrv PRIVATE win32)
+
+  # We need to call SetProcessInformation to disable power throttling,
+  # but the API itself doesn't exist until Windows 8.  Here we set the minimum
+  # Windows version to 8.  (Windows Server 2012, BTW)
+  target_compile_definitions(odasrv PRIVATE _WIN32_WINNT=0x0602)
 endif()
 
 target_link_libraries(odasrv fmt::fmt tinylibs ZLIB::ZLIB odamex-common odaproto minilzo cpptrace::cpptrace)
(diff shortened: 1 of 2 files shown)


CurrentResident:
The Flatpak build jobs are failing due to a 404 while downloading Timidity.  I don't understand why this happens.

electricbrass:
> The Flatpak build jobs are failing due to a 404 while downloading Timidity. I don't understand why this happens.

It's cause the timidy patches are no longer at that location. We fixed the url on the protobreak branch since development has been focused there the last few weeks, but it isn't fixed yet on `stable`. In the next couple days,`protobreak` should get merged to `stable` and that will stop.

electricbrass (approved):

PR #1508, opened 2025-12-23
Only add resurrect powerup to pool if there are > 0 players with no lives left

electricbrass:


electricbrass:
Now also fixes the last player hype messages being broken by revives.

PR #1520, opened 2025-12-27
Fix potential out-of-bounds read in strbin

elf-alchemist:
Brought to our attention here: https://github.com/fabiangreffrath/woof/pull/2521
Related fix: https://github.com/kraflab/dsda-doom/pull/823

This is already fixed in G/UZDoom, but wasn't in ZDoom, even as late as 2.8.1


electricbrass:
Looks like this could also use a fix in d_dehacked
ReplaceSpecialChars is nearly the same function

electricbrass:
Oh this should've gone to stable

PR #1525, opened 2025-12-28
[BUGFIX] Initialize skyboxes on clients when MT_SKYVIEWPOINTs are transmitted

bcahue:
Fixes skyboxes not working online

### DIFF
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -674,6 +674,49 @@ static void CL_SpawnMobj(const odaproto::svc::SpawnMobj* msg)
 			mo->height = msg->args().Get(1) << FRACBITS;
 	}
 
+	if (type == MT_SKYVIEWPOINT)
+	{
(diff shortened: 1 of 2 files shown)


client/src/cl_parse.cpp
```diff
@@ -674,6 +674,52 @@ static void CL_SpawnMobj(const odaproto::svc::SpawnMobj* msg)
...
+		if (mo->args[0] == 0)
+		{
+			sector->Skybox = AActor::AActorPtr();
+		}
+		else
+		{
+			TActorIterator<AActor> iterator(mo->args[0]);
+			AActor* box = iterator.Next();
+
+			if (box->type == MT_SKYVIEWPOINT && box != NULL)
```
electricbrass: This check is in the wrong order

bcahue: Good catch, fixed

electricbrass (approved):

PR #1529, opened 2025-12-31
Cleanup: Replace a bunch of `new`/`delete` pairs with `unique_ptr` (or sometimes `vector`)

electricbrass:
Just some more cleanup to make it harder to accidentally introduce bugs.


common/m_fileio.h
```diff
@@ -191,3 +191,13 @@ std::vector<std::string> M_PWADFilesScanDir(std::string dir);
...
  */
 bool M_GetAbsPath(const std::string& path, std::string& out);
+
+struct FileDeleter
+{
+    void operator()(FILE* f) const
+    {
+        fclose(f);
+    }
+};
```
leximayfield: Please put this in a `detail`/`impl` namespace, or namespace the class to pair with the new name of `uqFile`.  Speaking of which...

common/w_wad.cpp
```diff
@@ -341,20 +337,18 @@ void AddFile(const OResFile& file)
 		header.infotableofs = LELONG(header.infotableofs);
 		size_t length = header.numlumps * sizeof(filelump_t);
 
-		if (length > (unsigned)M_FileLength(handle))
+		if (length > (unsigned)M_FileLength(handle.get()))
```
leximayfield: If you're doing cleanup, changing this to better type seems apropos.

common/w_wad.cpp
```diff
@@ -399,60 +389,48 @@ static bool IsMarker (const lumpinfo_t *lump, const char *marker)
...
 	// Some pwads use an icky hack to get flats with regular Doom.
 	// This tries to detect them.
-	flatHack = 0;
-	if (!strcmp ("F_START", ustart) && !Args.CheckParm ("-noflathack"))
+	size_t flatHack = 0;
+	if (start == "F_START" && !Args.CheckParm("-noflathack"))
 	{
 		int fudge = 0;
-		unsigned start = 0;
+		unsigned istart = 0;
```
leximayfield: Could you come up with a slightly better name?

PR #1544, opened 2026-01-08
Threaded console reader; streaming from stdin or file

CurrentResident:
Background
----------

Until this MR, Odamex had custom implementations of a terminal input reader, which performed character-by-character input, including backspace and newline handling.  The implementations were separate between both client and server and Win32 and non-Win32 platforms, leading ultimately to 4 distinct implementations to do this work, of which the win32 client was entirely commented-out thus nonfunctional.

Additionally, there was a separate facility to optionally open up a `std::ifstream` and simply read lines until EOF.  Like the terminal reader, this was duplicated between client and server.

Overview
-----------

This MR refactors and consolidates the console reading logic into a single module which is used by both client and server and on all platforms.  The reader fundamentally uses C++ istream character streaming facilities, relying on the platform to manage terminal-level character-by-character IO behavior.  By default, the reader sources `std::cin`, but if the `-confile` argument is given, an arbitrary file is used instead.  Input is read line-at-a-time, and the read operation itself is performed on a program-lifetime thread.  The thread blocks 1. when reading the input to a string via `std::getline` and 2. when waiting for the main thread to signal that it's finished moving the string's data to its own variable.  The main thread does not block during this process.

This is part of a broader effort to script load testing of server-side sending for large numbers of clients.  It enables a script to launch and somewhat monitor and control one or more clients via the console.

Testing
---------

- Scripted client testing with a Python script on both Windows 11 and RHEL 10
- "Interactive" Manual client and server testing on Windows 11 via Terminal and RHEL 10 via xterm+bash
- "Scripted" Manual testing on Windows 11 and RHEL 10 via `-confile`

Coding standard caveat
--------------------------

I took a best-guess at the right homes and naming conventions for files and routines.  By all means, let me know if I missed the mark.

Script demo
--------------

https://github.com/user-attachments/assets/02345d7f-53fa-4f03-a4e2-010f5a94b5e3


CurrentResident:
Test the case where the thread exits early due to EOF or IO error...  Joining would be the most correct thing to do here.  (The destructor either detaches or joins based on whether the platform was successful at commanding an early thread cancel/terminate)

ETA:  Yep, the early-exited thread can be commanded to terminate/cancel and then joined without issue.

bcahue:
This looks good to me, I will approve once I test this. May need to borrow someone to test on linux/OSX as well.

electricbrass:
> May need to borrow someone to test on linux/OSX as well.

Just tested on Linux and didn't see any issues.

bcahue:
Works for me too (on Windows 11). Approved.

bcahue (approved):

rakohus:
Needs to be brought up, and can then be merged.

PR #1548, opened 2026-01-10
Automated (unsigned) universal mac build

fuzztooth:
## Summary

- Build macOS universal deps in CI (SDL2/SDL2_mixer/wxWidgets/PCRE2/zstd) and bundle outputs consistently for release/RC.
- Replace ad‑hoc DMG creation with create-dmg, including background, layout, docs/config samples, licenses, and odamex.wad symlink.

## Testing

- GitHub Actions macOS RC/release runs (multiple passes)
- Manual tester verification on macOS (odamex and odalaunch launch, DMG layout OK)


electricbrass:
Is there a reason curl and fltk were changed to older versions?

Edit: a bunch of the other libraries too it looks like

fuzztooth:
> Is there a reason curl and fltk were changed to older versions?
> 
> Edit: a bunch of the other libraries too it looks like

I'm happy to have a discussion on why's on discord, but if possible please reference specific lines where something may be erroneously defined.  The only place I'm explicitly defining library versions is within the macos build actions.

electricbrass:
Sorry, there wasn't a line to refer to, it was the submodule commits that changed. I could have been clearer about that.

libraries/fltk-lib.cmake
```diff
@@ -2,13 +2,23 @@
 
 if(BUILD_CLIENT AND USE_INTERNAL_FLTK)
   set(_FLTK_BUILDGEN_PARAMS
-    "-DFLTK_USE_SYSTEM_LIBJPEG=OFF"
-    "-DFLTK_USE_SYSTEM_LIBPNG=OFF"
-    "-DFLTK_USE_SYSTEM_ZLIB=OFF"
-    "-DFLTK_OPTION_PRINT_SUPPORT=OFF"
-    "-DFLTK_BUILD_GL=OFF"
+    "-DOPTION_USE_SYSTEM_LIBJPEG=OFF"
```
electricbrass: These options got reverted to earlier FLTK 1.3 options that no longer exist in FLTK 1.4.

fuzztooth: Are you sure?  Everything I'm seeing is saying 1.4 uses DOPTION.

electricbrass: Commit 8eaec41 fixed this, but then it was undone by 2ec7591

fuzztooth:
> Sorry, there wasn't a line to refer to, it was the submodule commits that changed. I could have been clearer about that.

Yep I got that one.  Thank you.

libraries/fltk-lib.cmake
electricbrass: Yes, I am sure. You can see them here: https://github.com/fltk/fltk/blob/release-1.4.0/CMake/options.cmake. I also was the one who originally changed them from the 1.3 to the 1.4 options, because the old ones were no longer having any effect.

electricbrass: There is also a migration document listing all the options that changed: https://github.com/fltk/fltk/blob/master/documentation/src/migration_1_4.dox

fuzztooth:
Updated to use dotex

bcahue:
I'm good to approve this as long as we have verification of the DMGs from both the release and rc pipelines.

fuzztooth:
> I'm good to approve this as long as we have verification of the DMGs from both the release and rc pipelines.

@HyperEye has confirmed the package built and working as expected.

bcahue (approved):

Issue #1539, opened 2026-01-05
[FEATURE] Mini Automap Feature


thefatwilly:
**Is your feature request related to a problem? Please describe.**
Doom's base Automap feature has always fallen to the wayside given that the only interaction with it is either forfeiting your view altogether to review the wireframe map or have it obnoxiously cover your screen with the overlay that Odamex offers.

**Describe the solution you'd like to see**
Help bring Doom's automap into a more modern eye and have a separate, additional option to allow the automap to be visible in about a half a quarter of screen space in the corner of the player's pov. Scaling, color palette or whether or not to include a border is up to whomever implements the feature and the limitations of Odamex via the Doom engine. Preferable to maintain the player arrow to indicate direction.

**Describe alternatives you've considered**
As mentioned, the overlay wireframe Automap is an option, but takes up too much real estate and makes it difficult to navigate while also viewing the map.

**Additional context**
Good ol' Diablo 2 or 3 come to mind if there is any desire what their minimap might look like.

teesquared:
This patch is a first pass at a basic version. It allows you to adjust the overlay map so it can be on screen all the time. Check it out and see if this works. If it does, we can add the cvars to the options menu. You can adjust the zoom and other settings using the full screen map.

To use it, enable the overlay (am_overlay 1) then adjust the cvars to try the different settings.

```
// Overlay location (0:LeftTop, 1:LeftMid, 2:LeftBot, 3:RightTop, 4:RightMid, 5:RightBot)
set "am_ovlocation" "1"

// Scale height of overlay automap
set "am_ovscaleheight" "0.5"

// Scale width of overlay automap
set "am_ovscalewidth" "0.25"

// Transparency of overlay background
set "am_ovbackalpha" "0.25"

// Overlay background color
set "am_ovbackcolor" "00 00 00"
```

<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/4ef27bd8-d6ad-456c-b325-aed18da4b838" />

[oda12-minimap.patch](https://github.com/user-attachments/files/24494123/oda12-minimap.patch)

teesquared:
Fix for thick lines (apply to branch _feature/minimap-patch_)

[oda12-minimap-thick-fix.patch](https://github.com/user-attachments/files/24528082/oda12-minimap-thick-fix.patch)

bcahue:
Closed as implemented.

PR #1549, opened 2026-01-10
[FEATURE] Mini Automap Feature

teesquared:
Implements #1539 

This feature allows you to resize and reposition the overlay automap so it can be used as a minimap instead of a full screen overlay.


teesquared:
Per feedback in Discord, I'm going to make the overlay a mode (off, full screen, mini) that you set in the options menu. If it is in mini map mode then the map will also auto follow the player and use the other cvar settings to resize and position itself.

rakohus:
This looks good to me. Going to merge, if it needs tweaking later we can do that.

PR #1553, opened 2026-01-12
Replace DeuTex with "dotex" implemented in Python

leximayfield:
Pretty simple - this is a python script that is capable of building odamex.wad.

It parses the same manifest file as DeuTex, but then uses asset conversion routines and WAD assemblers that are written by me.  All of the source assets have been turned into png files, and I have renamed all of our sounds so they have their original .wav/.ogg file extensions.

The `dotex` routines I chose to place into the public domain because I'm honestly sick of rewriting them at this point in my life, but the actual builder script that uses those routines to assemble odamex.wad specifically has copyright assigned to the Odamex team as per usual.

I have smoke-tested it myself, and all of our builds pass, but I'd appreciate some extra eyes on it, just to make sure I didn't overlook something.


fuzztooth:
I don't know how strongly anyone else feels about it, and it's very minor, but perhaps it'd be better if dotex was in tools/ rather than wad?  odawad.py can stay in wad that seems fine.

wad/flats/-noflat-.png
fuzztooth, line 1: I think the consensus on this one is the older colors are preferred.

leximayfield: Fair, I forgot about that when I opened the PR.  Fixed.

CurrentResident:
Ha, nice!  ..And just when I made a branch on my fork that builds DeuTex as a submodule.

fuzztooth (approved):
🫡 Deutex 🫡

PR #1555, opened 2026-01-13
Remove buggy qpc check, commonize the high-res time utility code

CurrentResident:
## Background

The Win32 implementation of `I_GetTime` had an old workaround for a PerformanceCounter bug on some WinXP-era systems that, under certain conditions, could erroneously cause the returned value to "freeze" forever.  To be clear, the function itself wouldn't freeze up - rather, it would just start returning the same value forever.

This is triggered if the time spent between calls to `I_GetTime` exceeds a certain timing window.  (e.g. Spending a lot of time at a breakpoint in a debugger)

The original PerformanceCounter bug itself is the counter not being synched between cores / CPUs on some systems, resulting in possible inconsistent counter values if the OS migrates the process between cores / CPUs at runtime.  An application would perceive this as a discontinuous time jump.

## Fix

Remove the workaround.  According to the Microsoft docs, Windows 7 goes to lengths to correct the counter sync bug, and Windows 8 improved / optimized the solution.

## Yes, and...

This PR also consolidates the timer functions between the client and the server so that we don't have to maintain two copies of the functions.  Because this makes the diff view a little tough to read, here is the specific commit that removes the workaround:  https://github.com/odamex/odamex/commit/646daae8e85b9a70b83111a5c33d0bbb1a517ec0


common/i_time.cpp
```diff
@@ -0,0 +1,156 @@
...
+		QueryPerformanceCounter(&initial_count);
+
+		initialized = true;
+	}
+
+	LARGE_INTEGER current_count;
+	QueryPerformanceCounter(&current_count);
+	return static_cast<dtime_t>(nanoseconds_per_count * (current_count.QuadPart - initial_count.QuadPart));
+
+#else
```
electricbrass: I might change this to
```
#elif defined CLIENT_APP
...
#else
#error "idk some error about not having a way to get time on this system"
```
just to make it more informative if someone is compiling the server on some non-Windows, non-Unix system instead of an error from failing to find SDL_GetTicks, that might be misleading.

CurrentResident: Sounds good, will do.

electricbrass (approved):

PR #1564, opened 2026-01-20
Minidump all accessible memory

CurrentResident:
### Background

Win32 crash dumps typically only contain enough process data to follow two levels of pointer indirection from the perspective of the relevant callstack.  This can be helpful for relatively simple crashes, but for cases where entities get into bad states, having access to snapshots and other data well outside of the crashing callstack would *really help* with post-mortem debugging.

### Fix

Instruct MiniDump to capture all accessible process memory.

#### Pros

Absolutely all data can be inspected.

#### Cons

The .dmp for my manufactured test crash (1 minute into a horde) goes from ~300 KB to ~360 MB.  It's huge!

#### But that's okay because

- Crashes are infrequent.
- Crashes due to an intermittent complex data situation have a better chance of being understood.


fuzztooth (dismissed):

electricbrass:
I think we want to keep dumps small enough for being attached to github issues and discord messages by default. On discord it was suggested to create both types of dumps, which I think could be an okay solution. Personally, I would make it only produce the smaller dumps by default, with the larger dumps an option for developers.

CurrentResident:
Further comment from Discord:

Hi all, regarding the [minidump PR](https://github.com/odamex/odamex/pull/1564):  How about we continue to produce timestamped small dumps, so those can continue to accumulate in the dump dir, but we only allow for a single, massive full dump at a time, such that the full dump always reflects the latest crash?  That way we continue to have small ones that can be posted on Discord, but have the option of asking the user to drop the full thing in a dropbox/onedrive/etc, without requiring them to take extra actions to enable full dump and just hope that they can reproduce the same crash.
..and without monster .dmps accumulating unbounded in the dump dir.

electricbrass:
> How about we continue to produce timestamped small dumps, so those can continue to accumulate in the dump dir, but we only allow for a single, massive full dump at a time, such that the full dump always reflects the latest crash?

Was this change made yet?

CurrentResident:
Yep, back here:  https://github.com/odamex/odamex/pull/1564/changes/920e71906cdba4a668b55eaed30a84272c89c10a

Thx!

electricbrass (approved):

PR #1565, opened 2026-01-20
Add odamex:// url registration for mac builds

### LINKED ISSUE #1441: [BUG] `odamex://` URIs don't work on macOS
**Describe the bug**
On Windows and Linux, the Odamex client is registered as a handler for `odamex://` URIs. On macOS, it is not.

All of em

**To Reproduce**
Click this link on a Mac: [odamex://odamex.electricbrass.net](https://odamex.electricbrass.net/join?server=survival)

**Expected behavior**
The Odamex client should open and connect to the specified server.

**Additional context**
Something like this is needed in the Info.plist, and I think a small Swift or Objective-C program that receives the URI and launches Odamex with the proper arguments is needed.
```xml
<key>CFBundleURLTypes</key>
<array>
    <dict>
        <key>CFBundleURLName</key>
        <string>Odamex</string>

        <key>CFBundleURLSchemes</key>
        <array>
            <string>odamex</string>
        </array>
</array>
```

fuzztooth:
Adds registration of the odamex:// url protocol upon installation to the MacOS platform using the DMG. 

- Generates client plist file at build time via cmake
- Enhance SDLMain.h with additional functionality to handle this
- Route to SDL_Main first (macos) to capture the args
- Generalize how the url parmaters are captured across platforms

### Additionally:

- Compensate for funkier branch names on a couple of the builds that were failing
- Handle fedora 32 EOL image issues


client/CMakeLists.txt
```diff
@@ -125,8 +128,12 @@ if(APPLE)
   find_library(CARBON_LIBRARY Carbon)
   find_library(AUDIOTBOX_LIBRARY AudioToolbox)
   find_library(AUDIOUNIT_LIBRARY AudioUnit)
+  find_library(SCREENCAPTUREKIT_LIBRARY ScreenCaptureKit)
```
fuzztooth, line 131: This seems to be required on newer macos versions.  The runners are older so they probably don't need this.

PR #1582, opened 2026-02-01
Reliability handling - odatest fixes

CurrentResident:


electricbrass:
As long as we're making the odatests setup steps more detailed, might as well specify that it should be v1.9 of the IWADs.

Also the nested list on step 3 has all of its steps labeled as 1.

CurrentResident:
> As long as we're making the odatests setup steps more detailed, might as well specify that it should be v1.9 of the IWADs.
> 
> Also the nested list on step 3 has all of its steps labeled as 1.

Done.

The repeated "1." is because the markdown will still render as properly ordered, but if / when we go back to add or remove steps, we save ourselves the trouble of having to manually renumber potentially a lot of other steps.  If you feel strongly otherwise, I'll go back and change it.

electricbrass:
> The repeated "1." is because the markdown will still render as properly ordered, but if / when we go back to add or remove steps, we save ourselves the trouble of having to manually renumber potentially a lot of other steps. If you feel strongly otherwise, I'll go back and change it.

I had no idea!

electricbrass (approved):

PR #1583, opened 2026-02-01
[BUGFIX] UDMX map32 teleporters blocking players

electricbrass:
There are a few related changes here:

1. A minor memory leak in P_LoadLineDefs that occurred when loading linedefs with invalid vertices has been fixed.
2. MBF21 linedef flags in Doom format maps are now always cleared if `ML_RESERVED` is set. This matches the behavior of DSDA or Woof when `comp_reservedlineflag` is enabled. The actual comp option is only needed for playing back a small subset of early MBF21 demos, which we can't play back anyway.
3. E2M7 is now handled by 2, but `p_compdb.cpp` has been introduced, with a table of level hashes and corresponding compatibility options. Currently it only handles the block players flags in Congestion 1024 map23 and UDMX map32, but it's designed to easily add more as issues are discovered.

### DIFF
--- a/common/p_compdb.cpp
+++ b/common/p_compdb.cpp
@@ -1,0 +1,75 @@
+// Emacs style mode select   -*- C++ -*-
+//-----------------------------------------------------------------------------
+//
+// $Id$
+//
+// Copyright (C) 1998-2006 by Randy Heit (ZDoom).
+// Copyright (C) 2006-2026 by The Odamex Team.
+//
+// This program is free software; you can redistribute it and/or
+// modify it under the terms of the GNU General Public License
+// as published by the Free Software Foundation; either version 2
+// of the License, or (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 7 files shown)


common/p_compdb.cpp
```diff
@@ -0,0 +1,75 @@
...
+		                        (size_t)(f.fingerprint[3]) << 24 |
+		                        (size_t)(f.fingerprint[4]) << 32 |
+		                        (size_t)(f.fingerprint[5]) << 40 |
+		                        (size_t)(f.fingerprint[6]) << 48 |
+		                        (size_t)(f.fingerprint[7]) << 56;
+		return halfhash;
+	}
+};
+
+static const std::unordered_map<fhfprint_t, levelcompdata_t> compdata = {
```
fuzztooth, line 47: It would be nice in the near future to get this info into a data format like json.

electricbrass: I could do that now. I considered it but didn't see the benefit.

fuzztooth (approved):
I think this is fine with the understanding that the compdb is inevitably going to need further iteration.

PR #1588, opened 2026-02-04
Reliable client --> server channel for player-initiated commands

CurrentResident:
### OdaMessenger

New utility for queuing outgoing and incoming messages, assembling packets, ordering reliable messages, and managing reliable retransmissions.  This utility is responsible for the overall improved reliable messaging, and the logic it implements has some key differences from the 12.x and prior behavior:

- Mid-frame packet sends on account of packet size are removed in favor of queuing for a single call to `SendAll`.
- Each client has its own messenger instance, and these instances are Thread-safe WRT each other.

### Multithreaded server-to-client messaging

The server code now includes a thread pool, and the writing of messages and transmission of packets is done in parallel.

The thread count is controlled by a new CVAR `net_maxthreads`.  A value of `0` (the default) means that the thread count will be set automatically based on local core count.

**Allowing multithreading is necessary for good server performance on maps with very large monster and player counts.**

Multiple thread-safety fixes were made in the network-related areas.  The fixes mostly came down to moving data objects from being static to client-specific, mostly as part of the Messenger instance.

### Client -> server reliable messaging

This MR brings the reliable sender and receiver together into a messenger utility, which the server and client now both use for building and sending packets.

This creates a client-to-server reliable channel, and this MR moves the player-initiated client commands into that channel.

The 10-slot circular buffer previously used for client commands was retired and replaced by configuring the client's reliable sender with a retransmission delay of 0 tics and and a maximum retransmit packet limit of 10.  This causes the tic-to-tic packet content to behave very similarly to the previous.  Assuming a 3-tic latency, the first three tics will see a sequence of packets that contain:

- tic 1:  command 1
- tic 2:  command 1*, command 2
- tic 2:  command 1*, command 2*, command 3

So far, the command sequence essentially matches the previous behavior.  However, with the reliability messenger, the next tics become:

- tic 3: command 2*, command 3*, command 4
- tic 4: command 3*, command 4*, command 5

The end result is that the server will always receive the client commands in order and won't miss any unless the communication breaks down so badly that the client's transmission queue fills up, which at default settings would take 7 seconds of complete non-responsiveness.

### Reliable messaging stats on `cl_netgraph`

Visual indication of how many reliable packets are "in flight" at any given time.

### Instrumentation:  Stopwatches for measuring real-time usage

New console commands:

- `instenable <regex>`: Enable one or more watches matching the given expression.
- `instdisable <regex>`: Disable one or more watches matching the given expression.
- `instrecord <filename>`: Record the enabled watches as a CSV file, with the given filename.
- `inststop`: Stop recording.

### "Canary" socket to detect ungracefully-terminated clients

The canary is a TCP connection between the client and server that exchanges no real data - only OS-managed keepalive echo/ack sequences.  Its purpose is so that the server can detect, drop, and cleanup after a client that terminates ungracefully.  The keepalive probe runs once every 10 seconds.

This includes a UPnP port mapping update to allow both UDP and TCP on the configured server port.

### Remove `net_[snd|rcv]buf` cvars, hardcode to 128 KB

Having these as CVARs has just led to trouble again and again.

### Change the `buf_t` data store from being a fixed heap-allocated array to a `vector`

This makes moving, copying much more natural and compiler-optimizable.

### Separately track Read position and Write position in `buf_t`

This helped remove some hoop-jumping needed to get the packet headers created.

### Fix for player corpses vanishing upon respawn

I accidentally induced this in protobreak a while ago.

### Fix rare crash due to race condition between PlayerState and LoadMap

If the sprite definitions change from one map to the next, and there's stale psprite data left in the player state after the sprites are swapped out, an out-of-bounds access can crash the client.

The fix was to make sure that the reliable PlayerInfo message that accompanies the Load Map message has the complete set of player data, and ensure that both PlayerInfo and PlayerState contain the same.


bcahue (approved):

bcahue:
Approved, its near 2000 net new lines of code, but at a glance it seems proper. We can iterate on it once we get to testing.

Issue #956, opened 2024-05-04
[BUG] When time skipping in a netdemo, the recorder's skin isn't saved from the POV of other players' perspectives.


Acts19quiz:
**Describe the bug**
When playing a back a netdemo with several players, the recording player uses a skin colour besides the green default, you use F12 to change the POV to another player, then use time skip to go forward or backward, the recording player's skin reverts to the default green.

Specifically, this involves the 20 second forward and backward skips that are by default bound to the right and left arrow keys, respectively. This doesn't involved the map skips with the up and down arrow keys.

**Build that the bug occurred in**
10.3.0 GB28EE-8578
10.4.0 G46E0E1-8728

**To Reproduce**
1. Ensure that you have Odamex setup to record a netdemo.
2. Play an online game with more than one player.
3. Ensure either that all players have distinctive colour skins besides the green default, or that it's a team game (CTF, Team DM, or Team LMS).
4. Play for long enough (at least two minutes, ideally a minute or so more) that you have enough room to time skip around.
5. Play the netdemo back.
6. Change the POV (with F12) to another player besides the recorder.
7. Use time skip to go ahead or backwards by 20 seconds.
8. Find a point in time where the recording player steps in front of the player whose POV you're following.

**Expected behavior**
For the recording player's skin colour to match what that player is actually using, and to maintain that colour through all time skips.

**Screenshots, NetDemos, & Crash Dumps**
![Odamex_HORDE_20240504_122204_edit](https://github.com/odamex/odamex/assets/63941237/dbc0b392-5a32-484d-8ec6-7847c32bbb8c)
The first screenshot shows what the recording player looks like if the netdemo is played out from the beginning, and what that player should look like.
![Odamex_HORDE_20240504_122127_edit](https://github.com/odamex/odamex/assets/63941237/c0cfedd6-3567-44a5-863f-7bbb6cadcb82)
The second screenshot shows what the recording player looks like if time skip is used.

[The netdemo](https://github.com/odamex/odamex/files/15210510/Odamex_HORDE_20230501_211421_HORDAMEX.WAD_MAP16.zip) using the above screenshots. It requires DOOM2.WAD and HORDAMEX.WAD. It's recorded with Odamex 10.3.0. 10.3.0 is used instead of the latest version (10.4.0 as of now) for a reason explained below.

**Additional context**
[...]

PR #1594, opened 2026-02-10
[FEATURE] Rebase Vanilla/indexed skin colors on 12.1.0.

Acts19quiz:
This is an effective duplicate of #932 in order to port those changes from Odamex 10.4.0 to 12.1.0.

Here are the change I've done:

* Incorporated Sean's suggestion of having a mid-break ENUM in d_netinf.h (`enum colorpreset_t`) that eliminates a magic number.

Here are outstanding issues/questions:

1.
In a netdemo, the recording player doesn't show skin color updates. However, this is a preexisting bug (#956). It predates both this PR, and the introduction of colour presets altogether.

2.
cl_main.cpp in 10.4.0 had this section:

```
	for (Players::iterator it = players.begin(); it != players.end(); ++it)
	{
		R_BuildPlayerTranslation(it->id, CL_GetPlayerColor(&*it));
	}
``` 

The iterators have been replaced with an auto function.

```
	for (auto& player : players)
	{
		R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), NUMCOLOR);
	}
```

This does simplify the code. However, I couldn't find an equivalent for `.begin()`, which is the first player rang up on the register, not necessarily Player ID 1. This is necessary because the indexed translator needs to stop if the player it's translating is subject to `r-forceblankcolor`, but ignore this exception if it's the viewing player running the game.

```
	for (Players::iterator it = players.begin(); it != players.end(); ++it)
	{
		if (it == players.begin() || consoleplayer().spectator || !G_IsTeamColor(r_forceteamcolor, r_forceenemycolor))
			R_BuildPlayerTranslation(it->id, CL_GetPlayerColor(&*it), it->userinfo.colorpreset);
		else
			R_BuildPlayerTranslation(it->id, CL_GetPlayerColor(&*it), NUMCOLOR);
	}
``` 

The one other thing to mull over: Needing player information to pass both the RGB color and the colorpreset is less than ideal, but those are two distinct values that are treated differently. The only away around that I could thing of would be to have out of bounds RGB values hardcoded to indexed translations. E.g., send 255,255,256 if you want indexed red, 255,255,257 for blue, etc. That seems like a gross hack, though, which is why I didn't go that route.

NOTE: Demos tested, 96 of 106 passed. Player power-up tinting also seems unaffected.


common/g_gametype.cpp
```diff
@@ -344,6 +344,14 @@ bool G_IsTeamGame()
 	return sv_gametype == GM_TEAMDM || sv_gametype == GM_CTF;
 }
 
+/**
+ * @brief Check if team colors are being enforced due to CVARs and gametype.
+ */
+bool G_IsTeamColor(bool forceteamcolor, bool forceenemycolor)
```
electricbrass: This seems like it also needs to consider team games

client/src/cl_main.cpp
```diff
@@ -217,14 +220,28 @@ argb_t CL_GetPlayerColor(const player_t& player)
...
+		if (player == players.begin() || consoleplayer().spectator || !G_IsTeamColor(r_forceteamcolor, r_forceenemycolor))
+			R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), player.userinfo.colorpreset);
+		else
+			R_BuildPlayerTranslation(player.id, CL_GetPlayerColor(&player), NUMCOLOR);
+	}
+*/
+
+	for (Players::iterator it = players.begin(); it != players.end(); ++it)
+	{
+		if (it == players.begin() || consoleplayer().spectator || !G_IsTeamColor(r_forceteamcolor, r_forceenemycolor))
```
electricbrass: I don't understand the need for the `players.begin()` check here. There shouldn't be any special meaning for a player being the first in the vector.

electricbrass (changes requested):

common/g_gametype.cpp
Acts19quiz: I tried some team games to see what would happen. Everything looks fine so far. The team colour translations are applied to whatever the indexed colour looks like.

Odamex_CTF_20260216_113414
Odamex_CTF_20260216_113439
Odamex_CTF_20260216_113451
Odamex_CTF_20260216_113512

electricbrass: This feels like then there's some duplicated logic then, if all the G_IsTeamColor calls are returning false in team games yet it still works.

client/src/cl_main.cpp
Acts19quiz: Yeah, this is something I touched on in my opening notes. I figured it would come up, which is why I was looking for a way to check for the first player rang up in the iterator in the way it's currently used.

I did some internal testing. The reason for the check for the first entry is due to the `r_forceXXXcolor` CVARs. If a player sprite should be subjected to them (e.g., I'm looking at an opponent and have `r_forceenemycolor 1` to force all enemies to be yellow), I do **_not_** want that sprite to be subjected to indexed translation even if that player's skin is red/blue/indigo since indexed translation prevents those CVAR's own translations from taking place. So `G_IsTeamColor` performs a check that stops indexed translation if either of those CVARs are relevant.

**_However_**, if we're talking about the viewing player, we want it to be always included even if those CVARs are active. The following screenshots are from a test DM game.

Odamex_DM_20260216_112126

Screenshot 1 is indexed translation acting as normal. This looks as intended. If we don't check for `r_forceXXXcolor`, this is also what's shown even if `r_forceenemycolor 1` is set since the forced colour translator doesn't run properly.

Odamex_DM_20260216_112145

Screenshot 2 is with the viewing player using `r_forceenemycolor 1` and the first player in the iterator being excluded with `players.begin()`. This also looks as intended (the enemy gets the player-forced translation).

Odamex_DM_20260216_112504

Screenshot 3 is with the viewing player using `r_forceenemycolor 1` and `players.begin()` removed from the function. Uh oh! Both sprites stopped using indexed translation because the translator doesn't know to exclude the first player in the iterator (which happens to be whichever client is being viewed).

electricbrass: So should this be a check for `displayplayer`?

common/g_gametype.cpp
Acts19quiz: This may be due to me not using the most intuitive name for the function. It doesn't really check for team colours (as implied by the name); all it really does is check the status for the `r_forceXXXXcolor` CVARs.

electricbrass: Yes, but if the cvars are enabled and it's a team game, the function will return false still. It will always return false in a team game instead of reflecting the value of the cvars.

client/src/cl_main.cpp
Acts19quiz: That was my first thought, but it complained to me that they were not the same operand type. E.g., `player == displayplayer()` (current iterator method) or `it == displayplayer()` (prior iterator method).

electricbrass: That's because `player_t` doesn't have an `operator==`. `&player == &displayplayer()` should work

common/g_gametype.cpp
Acts19quiz: Hmm. So what's the suggestion here on how to handle this/the following function in g_gametype.cpp (if removing the function entirely prevents the r_forceXXXXcolor CVARs from working)?

`bool G_IsTeamColor(bool forceteamcolor, bool forceenemycolor)`

client/src/cl_main.cpp
Acts19quiz: That did the trick, thanks. Next iteration of commit will reinstate the current iterator function and rephrase the check this way.

common/g_gametype.cpp
electricbrass: Ideally it would be to figure out *why* does it still work in team games, and see if it could be cleaned up some. It also just seems like that `G_IsTeamColor` should probably have `|| ((r_forceteamcolor || r_forceenemycolor) && G_IsTeamGame())` appended to the condition. Then it would actually return true in all cases where those cvars take effect, and false otherwise.

Acts19quiz: I made a minor change to the following line, then started a coop game to see what would happen:

`return (forceteamcolor && G_IsCoopGame()) || (forceenemycolor && G_IsFFAGame());`

to:

`return (forceteamcolor && G_IsCoopGame()) || (forceenemycolor);`

`r_forceenemycolor`, even if enabled, doesn't do anything in the background if it's a coop game. However, when the check for a DM game is removed as a condition for checking for `r_forceenemycolor`, the index translator stops working in coop games because it doesn't know that we don't care about `r_forceenemycolor` in this context.

Odamex_COOP_20260216_122013

`r_forceenemycolor 0` in a Coop game (or with G_IsFFAGame() checked).

Odamex_COOP_20260216_122018

`r_forceenemycolor 1` in a Coop game. Uh oh; the indexer is deferring to `r_forceenemycolor` and disabling itself in a context we don't want it to.

Acts19quiz: Making the following change in cl_main.cpp:

`if (&player == &displayplayer() || consoleplayer().spectator || ((r_forceteamcolor || r_forceenemycolor) && G_IsTeamGame()))`

Caused this in a Coop game:

Odamex_COOP_20260216_131904

The indexer stops working on the other player. Since the other player is not the display player, the display player is not a spectator, and it's not a team game, the indexer has no chance to run.

Acts19quiz: After further investigation, https://github.com/odamex/odamex/pull/1594/changes#diff-35f3796b0e2ee42c05c8fed9743f4c95e57a617cba3b61ce67d0d0480e7658ccR553 isn't working quite right. Even when CTF is the game mode (which is G_IsTeamGame true, which means it should skip indexed translation), and the player colour preset is purple (which is past 6 in the enumerator), it's still applying vanilla translation. I.e., it's indexed red being applied to the non-indexed purple, so it's red, but clearly a Vanilla-oriented red. Doing down to something dark that's also not indexed, like Jungle Green or Black, creates a dark indexed red.

I'll need to investigate further.

fuzztooth:
If this is still WIP, please mark it as draft.

common/g_gametype.cpp
Acts19quiz: After looking into this more, I believe I can render an answer.

First, the thing about the indexing happening in CTF was wrong. That was operator error; I had forgotten to enable 32-bit colour on my Debug build, which made things looked indexed that weren't. Oops! I enabled it, and found indexing was not happening in CTF, and everything was working as intended. Sorry about that!

Now, on to the question on hand:

https://github.com/odamex/odamex/pull/1594/changes#diff-35f3796b0e2ee42c05c8fed9743f4c95e57a617cba3b61ce67d0d0480e7658ccR553

This is the section that ensures team games are never subjected to indexed translation; I don't want indexing happening in team games. Veteran or hardcore CTF players are probably used to their reds and blues looking a certain way, regardless of the underlying colour they've selected.  An underlying RGB value is submitted even for colours that otherwise qualify for indexed translation.

https://github.com/odamex/odamex/pull/1594/changes#diff-b47f2fe5b53d3072d187614fbc36c06db9b887ced19e52cc124d249209f4b318R350

This is the section in question here. This section actually doesn't care for or check for team colours. `forceteamcolor` is checked for, however the term team is a bit of a red herring here. This isn't used exclusively in team games; it can also display in coop games (every player is on your team; you're on a de facto team of many). 

Same with `forceenemycolor`. This is also used for DM FFA; every player is on the opposite team (you're a de facto team of 1).

So why are we checking for FFA in the case of `forceenemycolor` when the first line cited ensures team games never get this far?

Odamex_COOP_20260216_122018

We want `forceeXXXXcolor 1` to override indexed translation, but only in situations where they would ordinarily display. If `forceenemycolor 1` is on in a DM game, we want it to disable indexing; however if we don't check for DM, it will also stop indexing in Coop even when there are no enemies possible! In the screenshot above, `forceenemycolor 1` is used in a Coop game, but indexing is still disabled because `G_IsTeamColor` isn't checking to ensure it's FFA.

It works the other way around, too: If `forceteamcolor 1` is on in Coop, we want it to stop indexing; but if it's on in DM, it will still stop indexing (despite friendlies being impossible) unless we tell it to check for Coop!

tl;dr:

1. Team games never make it to `G_IsTeamColor`, indeed.
2. However, none of the checks in `G_IsTeamColor` have anything to do with team games; `G_IsTeamColor` already assumes anything it gets isn't a team game.
3. The extra game mode checks in `G_IsTeamColor` just make sure team[mate] colour checks aren't performed in FFA, and enemy colour checks aren't performed in Coop.

Proposal:

The original one, rename `G_IsTeamColor` to something more appropriate, like `G_IsForcedColor`.

Acts19quiz:
Marked back "Ready for review"; the error I experienced before was operator error (I had forgotten to enable 32-bit colour, which made things look off when they were fine). Sorry for the confusion.

EDIT: Also fair milestone warning, Odamex 12.1.0 stable crashes on this PR (why 11 Protobreak was the target when this was based on 10.4.0)

client/src/cl_main.cpp
```diff
@@ -814,18 +817,19 @@ BEGIN_COMMAND (playerinfo)
...
 		       GetTeamInfo(player->userinfo.team)->ColorizedTeamName());
 	}
-	PrintFmt(PRINT_HIGH, " userinfo.aimdist - {:d} \n",		player->userinfo.aimdist >> FRACBITS);
-	PrintFmt(PRINT_HIGH, " userinfo.color   - {:s} \n",		color);
-	PrintFmt(PRINT_HIGH, " userinfo.gender  - {:d} \n",		player->userinfo.gender);
-	PrintFmt(PRINT_HIGH, " time             - {:d} \n",		player->GameTime);
-	PrintFmt(PRINT_HIGH, " spectator        - {:d} \n",		player->spectator);
-	PrintFmt(PRINT_HIGH, " downloader       - {:d} \n",		player->playerstate == PST_DOWNLOAD);
+	PrintFmt(PRINT_HIGH, " userinfo.aimdist     - {:d} \n",		player->userinfo.aimdist >> FRACBITS);
+	PrintFmt(PRINT_HIGH, " userinfo.colorpreset - {:s} \n",		player->userinfo.colorpreset);
```
electricbrass: This will crash when trying to print, because it's trying to treat colorpreset as a string.

common/g_gametype.cpp
electricbrass: > The original one, rename `G_IsTeamColor` to something more appropriate, like `G_IsForcedColor`.

I worry that just renaming it might lead to future issues if/when this code gets modified again, because `G_IsForcedColor` still implies it should be true any time you're using it to check if player colors are being forced, which may be true currently because it is only called for non-team games, but relying on assumptions like that is very fragile.

client/src/cl_main.cpp
electricbrass: You can remove the `:s` to have it use the formatting function you defined, but I would probably recommend having a function to convert it to a string of the name instead of just printing out a number.

common/g_gametype.cpp
Acts19quiz: The team check could (probably) be moved to this function with no real impact throughout the rest of the code.

client/src/cl_main.cpp
Acts19quiz: That :s was just an oversight, oops!

I was mostly aping how gender is printed in like fashion. It's also printed as an integer, and didn't think at that time to give colorpreset special treatment. But if you think that's a good idea, I could certainly go for that.

electricbrass: > I was mostly aping how gender is printed in like fashion. It's also printed as an integer

Ah okay, I missed that other fields work that way. Up to you then

Acts19quiz: Well, I'd like for it to print strings, but I'd equally like for it to be consistent with the rest of playerinfo. But changing the others is just a touch out of scope with this PR.

What about print strings now, and change the other prints at some point in the future? (or making a 'casserole' commit with the others changed)

common/g_gametype.cpp
Acts19quiz: Uh oh. I moved the `G_IsTeamGame()` check to `G_IsForcedColor`. This introduced a problem in the following line:

https://github.com/odamex/odamex/pull/1594/changes#diff-8bd30503895285a591b60aece579e5a43be48b31859f48d18ca78f69258ec90cR226

`&player == &displayplayer()` ensures that if, say, `r_forceteamcolor 1` is used in a Coop game, the other players get the forced colour, but the display player stays with indexed translation (if applicable).

However, now indexed translation is being retained for the display player in team games, which is undesirable, because `&player == &displayplayer()` is unaware of whether or not it's a team game.

The counter-solution:

`if (&player == &displayplayer() && !G_IsTeamGame() || consoleplayer().spectator || !G_IsForcedColor(r_forceteamcolor, r_forceenemycolor))`

So what we've really done is moved the `G_IsTeamGame()` check to cl_main.cpp, and added a redundant check to `G_IsForcedColor` to make it semantically correct (catch all forced colors).

Acts19quiz: Another uh-oh, moving to cl_main.cpp didn't fully resolve it either. What happens then is the player is correctly not indexed in team games, but the player sprite in the player setup menu (m_menu.cpp) still shows indexed translation.

https://github.com/odamex/odamex/pull/1594/changes#diff-35f3796b0e2ee42c05c8fed9743f4c95e57a617cba3b61ce67d0d0480e7658ccR553

The team check must be here in order to catch the player sprite both in-game and in the player setup menu so as to eliminate duplicate `!G_IsTeamGame()` checks throughout. So, the end result is not eliminating the `!G_IsTeamGame()` check, but adding a `G_IsTeamGame()` check to `G_IsForcedColor` to make it fully accurate and representative.

Acts19quiz: If `&player == &displayplayer()` would work in r_draw.cpp, that would eliminate a lot of duplicated code.

Acts19quiz: Okay, @electricbrass I think we're cooking with gas here.

I figured out how to put the display player check into r_draw.cpp. The entire function of G_IsTeamColor (now R_IsForcedColor) has been retooled to be the sole entity responsible for checking all circumstances that a specific color might be enforced in. A lot of duplicated code and checks have been eliminated.

G_IsTeamColor, now called R_IsForcedColor, handles sprite conditions exclusively and really has less to do with game modes now, so it's been moved.

I couldn't get the text strings for gender and color preset to be presented. sv_main.cpp in particular complained about typedefs or something. That's another topic for another day, though.

electricbrass: Alright, I think this looks a lot better now! I think there's just one last little issue. I'll make a new conversation for that.

client/src/cl_main.cpp
```diff
@@ -1323,6 +1324,7 @@ void CL_SendUserInfo(void)
 	MSG_WriteString	(&net_buffer, coninfo->netname.c_str());
 	MSG_WriteByte	(&net_buffer, coninfo->team); // [Toke]
 	MSG_WriteLong	(&net_buffer, coninfo->gender);
+	MSG_WriteLong	(&net_buffer, coninfo->colorpreset);
```
electricbrass: Adding this field to the middle of this message will result in the client-to-server userinfo message being incompatible with 12.0 and 12.1. Moving this, and the corresponding read in sv_main, to happen last in this function should work without any issue. (With Odamex 13 the client messages will start using protobuf too and this should no longer be an issue.)

common/d_netinf.h
```diff
@@ -42,23 +42,33 @@ inline auto format_as(gender_t eGender)
 	return fmt::underlying(eGender);
 }
 
-enum colorpreset_t	// Acts 19 quiz the order must match m_menu.cpp.
+enum colorpreset_t // [Acts 19 quiz] The order must match m_menu.cpp.
```
electricbrass, line 45: Changing the order of this enum will cause issues when playing on 12.0/12.1 servers alongside players on 12.0/12.1 clients. Players using presets will be displayed with the wrong ones.

electricbrass (commented):
There are two options for these last two issues that I see:
1. Change them both to keep compatibility with earlier 12.x versions
2. Change the base branch to protobreak (I believe these changes should be able to be rebased cleanly, or at least with minimal issues) and get this in for Odamex 13 instead of 12.2

> EDIT: Also fair milestone warning, Odamex 12.1.0 stable crashes on this PR (why 11 Protobreak was the target when this was based on 10.4.0)

Just noticed this comment. My guess is that's probably due to adding the color preset to the client userinfo message.

common/d_netinf.h
Acts19quiz: Original order:

```
	COLOR_CUSTOM,
	COLOR_BLUE,
	COLOR_INDIGO,
	COLOR_GREEN,
	COLOR_BROWN,
	COLOR_RED,
	COLOR_GOLD,
	COLOR_JUNGLEGREEN,
	COLOR_PURPLE,
	COLOR_WHITE,
	COLOR_BLACK,

	NUMCOLOR
```

New order:

```
	COLOR_GREEN,
	COLOR_INDIGO,
	COLOR_BROWN,
	COLOR_RED,
	COLOR_BLUE,
	COLOR_ORANGE,
	COLOR_GOLD,
	NUMVANILLACOLOR = COLOR_GOLD, // doesn't count towards total
	COLOR_JUNGLEGREEN,
	COLOR_PURPLE,
	COLOR_WHITE,
	COLOR_BLACK,
	COLOR_CUSTOM,

	NUMCOLOR
```
The reason for the order change is that how this works at all is by reappropriating the indexed translator meant to only be used with Vanilla netgame demos. The indexer there has INDIGO (gray) set as 1, BROWN as 2, and RED as 3. I've put GREEN at 0 since I'm assuming that's what the netdemo Vanilla indexer would do anyway.

https://github.com/odamex/odamex/blob/a0e7732265b70527a66f678430c94da5fa4ac379/client/src/r_draw.cpp#L487C6-L487C38

I'm trying to keep the order so that this function will get the colors in the order it expects them--enum 0 is assumed GREEN, enum 1 will be INDIGO, enum 2 will be BROWN, and enum 3 will be RED; and these enums will be the same all throughout.

So, the order within `R_BuildClassicPlayerTranslation` can't be changed. That would mess with Vanilla netdemos, which is bad. It currently assumes green for 0, but I've taken the liberty of stating this explicitly for non-demo uses (and adding a couple of extra colours for 4 and 5, which are impossible scenarios in Vanilla netdemos).

If we take the current enum list order, BLUE is 2. A player that wants BLUE (actual blue, mind you, not fake DOOM indigo gray) will have that enum passed to the Vanilla indexer, which will then turn it into indexed gray. We don't want that.

A player wanting GREEN, which as of now is enum 3, will get passed to the Vanilla indexer, and get a RED sprite instead.

So I had to change the list order to give them to the indexer in the order they expect them in.

That's why this was originally assigned to Protobreak 11 while stable was 10; we may have to wait for 13.

Acts19quiz: Ultimately, moving this to Protobreak is what would need to be done.

electricbrass (dismissed):

common/d_netinf.h
electricbrass: It would be possible to just convert the numbers sometime before passing them to the vanilla indexer. But I agree its just simpler to move it to protobreak.

electricbrass (dismissed):

electricbrass:
What happened? This was up to date with protobreak and passing just a bit ago

electricbrass:
Oh, looks like you redid my rebase and committed an old version of protobuf

Acts19quiz:
Yeah, I did a re-base on Protobreak so I could prepare it for me to change the base, not knowing you had already did that. I fetched the latest Protobreak earlier today for what it's worth.

electricbrass:
Just need to undo the protobuf change and this should be good to go. Protobreak is using a slightly newer version of protobuf and so you need to make sure to update your submodules

electricbrass (approved):

Issue #1587, opened 2026-02-02
[BUG] Neon Overdrive map01 multiplayer crash


nikovaan:
**Describe the bug**
We tried to play Neon Overdrive in survival co-op online with 4 people and map01 kept crashing either game clients or the server. There was also one crash with 3 people after which it stopped crashing for whatever reason?

**Build that the bug occurred in**
12.1.0

**To Reproduce**
We've not had a chance to try and reproduce the crashes.

**Expected behavior**
Neither game client nor server should crash.

**Screenshots, NetDemos, & Crash Dumps**
Stream when it happened: https://www.twitch.tv/videos/2684491535?t=1h35m23s (URL valid until 6.2.2026)

[odamex neonover map01 crash  files.zip](https://github.com/user-attachments/files/25017770/odamex.neonover.map01.crash.files.zip)

**Additional context**
The server was first hosted with server_mbf_cfg. After the first crash Is witched to server_dsdahacked.cfg, since I thought "oh well maybe this was a dsdahacked pwad. After the third server crash I switched back to server_mbf.cfg, since I went to check and found that no it was mbf21. The last client crash was when there were only 3 players and we were back to the mbf config file.
Attached archive includes 3 client crash dumps, 3 server crash dumps, and both config files.
Link to wad: https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/neonover

electricbrass:
Okay it looks like there's a few different reasons for crashes here:
- `odamex_g_26688_20260130T203721.dmp`
  - It crashes in `P_SmartMove` because `target->subsector` is null. `actor` is an imp, and for some reason `target->type`
 is `MT_NULL` and `target->health` is greater than 0. Off the top of my head at least, I have no idea what could cause this. https://github.com/odamex/odamex/blob/7a791a9e9a4b7c5d16d0d93a5e0fc3086b7dc1f5/common/p_enemy.cpp#L464-L473
- `odamex_g_28200_20260130T204340.dmp`
  - Way down inside calls to `P_RecursiveSound` a line with an invalid sidenum[1] was accessed. So it seems like either this is a problem with the map, with a single sided line somehow being marked as two sided, or the memory holding the linedef data was corrupted.
https://github.com/odamex/odamex/blob/7a791a9e9a4b7c5d16d0d93a5e0fc3086b7dc1f5/common/p_enemy.cpp#L161
- `odamex_g_34084_20260130T204122.dmp`
  - Exact same thing
[...]

PR #1596, opened 2026-02-12
[BUGFIX] Crash when propagating sound in maps with single-sided lines marked as double-sided

electricbrass:
Partially addresses #1587

The seg-loading functions for extended nodes were missing a check to detect such lines and remove the double-sided flag.

Also contains a fix for a crash that can occur when rendering maps with XGL nodes, as those can contain segs with null linedefs.


electricbrass:
Well, this turned into a much larger PR than expected. `BYTE`/`DWORD`/`SDWORD`/etc had annoyed me for a long time already, but this time they actually really got in the way of my attempts at making the byteswapping functions more template friendly and more portable, so I decided that they needed to die. I've now removed these types, and every instance of them, and replaced them with `stdint` types.

fuzztooth (approved):

PR #1601, opened 2026-02-16
Add "HUD" graphic for said section in options menu.

Acts19quiz:
Add a custom HUD menu graphic for "HEADS-UP DISPLAY" in said menu section instead of reusing "DISPLAY OPTIONS".

The lump "M_HUD" is shared across other source ports for the same text.


electricbrass (approved):

fuzztooth (approved):
Approving, though I think it may need to be further addressed in the future given the font.

PR #1607, opened 2026-02-18
Fix jitter when playing a demo and changing cl_movebob

### LINKED ISSUE #527: [BUG] Netdemo playback "jumpy" jitter depending on cl_movebob settings
**Describe the bug**
Netdemo playback will jitter vertically if `cl_movebob` value does not match that of the recorder

Odamex 10

**To Reproduce**
Record a netdemo with `cl_movebob .2`. End the demo. Launch the demo and change to `cl_movebob 1`. The view should start to jackhammer as it tries to reset the vertical position of the player camera 

**Expected behavior**
Play smooth regardless of `cl_movebob` value at the time of recording

[Odamex_CTF_20220125_233416_WDL2021.WAD_MAP01.odd.zip](https://github.com/odamex/odamex/files/7938963/Odamex_CTF_20220125_233416_WDL2021.WAD_MAP01.odd.zip)

^^Recording on wdl2021.wad with `cl_movebob .2`

endless-r0ad:
Long story short but setting viewz here is not necessary. At this point viewz has the bob value from the demo built into it. Prevviewz is set with this viewz but then viewz is overwritten in P_CalcHeight (using user's bob value). Now when there was a large discrepency between prevviewz (demo) and viewz (user) the jitter would happen in interpolateCamera(). That large discrepency would happen when user's bob setting is different than the demo's. 

Solution is removing the intial set of viewz by the demo so that prevviewz will bet set with the viewz made in P_CalcHeight.  Also pretty sure its not necessary to write viewz at all to demos, didnt want to do that without permission though.

Fixes #527

### DIFF
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -2958,10 +2958,13 @@ static void CL_NetdemoCap(const odaproto::svc::NetdemoCap* msg)
 	player_t* clientPlayer = &consoleplayer();
 	fixed_t x, y, z;
 	fixed_t momx, momy, momz;
-	fixed_t pitch, viewheight, deltaviewheight; // ,viewz
+	fixed_t pitch, viewheight, deltaviewheight;
 	angle_t angle;
 	int jumpTics, reactiontime;
 	byte waterlevel;
+
+	// Note clientPlayer->viewz should not be set with the value from the demo here
+	// it is an aggregate value and will be set correctly later
 
 	clientPlayer->cmd.clear();
 	clientPlayer->cmd.unserialize(msg->player_cmd());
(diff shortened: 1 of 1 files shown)


electricbrass:
This looks like it's probably fine. I would rather the lines be removed fully instead of just commented out though, probably with a comment left to explain why viewz isn't being set.

We should also be sure that this doesn't cause any issues with netdemos staying in sync due to this line, which I believe is the only one where viewz has any effect that isn't just visual:
https://github.com/odamex/odamex/blob/acd1b941b3cad2448d20cfa3075e6d04473008b7/common/p_spec.cpp#L3050

endless-r0ad:
Sounds good, Ill continue testing and revise as needed

endless-r0ad:
Confirmed that viewz is set already earlier in the sim loop (P_PlayerThink -> P_CalcHeight) so there are no changes to the previous functionality. Only change is prevviewz is now set with the correct viewz (not the one from the demo)

image

endless-r0ad:
Realized I never tested multiplayer demos, diffrerent gamemodes and such. Found no issues, should be good

fuzztooth (approved):

Issue #782, opened 2022-09-20
[FEATURE] Add client option to disable view bobbing but keep weapon bobbing


Calinou:
*Related to https://github.com/odamex/odamex/issues/781.*

**Is your feature request related to a problem? Please describe.**
When disabling view bobbing for accessibility reasons (reducing motion sickness) or gameplay reasons (always having your "true" height represented correctly), weapon sprite bobbing is also disabled. This makes it look like moving around has that infamous "jetpack on low batteries" effect.

**Describe the solution you'd like to see**
Add a cvar and menu option to scale view bobbing height independently of the weapon bobbing intensity. This could be either a boolean or a float that goes between `0.0` and `1.0` (full intensity).

The movement bobbing option would remain available in the menu.

**Describe alternatives you've considered**
Disabling view bobbing works OK, but it doesn't feel that great to play due to the weapon sprite remaining static. This can be fine for competitive players, but it's not ideal when you want *some* feedback on movement actions to remain visible.

**Additional context**
[Doomsday Engine](https://dengine.net/) can achieve this via a console command, but not any other source port I know of. That said, I'd like to give a shot at this implementing in GZDoom in the future… and Odamex if this feature is approved here :slightly_smiling_face:

PR #1614, opened 2026-02-23
Change movebob behavior with cl_movebobtype

endless-r0ad:
This would add two new types of movebob behavior beside the default. The type will only trigger under the same conditions that cl_movebob is triggered (sv_allowmovebob is on or playing locally). 

0 = weapon only bobbing
1 = weapon and view bobbing (default, current behavior)
2 = view only bobbing

Would complete #782


electricbrass:
Just remembered a message I'd sent in discord that sort of got lost: I wonder if instead cl_movebob should be split into 2, one that affects weapon bob and one that affects viewbob, both with the 0.0-1.0 range. That gives more flexibility for controlling amount for each separately. However it does come at the cost of making it harder to set them both together. To help with this, cl_movebob could be replaced with a console command that sets both at the same time. I wonder what anyone else's thoughts on this are.

endless-r0ad:
> Just remembered a message I'd sent in discord that sort of got lost: I wonder if instead cl_movebob should be split into 2, one that affects weapon bob and one that affects viewbob, both with the 0.0-1.0 range. That gives more flexibility for controlling amount for each separately. However it does come at the cost of making it harder to set them both together. To help with this, cl_movebob could be replaced with a console command that sets both at the same time. I wonder what anyone else's thoughts on this are.

I think that's a great idea and would probably be easier for the user to understand. IMO no need for a repurposed cl_movebob since its only 1 extra step for the user, setting both cvars be the same, in order to match current cl_movebob behavior. Would probably want Movebob X and Movebob Y sliders in the display menu as well.

endless-r0ad:
Closing per above discussion

PR #1617, opened 2026-02-26
Change the client->server NetCommand into protobuf ClientCommand

CurrentResident:
### Summary

- Swapped out the bespoke `NetCommand` class with a new protobuf auto-coded `ClientCommand`.
- Added human-readable `ClientCommand` contents to the `hud_demoprotos` display.
- Housekeeping: Retired `player_s`, make the player class just `player_t`, which the vast majority of the code already uses.

Does *not* affect netdemo file format.

Screenshot 2026-02-26 171131


electricbrass:
Maybe it's just because this is just part of the larger transition towards using protobuf for client messages, but it feels strange to me to be mixing the client and server commands like this in the `svc_t` enum and `server.proto` file. Is the plan to eventually move all client messages to a `client.proto` and move the `clc` messages back to their own enum? Also, I might name ClientCommand something more descriptive like InputCommand, since it's just one of many types of client commands (not that the old NetCommand name was any better).

And thank you for cleaning up the `player_s`/`player_t` stuff. It's always really bugged me, and I was just thinking last night how I might open a PR to get rid of `player_s`.

CurrentResident:
> Maybe it's just because this is just part of the larger transition towards using protobuf for client messages, but it feels strange to me to be mixing the client and server commands like this in the `svc_t` enum and `server.proto` file. Is the plan to eventually move all client messages to a `client.proto` and move the `clc` messages back to their own enum?

Yes, I guess I should explain where I picture this going.  I envision "svc" and "clc" merging and becoming just "msg" - all message types and the enumeration just composing a single collection without any possibility of conflicting definitions / identifiers, just a single data dictionary.

My thought was to gradually retire "clc" messages and create "svc" editions of the messages, and then once clc is gone, rename "svc" to "msg,"

I can see an argument that I've really only taken the first half of the first step here because I just added "CLC_" functions... which, yeah...  Let me know if you'd like me to change that.

> Also, I might name ClientCommand something more descriptive like InputCommand, since it's just one of many types of client commands (not that the old NetCommand name was any better).

I'm down with renaming it, will do.

What I *ultimately* want to do is break ClientCommand into logical messages that make up the higher-level application interfaces.  e.g.  PlayerAttack, PlayerImpulse, PlayerMove, etc.  The fact that we have one message with multiple optional fields (which is 1:1 with what NetCommand had) is a clue that the message we have is essentially a shoehorning of multiple messages into one payload.  I mean, I could start going down that road now, but I thought it would be best / more easily readable to do it in a future PR.  Please let me know if you feel otherwise.
 
> And thank you for cleaning up the `player_s`/`player_t` stuff. It's always really bugged me, and I was just thinking last night how I might open a PR to get rid of `player_s`.

My pleasure!

electricbrass:
> Yes, I guess I should explain where I picture this going. I envision "svc" and "clc" merging and becoming just "msg" - all message types and the enumeration just composing a single collection without any possibility of conflicting definitions / identifiers, just a single data dictionary.
> 
> My thought was to gradually retire "clc" messages and create "svc" editions of the messages, and then once clc is gone, rename "svc" to "msg,"

Ah okay. I kind of like the idea of keeping them separate largely for organization. In general the client and server are going to be sending different things, right? So it feels like it should be two different things for keeping it easier to understand what gets used where and making it obvious what code is relevant to each half. But I can also see an argument for it being simpler to have it all unified. I at least don't have a particularly strong opinion here, just wanted to know what the plan was.

> What I _ultimately_ want to do is break ClientCommand into logical messages that make up the higher-level application interfaces. e.g. PlayerAttack, PlayerImpulse, PlayerMove, etc. The fact that we have one message with multiple optional fields (which is 1:1 with what NetCommand had) is a clue that the message we have is essentially a shoehorning of multiple messages into one payload. I mean, I could start going down that road now, but I thought it would be best / more easily readable to do it in a future PR. Please let me know if you feel otherwise.

I think the old NetCommand was basically just a copy of Doom's original `netcmd_t` with extra fields tacked on over time and logic for (de)serializing it. Properly updating it and splitting it apart definitely seems like the way to go. No need for it to be right now though.

CurrentResident:
Okay, how about this - the messages' names will identify their originator just as they do now ("svc" or "clc"), and the handling code for each will be separated as they are in this PR  (svc_message, clc_message), but ultimately the enumeration that identifies them will be unified.  (I do feel strongly that allowing more than one message type to be associated with a single ID / numeric value as it would appear in a header creates ambiguity and an opportunity for misinterpretation.  In other words, I want to be able to look at a captured header and data payload and be able to decode it without any special knowledge beyond the message library itself.)

So that way, we can know where a message came from at a glance when we see its name, and we know exactly what interface it represents when looking at the encoded payload's header.  Sound good?

electricbrass:
> So that way, we can know where a message came from at a glance when we see its name, and we know exactly what interface it represents when looking at the encoded payload's header. Sound good?

Sounds good to me

electricbrass (approved):

PR #1620, opened 2026-02-27
Fix protobreak compilation errors

electricbrass:
C++20 PR was out of date so it hadn't replaced some of the more recent uses of nonstd::span with std::span


fuzztooth (commented):
thanks and sorry!

fuzztooth (approved):

Issue #1623, opened 2026-03-01
[BUG] Player skin color is not properly changed when starting a single-player game while online.


Acts19quiz:
**Describe the bug**
When starting a single-player game while in the middle of an online game, the player skin color (visible using chasecam) is not correct, and does not match what is in PLAYER SETUP. If the online game was a team game, it may display the color of the team the player was on, or it may display a default green.

**Build that the bug occurred in**
0.8.0; 12.1.0

**To Reproduce**
_A: Mismatched team game sprite_
1. Launch immediately into an online team game match.
2. Go to PLAYER SETUP and view the preview sprite.
3. Start a new single-player game, and use the chasecam.

_B: Default sprite_
1. Start Odamex into a single-player game.
2. Go to console, and join an online game.
3. Don't view PLAYER SETUP, or it will trigger the first variant instead!
4. Start a new single-player game, and use the chasecam.

**Expected behavior**
For the player sprite to always be correct, and match what is in PLAYER SETUP

**Screenshots, NetDemos, & Crash Dumps**
Progression of situation A.
<img width="1601" height="900" alt="Image" src="https://github.com/user-attachments/assets/2496b173-b108-407a-ad13-a692db7277f6" />
<img width="1601" height="900" alt="Image" src="https://github.com/user-attachments/assets/4306674c-f906-4f11-915e-1e2d46e14351" />

Progression of situation B.
<img width="1601" height="900" alt="Image" src="https://github.com/user-attachments/assets/26635b0b-64ed-484f-80c6-f264c67b1095" />
<img width="1601" height="900" alt="Image" src="https://github.com/user-attachments/assets/07376c43-7546-482f-ba80-5d4920687367" />
<img width="1601" height="900" alt="Image" src="https://github.com/user-attachments/assets/a467197c-7cd3-4088-a5ec-22fc1589696f" />

**Additional context**
n/a

PR #1625, opened 2026-03-02
Properly refresh player sprite when jumping from online to offline.

Acts19quiz:
Made to address #1623

The call to `R_BuildPlayerTranslation` in d_netinfo.cpp is the only call that occurs upon every start of a single-player game. For whatever reason, `consoleplayer_id` isn't properly catching the player id when going from mutliplayer to single-player.

The condition check before ensures that this never impacts online games or netdemos, and just concerns single-player.

Protobreak due to the fix needing to account for the indexed player skins.

### DIFF
--- a/client/src/d_netinfo.cpp
+++ b/client/src/d_netinfo.cpp
@@ -203,7 +203,7 @@ void D_SetupUserInfo(void)
 
 	// update color translation
 	if (!demoplayback && !connected)
-		R_BuildPlayerTranslation(consoleplayer_id, color, coninfo->colorpreset);
+		R_BuildPlayerTranslation(0, color, coninfo->colorpreset);
 }
 
 void D_UserInfoChanged (cvar_t *cvar)


LotusOpenclaw (approved):
.

electricbrass (changes requested):
Id 0 is nullplayer, the id of consoleplayer in singleplayer should be 1, not 0. Even then though, I think this isn't a fix that addresses the real problem here. It would be best to find out if the id is being set incorrectly elsewhere.

Acts19quiz:
I tried 1, didn't work. Perhaps there could be a deep dive performed if you think trouble brews elsewhere.

Acts19quiz:
It could be something to do with ID assignment from multiplayer to single-player considering I had to do a direct jump to grief it; single player startups are never affected.

electricbrass:
It looks like at startup, the player ids haven't actually been set when D_SetupUserInfo is first called, so it *is* actually using 0 then. Everything being set there in coninfo is actually garbage data due to the consoleplayer() call earlier in the function also getting player 0 instead of player 1. So there's definitely a deeper issue here.

Acts19quiz:
So what's the plan? Set up a dragnet of breakpoints checking player id, and find which ones gives a rotten answer?

Upon starting a single-player game, your ID is 1, and I'm fairly certain when I did a breakpoint for consoleplayer_id in that call, it gave 1 as well, so.....

electricbrass:
My guess is that whats happening is that there's an off by one error in R_BuildPlayerTranslations that is being bypassed by the incorrectly initialized consoleplayer_id at startup, but then later when switching from multiplayer to singleplayer, the correct id is used and the off-by-one becomes visible.

Acts19quiz:
Maybe. I specifically scoped this out to 0.8.0 to eliminate any recent changes causing this.

Acts19quiz:
Thank you for looking into this. Sorry if I sent you down a rabbit hole.

PR #1631, opened 2026-03-03
[BUGFIX] Various issues related to player colors

electricbrass:
Addresses #1623

This is an alternative to #1625 after doing a deep dive on the cause of the issue. It turns out there were a lot of related issues here, which are all fixed by this PR:
1. consoleplayer and displayplayer ids were not being initialized on startup, so anything using them at startup would be going to id 0 instead of 1
2. when setting everything up at startup, only the translation for the "player" in the player setup menu was being initialized. the ingame player in singleplayer appeared correct anyway due to issue 1, as id 0 is the id of the menu player
3. the initial translation being built for the menu player always used the classic green translation due to a missing call to D_ColorPreset, and string cvars still having the ability to implicitly convert to a float (but would be correct instantly upon opening the menu)

There's also some minor code cleanup, including using `menuplayer_id` instead of a literal 0 whenever the player setup menu "player" is being referred to for purposes of translations.


drivingspectrum (dismissed):
Looks good to me.

fuzztooth (approved):

PR #1634, opened 2026-03-03
Service reliable channels until idle when gracefully disconnecting

CurrentResident:
### Summary

There are three main types of disconnects:

- Ungraceful
- Server-initiated graceful
- Client-initiated graceful

Graceful disconnects are fundamentally finalized by the server sending a reliable message - either Disconnect, Reconnect for the server-initiated cases, or DisconnectClient for client-initiated.  When the client sees this happen, its job is to send the reliable Acknowledge for the finalization message, drain remaining packets, then go back to idle.

The server takes the disconnecting clients' messengers, clears out any outgoing queued gameplay messages, sends the finalization message, and continues to service them (including the Reliable channel) until the final ack comes in.

Ungraceful disconnects due to the canary closing continue to operate as before:  the disconnect is applied immediately and the messenger is dropped immediately.

### Testing

Under both "good" network conditions and "bad" network conditions as induced via Clumsy:

- Client: `disconnect`
- Client: `reconnect`
- Client: `quit`
- Client: External force-kill
- Forced temporary disconnect to download a wad and then reconnect
- Server: `quit`
- Server: `rquit`
- Server: `kick <player id>`


drivingspectrum (dismissed):
LGTM

fuzztooth (approved):

PR #1636, opened 2026-03-04
[BUGFIX] Thing lighting ignoring height transfers when r_thingsectorlight is disabled

electricbrass:
r_thingsectorlight didn't account for height transfers, always using the light level of the "real" sector. Now it should always account for height transfers.


drivingspectrum:
Is this a WIP? I see it's been set as a draft PR

electricbrass:
> Is this a WIP? I see it's been set as a draft PR

Yes there were a few questions I had about how it was supposed to work. Should be ready now.

fuzztooth (approved):

PR #1646, opened 2026-03-13
Bring protobreak up to date with stable

electricbrass:


electricbrass:
Just realized this should probably wait til it can bring #1645 with it, otherwise it will introduce the bug that that fixes to protobreak, which will prevent a number of maps from loading.

drivingspectrum (approved):

PR #1647, opened 2026-03-14
Add a bunch of subsector null checks

CurrentResident:
This was prompted by the crash reported here:

https://discord.com/channels/236518337671200768/236518337671200768/1482140638349885541

### DIFF
--- a/common/p_boomfspec.cpp
+++ b/common/p_boomfspec.cpp
@@ -1245,12 +1245,12 @@ void P_PlayerInCompatibleSector(player_t& player)
 	if (player.spectator)
 		return;
 
-	// Falling, not all the way down yet?
-	if (player.mo->z != P_FloorHeight(player.mo) && !player.mo->waterlevel)
-		return;
-
 	// Being destroyed / unlinked from the map?
 	if (not (player.mo && player.mo->subsector && player.mo->subsector->sector))
+		return;
+
+	// Falling, not all the way down yet?
+	if (player.mo->z != P_FloorHeight(player.mo) && !player.mo->waterlevel)
 		return;
 
 	sector_t& sector = *player.mo->subsector->sector;
@@ -1383,12 +1383,12 @@ void P_PlayerInCompatibleSector(player_t& player)
 //
 bool P_ActorInCompatibleSector(AActor* actor)
 {
-	if (not (actor && actor->subsector && actor->subsector->sector))
+	if (not (actor && actor->subsector))
 		return false;
 
-	sector_t* sector = actor->subsector->sector;
+	const bool sectorHasKillMonstersAction = actor->subsector->sector & KILL_MONSTERS_MASK;
 
-	if (sector && sector->special & KILL_MONSTERS_MASK && actor->z == actor->floorz &&
+	if (sectorHasKillMonstersAction && actor->z == actor->floorz &&
 	    !P_IsPlayerOrAvatar(*actor) && actor->flags & MF_SHOOTABLE && !(actor->flags & MF_FLOAT))
 	{
 		P_DamageMobj(actor, NULL, NULL, 10000);
--- a/common/p_boomfspec.cpp
+++ b/common/p_boomfspec.cpp
@@ -1246,7 +1246,7 @@ void P_PlayerInCompatibleSector(player_t& player)
 		return;
 
 	// Being destroyed / unlinked from the map?
-	if (not (player.mo && player.mo->subsector && player.mo->subsector->sector))
+	if (not (player.mo && player.mo->subsector))
 		return;
 
 	// Falling, not all the way down yet?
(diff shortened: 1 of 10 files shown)


common/p_zdoomhexspec.cpp
```diff
@@ -226,6 +226,9 @@ void P_PlayerInZDoomSector(player_t& player)
 	if (player.mo->z != P_FloorHeight(player.mo) && !player.mo->waterlevel)
 		return;
 
+	if (not (player.mo && player.mo->subsector))
```
electricbrass: `player.mo` check needs to be earlier, it's already being dereferenced a couple lines earlier.

common/p_boomfspec.cpp
```diff
@@ -1379,7 +1383,7 @@ void P_PlayerInCompatibleSector(player_t& player)
 //
 bool P_ActorInCompatibleSector(AActor* actor)
 {
-	if (!actor)
+	if (not (actor && actor->subsector && actor->subsector->sector))
```
electricbrass: Would be cleaner to not be checking the sector in both if conditions.

common/p_boomfspec.cpp
```diff
@@ -1249,6 +1249,10 @@ void P_PlayerInCompatibleSector(player_t& player)
 	if (player.mo->z != P_FloorHeight(player.mo) && !player.mo->waterlevel)
 		return;
 
+	// Being destroyed / unlinked from the map?
```
electricbrass: `player.mo` check needs to be higher, it's already being dereferenced a couple lines earlier.

common/p_enemy.cpp
```diff
@@ -140,6 +140,11 @@ void P_RecursiveSound (sector_t *sec, int soundblocks, AActor *soundtarget)
 	line_t* 	check;
 	sector_t*	other;
 
+	if (not sec)
+	{
```
electricbrass: I would make sec a reference instead of a pointer. `other` is guaranteed non-null (well, unless something has gone *very* wrong, but then that would mean the map data is totally corrupted and we probably want it to crash at that point), so checking at the call site in P_NoiseAlert would be enough and eliminates the need for checks in the recursive calls.

common/p_map.cpp
```diff
@@ -1249,10 +1252,14 @@ bool P_TryMove (AActor *thing, fixed_t x, fixed_t y,
 {
 	fixed_t		testz = thing->z;
 
-	if(!thing->subsector)
+	if(not thing->subsector)
 	{
 		I_Error("P_TryMove: Thing {{type: {}, info->type: {}}} subsector was null", thing->type, thing->info->type);
 	}
+	if(not thing->subsector->sector)
```
electricbrass: `subsector` member of `subsector_t` is guaranteed non-null.

common/p_mobj.cpp
```diff
@@ -471,18 +471,9 @@ void P_CheckTouchy(AActor* mo)
...
-	{
-		levelgravity = FixedDiv(FLOAT2FIXED(level.gravity), 100 << FRACBITS);
-		sectorgravity = FLOAT2FIXED(mo->subsector->sector->gravity);
-	}
-	else
-	{
-		levelgravity = GRAVITY * 8;
-		sectorgravity = FLOAT2FIXED(mo->subsector->sector->gravity);
-	}
+	const float   sectorGravityFloat = (mo && mo->subsector && mo->subsector->sector) ? mo->subsector->sector->gravity : 1.0f;
```
electricbrass: I would probably use level.gravity instead of 1.0f. Standard gravity is much smaller than 1.

common/p_mobj.cpp
```diff
@@ -1378,7 +1373,7 @@ static void P_ApplyXYFriction(AActor* mo)
 //
 void P_XYMovement(AActor *mo)
 {
-	if (!mo || !mo->subsector)
+	if (not (mo && mo->subsector && mo->subsector->sector))
```
electricbrass: Again, no need for extra sector check

electricbrass (changes requested):

CurrentResident:
Thanks for the thorough review.  I had the feeling somehow that I was being too..  clumsy, I guess, going through thinking, "is a pointer about to be dereferenced? add a check," without looking at broader context.  Fixes incoming.

common/p_enemy.cpp
CurrentResident: There's plenty of places where I could make similar ptr -> ref changes, and I wasn't sure if you'd be thrilled with me doing that.  Arguably out of scope, but also, if we're taking the policy of cleaning these kinds of things up as we go, I'm happy to do it.

common/p_map.cpp
CurrentResident: Alright, I didn't realize that - there's a number of added checks that I'll update then.

common/p_mobj.cpp
CurrentResident: Okay, I'm confused.  In `sector_t`'s declaration, we have this:

```c++
	float gravity = 0.0f;		// [RH] Sector gravity (1.0 is normal)
```

Does the above comment need to be corrected/updated?

common/p_enemy.cpp
electricbrass: I've generally been updating these things when possible, especially in spots like this one where it can be done easily without affecting anything but the one function + the single spot it gets called.

common/p_mobj.cpp
electricbrass: Hm, good question. I also think I was mixing it up with aircontrol. But I do see somewhat conflicting sources for what is normal gravity. level.gravity comes from sv_gravity, which is 800 by default.

CurrentResident: Okay, I pushed all my updates, but I'm leaving the `1.0f` at least for now, based on the hardcoded initialization in `P_LoadSectors` in p_setup.cpp, which sets it to `1.0f`, and in use, it *seems* to be used as a scale factor applied to `level.gravity`, though I haven't chased it down fully yet.

electricbrass (approved):

PR #1648, opened 2026-03-15
Enable some extra compiler errors on clang/GCC to prevent common bugs

electricbrass:
This PR makes `-Wold-style-cast`, `-Wclass-memaccess`, and `-Wcast-function-type` into errors. The latter two only had a few instances in the codebase, so it was very small changes to address those, and I thought I might as well enable them. Casting function pointers is almost always undefined behavior, and memsetting non-trivial types is both very easy to avoid and prone to causing issues.

The bulk of this PR is getting everything to compile again after enabling `-Wold-style-cast`. C-style casts have multiple times now resulted in hidden `reinterpret_cast`s on unrelated types that have resulted in crashes or fatal errors, where enforcing the use of C++ casts would have revealed the issue (as long as `static_cast` was tried first, of course). Another benefit is that now we can easily search for all instances of `reinterpret_cast` and better consider whether we need to fix the code using them. With C-style casts, it is both impossible to search for them without actually parsing the code, and difficult to tell when they are acting as a `static_cast`, `const_cast`, or `reinterpret_cast`. There ended up being many many more of these casts than I expected, so the diff is quite large, but I strongly believe C-style casts are *never* the correct option in C++ code, and it's better to do this sooner rather than later, so that we can prevent future issues.

A few miscellaneous compiler warnings were addressed as well.


electricbrass:
Something got messed up with alignment of console text, converting to draft for now.

fuzztooth (approved):
github can't even handle the scrolling of this properly, so I don't think there's going to be any real use of manually reviewing this.

If all the demos pass and it still functions on a smoke test then fine.

Issue #825, opened 2023-01-05
[FEATURE] Enable the 'spectate camera' as a F12 option in appropriate game modes


Acts19quiz:
**Is your feature request related to a problem? Please describe.**
I believe it would be beneficial to analysis of playback of demos and netdemos if one could F12 into the spectate camera and observe action throughout an entire map, particularly to monitor a particular trouble spot or to monitor the movements of a particular object, enemy, etc.

I also believe there would be entertainment value in enabling the 'spectate camera' as a F12 option, not just in demo and netdemo playback, but also modes like Horde, Cooperative, and any other modes where spectate spying is appropriate. Currently, the only way to do this is actually after first connecting to a server, but before joining the game. At this point, enable the spectate viewpoint in F12 only if you're dead.

**Describe the solution you'd like to see**
Adding the 'spectate camera' in F12's cycle of viewpoints in Horde, Cooperative, and other game modes where global spectate is already enabled or considered appropriate. Preferably, only after you're dead so as to prevent odd situations where you can view the entire map with yourself still in it, potentially giving some kind of weird advantage.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rakohus:
Kind of linked to #317

Acts19quiz:
Duplicate issue. O: I searched, but I guess I didn't search for the exact right keywords.

Issue #317, opened 2021-04-25
[FEATURE] Allow free roaming spectator when dead/in demos


Xenaero:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like to see**
When viewing a demo or when knocked out a round ala LMS/TLMS, allow the dead/demo-viewing player to free-roam in a flying spectator state.

**Describe alternatives you've considered**
None, gimme the wings!!!!!

**Additional context**
It is nice to view demos in a third person view at some points, as well as fly around a level and view the rest of the round in LMS/TLMS when out of lives.

segovia-no:
Im also looking forward for this feature, it would be nice to get some free roam to edit videos

rakohus:
Implemented and should be in 13.0.0. thanks @endless-r0ad

PR #1668, opened 2026-03-25
[FEATURE] Clientside Freecam for netdemos and survival gamemodes

endless-r0ad:
Been cooking on this for some time now, and testing during horde survival sessions, I think it's ready for a PR. Please let me know if you have any questions, concerns, feedback, etc - Ill give a general rundown here

Overview:
- The freecam is a clientside-only player_t with id 255 that is added during G_InitNew, if we are in a netdemo or survival gamemode
- The freecam is in the rotation for spynext/spyprev
- The consoleplayer's commands are copied to the displayplayer if the displayplayer is the freecam (allow control)
- The freecam position is saved for rewind/fastforward in demos, and will reset on level/wad change
- The freecam is usable in survival only when you run out of lives
- The majority of code paths are handled by the freecam being a spectator as well as ingame() equating to false. Anything else that was needed (like allow mouse movement) checks the new bool isFreecam (default false)

Confirmed during testing:
- Odasrv is never told about the freecam player and never sees it
- The freecam player is never serialized (meaning when you are in survival game and recording a netdemo - the freecam player is not written to that netdemo)
- The clientside freecam will retire and become unusable if a real 255th player joins the game/netdemo (tested this by setting freecamplayer_id to 2 instead of 255)

Note I started another implementation where freecam was it's own real object (not static), and both Freecam and player_t inherited from a base class. However that was quickly scrapped because I realized the codebase is chained to player_t and any attempt to untangle would require a herculean rewrite. - that being said I think this works quite well!

Linked to #317 and #825

https://github.com/user-attachments/assets/bc8dcc4e-aa38-45db-b42c-a6eef5631a24


electricbrass:
> * The freecam is a clientside-only player_t with id 255 that is added during G_InitNew, if we are in a netdemo or survival gamemode

Does this limit players to 254? Or cause unexpected issues when 255 players are present? It seems like this should just use the client's existing player_t

Edit: Missed this:
> The clientside freecam will retire and become unusable if a real 255th player joins the game/netdemo (tested this by setting freecamplayer_id to 2 instead of 255)

Definitely seems weird to have a feature that suddenly disappears at max players.

client/src/cl_freecam.cpp
```diff
@@ -0,0 +1,136 @@
+// Emacs style mode select   -*- C++ -*-
+//-----------------------------------------------------------------------------
+//
+// $Id$
+//
+// Copyright (C) 1998-2006 by Randy Heit (ZDoom).
+// Copyright (C) 2006-2026 by The Odamex Team.
```
CurrentResident: Since this file is new as of this year, copyright should only be asserted for 2026.

client/src/cl_freecam.h
```diff
@@ -0,0 +1,54 @@
+// Emacs style mode select   -*- C++ -*-
+//-----------------------------------------------------------------------------
+//
+// $Id$
+//
+// Copyright (C) 1998-2006 by Randy Heit (ZDoom).
+// Copyright (C) 2006-2026 by The Odamex Team.
```
CurrentResident: Same comment about copyright year.

client/src/cl_freecam.h
```diff
@@ -0,0 +1,54 @@
...
+
+private:
+	static fixed_t x;
+	static fixed_t y;
+	static fixed_t z;
+	static angle_t angle;
+	static fixed_t pitch;
+	static bool wipedOnLevelChange(player_t* cam);
+	static void buildCam(player_t* p_cam);
+};
```
CurrentResident, line 44: By having everything be a static member, there really isn't any use in declaring this to be a class.  Consider either making this a proper singleton class with real members, or just making it a namespace with its data declared in the .cpp file.

client/src/cl_freecam.h
```diff
@@ -0,0 +1,54 @@
...
+	static bool allowAdd();
+	static bool allowSpy();
+	static void retireFor255thPlayer(player_t* cam);
+
+private:
+	static fixed_t x;
+	static fixed_t y;
+	static fixed_t z;
+	static angle_t angle;
+	static fixed_t pitch;
```
CurrentResident: If you decide to make this a real class, consider prefixing the data members with something that makes it clear in the .cpp file that we're looking at class-scoped members.

client/src/cl_game.cpp
```diff
@@ -693,7 +694,7 @@ void G_AddViewPitch(int pitch)
 		pitch = -pitch;
 
 	if ((Actions[ACTION_MLOOK]) || (cl_mouselook && sv_freelook) ||
-	    consoleplayer().spectator)
+	    consoleplayer().spectator || displayplayer().isFreecam)
```
CurrentResident, line 698: Probably a naive question (and not strictly related to this particular SLOC), but taking a step back, how does freecam differ from the existing spectator feature?  Could we achieve the same thing by expanding the spectator feature instead of having both a "spectator" and a "freecam?"

If not, consider adding comments to the cl_freecam.h/.cpp explaining why it was needed / why spectator alone isn't sufficient.

common/p_mobj.cpp
```diff
@@ -2778,6 +2778,18 @@ int P_IsPickupableThing(short type)
...
+// P_IsSpawnThing
+//
+// Returns true if the thing type is a spawn
+//
+bool P_IsSpawnThing(short type)
+{
+	return (type == 1 || type == 2 || type == 3 || type == 4 ||  // player1-4
+			type == 11 ||										 // DM
+			type == 5080 || type == 5081 || type == 5083);		 // Team
+}
```
CurrentResident: Where did these magic numbers come from?  (I presume it's a wad standard..)  Is there a type, enumeration, or set of macros that we could use instead?  If not, consider defining them somewhere.

If we do go ahead with defining them, we should probably disambiguate names a little bit.  I found it confusing at first given that there is a Mobj "type" field, which is definitely not the "type" being looked at here.

CurrentResident (changes requested):

common/p_mobj.cpp
electricbrass: These are the ednums for player starts. This is the `type` for MapThings, which are the actual objects placed in the map, rather than the `type` of the mobjs spawned by them.

client/src/cl_game.cpp
electricbrass: Agreed, this seems like reusing the existing `player_t` makes more sense than using a dummy freecam player, and basing that off existing spectator functionality.

endless-r0ad:
> > * The freecam is a clientside-only player_t with id 255 that is added during G_InitNew, if we are in a netdemo or survival gamemode
> 
> Does this limit players to 254? Or cause unexpected issues when 255 players are present? It seems like this should just use the client's existing player_t
> 
> Edit: Missed this:
> 
> > The clientside freecam will retire and become unusable if a real 255th player joins the game/netdemo (tested this by setting freecamplayer_id to 2 instead of 255)
> 
> Definitely seems weird to have a feature that suddenly disappears at max players.

Agree it is a drawback of this implementation.  I guess I just chose the easiest solution which to me was just removing the functionality. The freecam could have id 256 and keep MAXPLAYERS at 255 - that would then require the player ids to be an int16 or something (not sure how much of an overhaul that would be). But using the client's existing player_t is not possible (at least in netdemos) because the consoleplayer commands are ultimately overwritten by netdemo.readMessages.

common/p_mobj.cpp
electricbrass: Additionally, the team starts should probably be gotten from `TeamInfo`'s `TeamSpawnThingNum`

electricbrass:
> But using the client's existing player_t is not possible (at least in netdemos) because the consoleplayer commands are ultimately overwritten by netdemo.readMessages.

Ah yea, I was just thinking about multiplayer and not netdemos. That makes sense.

client/src/cl_game.cpp
endless-r0ad: The point of the freecam is for the client to have another object in the game (besides the consoleplayer) that can be controlled. In netdemos, the consoleplayer cannot be controlled because it's commands are overwritten by netdemo.readMessages(), so something needs to be added. 

I am not sure what you are getting at because my implementation already has expanded on the spectator feature (freecam player has spectator set to true so that it is invisible and unhittable to other players).  The expansion is the isFreecam boolean that is needed in order to control another object that is not the consoleplayer.

CurrentResident: Yeah, I think the missing link for me was realizing the control of `displayplayer` vs `consoleplayer`.  Thanks!

client/src/cl_freecam.h
endless-r0ad: Update to a namespace in latest commit

common/p_mobj.cpp
endless-r0ad: Most of these I took from what I read in the P_SpawnMapThing() function. There is no enum for the type field on mapthing2_t and not sure if it's possible since the THINGS lump seems to be free use for mappers and modders. If you read through P_SpawnMapThing you will see stuff like "musicchanger thing" which goes all the way to type 14164.  I can define the spawns somewhere similar to what is done in P_IsHordeThing, but this seems like the start of a refactor/regression issue.

Also for TeamSpawnThingNum they are hardcoded in InitTeamInfo() for team red, blue, and green. I think they can only ever be 5080, 5081, or 5083

electricbrass: Yea, for the coop/dm starts I would just leave them as is.

> Also for TeamSpawnThingNum they are hardcoded in InitTeamInfo() for team red, blue, and green. I think they can only ever be 5080, 5081, or 5083

Yes, and theres a few other places that just hardcode the numbers. It will work correctly, *but* iterating over the teams and getting the types from there will make it future proof for if we ever add real teaminfo support. A case of old code can wait to be changed, but might as well write new code to consider it.

endless-r0ad: Ok got it, didn't realize those could potentially change in the future - added in latest commit

CurrentResident (approved):
Awesome!

endless-r0ad:
There was a jarring viewz bob motion that was happening for a few frames on fast forward or rewind - the issue was hidden to me by the screen wipe effect. This last commit resolves that

endless-r0ad:
In case it flew under the radar, it was just a few extra lines to allow using the freecam while a netdemo is paused, at commit 7fe5455. This last commit fixes interpolation since it was only working at 35fps before - now everything is interpolated except things/sprites when paused


https://github.com/user-attachments/assets/6632c1e0-9115-4a47-827f-7f8ab91493ec

bcahue (approved):

PR #1669, opened 2026-03-25
Use std::chrono for banlist, remove strptime and timegm

electricbrass:
C++20 finally provides a way to handle everything needed for this with the STL instead of relying on POSIX extensions to the C library, letting us drop a couple small dependencies that provided Windows implementations for `strptime` and `timegm`.


electricbrass:
...apparently macOS is still missing `std::chrono::parse`...

electricbrass:
:/ guess this doesn't work on almost anything yet...

PR #1671, opened 2026-03-29
Replace `std::enable_if` with `requires` and clean up comparison operators with `<=>`

electricbrass:
Just a bit of minor cleanup


electricbrass:
Also fixes a couple unsafe loops that invalidated iterators by using std::erase_if

electricbrass:
This ended up revealing an issue where std::strings were being implicitly converted to OStrings, so now OString constructors are explicit, since we pretty much only ever want OStrings explicitly.

fuzztooth (approved):

Issue #1675, opened 2026-03-31
[BUG] netdemos will pause without pressing pause button


endless-r0ad:
**Describe the bug**
Netdemos will pause without hitting the pause button

**Build that the bug occurred in**
12.1.0 win 64

**To Reproduce**
- netplay any demo
- pause the demo
- use netff (right arrow) to increment by 1 tic
- unpause the demo
- rewind with netrew
- let demo play
- the demo will pause again without pressing pause

**Expected behavior**
The demo should not pause again after rewinding in this scenario

PR #1677, opened 2026-04-02
Fix demo pausing without pressing pause

endless-r0ad:
Fix for the issue in #1675

### DIFF
--- a/client/src/cl_demo.cpp
+++ b/client/src/cl_demo.cpp
@@ -781,7 +781,7 @@ void NetDemo::ticker()
 	netdemotic++;
 	if (netdemotic == pause_netdemotic)
 	{
-		pause_netdemotic = netdemotic - 1;
+		pause_netdemotic = 0;
 		pause();
 		::paused = true;
 	}


electricbrass:
I think that was there intentionally, as a way to (sort of) allow rewinding by a single tic. It seems like maybe it would be better if the call to `resume` in `NetDemo::nextTic` were replaced with just doing `state = oldstate;` and clearing `pause_netdemotic` to 0 were added to `NetDemo::resume`. Or maybe adding an argument to resume to specify whether to clear it instead of removing that call to resume?

endless-r0ad:
Feel free to make additional changes but the only thing this line is doing is causing the bug.

electricbrass:
> Feel free to make additional changes but the only thing this line is doing is causing the bug.

That line makes it so that rewinding while paused will load the last snapshot and then play up until 1 tic before where it was paused.

endless-r0ad:
> > Feel free to make additional changes but the only thing this line is doing is causing the bug.
> 
> That line makes it so that rewinding while paused will load the last snapshot and then play up until 1 tic before where it was paused.

Wait but rewinding while paused does nothing. The only way the subsequent pause happens is outlined in #1675. Not sure how that is not a bug but please close if its the expected behavior.

electricbrass:
> Wait but rewinding while paused does nothing. The only way the subsequent pause happens is outlined in #1675. Not sure how that is not a bug but please close if its the expected behavior.

After investigating a bit more, it looks like that was the *intention* but it was never actually implemented properly, so there's actually a pair of bugs here. I've made a new PR that should address your issue as well as fixing rewinding while paused.

PR #1678, opened 2026-04-05
[BUGFIX] Netdemo rewind by 1 tic works when playing instead of when paused

### LINKED ISSUE #1675: [BUG] netdemos will pause without pressing pause button

electricbrass:
Addresses #1675
Supercedes #1677 

This PR fixes the rewind functionality of netdemos so that the rewind and then play back up until the previous tic behavior now only happens when a netdemo is paused.


libraries/fltk
CurrentResident, line 1: Did you intend to change the fltk submodule pointer in this PR?

client/src/cl_demo.cpp
```diff
@@ -1224,7 +1225,24 @@ void NetDemo::nextTic()
...
+//		and replay from there.
+//
+void NetDemo::prevTic()
+{
+	if (!isPaused())
+		return;
+
+	pause_netdemotic = netdemotic - 1;
+	state = oldstate;
 	::paused = false;
```
CurrentResident, line 1246: The main constraint of this function is expressed as a negative-logic early return, which is tough to read from a mental model / natural language standpoint.  To me, it has a similar awkwardness as saying "here's how we do this thing, except if we're not paused."

Consider just having it be:

```c++
if (isPaused())
{
    stuff;
}
```

client/src/cl_demo.cpp
```diff
@@ -1224,7 +1225,24 @@ void NetDemo::nextTic()
...
+	::paused = false;
+}
+
+//
+// prevTic()
+//
+//		Rewind to the previous gametic.
+//		It has to rewind to the last snapshot
+//		and replay from there.
+//
```
CurrentResident, line 1238: I totally agree with that description on how to get an accurate previous-tic state, but I don't see how this code results in that behavior.  It looks like `prevTic` just resumes the demo and leaves it to the user to know to command a rewind to the previous snapshot and then let it play until it auto-pauses at `pause_netdemotic`.

So because it's still playing under user control and then pausing without the user hitting pause, the bug as written isn't fully resolved.  Perhaps we should consider fully backing up to the previous snapshot and then doing a (muted) faster-than-realtime fast forward to the pause point?

libraries/fltk
electricbrass: Nope!

client/src/cl_demo.cpp
electricbrass: > It looks like `prevTic` just resumes the demo and leaves it to the user to know to command a rewind to the previous snapshot and then let it play until it auto-pauses at `pause_netdemotic`.

Rewind command is what calls this function in the first place.

> Perhaps we should consider fully backing up to the previous snapshot and then doing a (muted) faster-than-realtime fast forward to the pause point?

I agree that would be better, my goal here was minimal changes to get what was, as far as I can tell, the original intended behavior.

client/src/cl_demo.cpp
CurrentResident: BTW, please feel free to push back if you don't agree.  One totally valid reason to not change this would be because it would then be inconsistent with the pattern used in other functions in this class, and consistency is certainly important, but is it really worth the effort to change all instances of it in this class?

electricbrass: Yes, I did it that way both for consistency with the the rest of the class, and because I personally find it more readable than the alternative.

CurrentResident: Well, different strokes I suppose.  Consider this resolved!

client/src/cl_demo.cpp
CurrentResident: > Rewind command is what calls this function in the first place.

Ach, I missed that.  Thanks for clarifying.

> I agree that would be better, my goal here was minimal changes to get what was, as far as I can tell, the original intended behavior.

I see, and I think it achieves that well.  My comment is more about how Github seems to be saying that issue https://github.com/odamex/odamex/issues/1675 may be auto-closed by merging this PR, which in my read, would be premature without buy-in from @endless-r0ad as the author of the bug report, because after this is merged, the duplication steps still result in what appears to be an unprompted pause.

Now, I could be misinterpreting what github is saying and will do, and the issue won't actually be auto-closed, but if it would be, then I do think it's worth a chat with @endless-r0ad first.

endless-r0ad: Fine with me if it's the intended behavior

CurrentResident: Okee doke, consider my comment resolved!

CurrentResident:
BTW, for whatever reason, github isn't letting me mark the comments resolved, not sure why.

bcahue (approved):

PR #1679, opened 2026-04-06
Weapon state reconcile - part 1

CurrentResident:
### Overview

This is the first part of a rapid weapon desync rollback feature.  It is written with the assumption that desyncs are inevitable given that the client-side uses dead-reckoning prediction, therefore the quickest recovery is to rollback using canonical truth data messages from the server and adjust history and current state accordingly.

### Details

This first pass implements rollback for ammo quantities, base weapon availability and selection.  Weapon selection rollback is only *partial* at this time because psprite states are not yet rolled-back, which drives fine-grained scheduling of weapon state, so weapon selection desyncs still result in obstructive, buggy behavior.

Rollbacks are recorded and replayed in netdemos to accurately reconstruct the client state.  This required adding the rollback state to the netdemo snapshots.  `FArchive` was updated to handle `std::array`s and generic integers.

Additionally, in order to fix erratic rollbacks in `weaponowned` state, `wp_none` was adjusted to be a proper enumeral of the `weapontype_t` enumeration, along with accompanying fixes to array handling and treatment of the array's `NUMWEAPONS` sizer.

### Next steps

- Update netdemo recording capture the PlayerInput messages rather than the ticcmd messages, allowing clear analysis of messages that contain inventory requests
- Server-initiated inventory messages following weapon and ammo pickups
- Client-initiated inventory messages following weapon pickups
- psprite rollback

### Alternatives considered

The initial implementation used a separate message for client-initiated inventory requests.  This was discarded because testing revealed that the server's inventory messages are only fully correct as query responses at the beginning of the tic that follows the tic being queried - i.e. the tic whose PlayerInput message resulted in a pickup that deserved verification.  Because the server can handle multiple PlayerInput messages per tic under certain conditions, we need to ensure that the command that resulted in a pickup+query does NOT produce results that could be muddied up with another PlayerInput message.  Therefore, the inventory query indicator is best bundled up inside the PlayerInput message, where it will be A)  queued up and processed with the Inputs that resulted in the inventory change, and B) serve as a barrier to avoid conflating multiple Inputs that may or may not produce diverging results.


client/src/PlayerStateRoller.cpp
```diff
@@ -0,0 +1,198 @@
...
+	{
+		m_mostRecentTic = currentTic - 1;
+		m_oldestTic     = currentTic;
+	}
+
+	if (currentTic == ExpectedTic())
+	{
+		m_mostRecentTic = currentTic;
+
+		if (m_history.size() == m_history.bucket_count())
```
CurrentResident, line 34: Add a comment explaining why we do this check.

client/src/PlayerStateRoller.cpp
```diff
@@ -0,0 +1,198 @@
...
+		{
+			Roll(i_oldTic, [&deltaItemData](auto& rollingIter)
+				{
+				    ApplyDeltaArray(rollingIter->second.ammo, deltaItemData.ammo);
+				});
+		}
+
+		if (maxammoRequiresRoll)
+		{
+			Roll(i_oldTic, [&deltaItemData](auto& rollingIter)
```
CurrentResident, line 155: Add comment(s) explaining why these are rolled via delta.

common/farchive.cpp
```diff
@@ -990,12 +936,12 @@ uint32_t FArchive::FindObjectIndex (const DObject *obj) const
 {
 	if(!m_ObjectMap)
 		return ~0;
-	uint32_t index = m_ObjectHash[HashObject (obj)];
+	size_t index = m_ObjectHash[HashObject (obj)];
```
CurrentResident: Oof, this warning "fix" actually causes a crash.

drivingspectrum (approved):

electricbrass:
I assume this should be closed now after #1753?

CurrentResident:
> I assume this should be closed now after #1753?

Oh yeah, this PR is long OBE.

PR #1696, opened 2026-04-18
Fix CMake issues with the new unit tests

chewi:
This fixes the test dependencies, prevents the installation of the internal GoogleTest, and allows the use of an external GoogleTest.

Distributions will want to run these tests but will not want to build GoogleTest yet again. It is widely available and highly compatible.


chewi:
Hmm hold on. Those cmake_dependent_option changes are wrong. Semicolon ANDs the conditions, it doesn't OR them. Giving an entire condition string requires CMake 3.22, so I guess we'll need additional variables. I'll fix that later today.

CMakeLists.txt
```diff
@@ -53,14 +53,15 @@ option(USE_LTO "Build Release builds with Link Time Optimization" 1)
 option(ODAMEX_INSTALL_BINDIR "Smart default for bin dir location" "/usr/local/bin")
 option(ODAMEX_INSTALL_DATADIR "Smart default for data dir location" "/usr/local/share")
 option(USE_EXTERNAL_LIBDWARF "Use external libdwarf for cpptrace" 0)
-cmake_dependent_option( USE_INTERNAL_ZLIB "Use internal zlib" ${USE_INTERNAL_LIBS} BUILD_CLIENT OR BUILD_SERVER 0 )
-cmake_dependent_option( USE_INTERNAL_ZSTD "Use internal zstd" ${USE_INTERNAL_LIBS} BUILD_CLIENT OR BUILD_SERVER 0 )
+cmake_dependent_option( USE_INTERNAL_ZLIB "Use internal zlib" ${USE_INTERNAL_LIBS} "BUILD_CLIENT;BUILD_SERVER;BUILD_TESTS" 0 )
+cmake_dependent_option( USE_INTERNAL_ZSTD "Use internal zstd" ${USE_INTERNAL_LIBS} "BUILD_CLIENT;BUILD_SERVER" 0 )
```
electricbrass: Tests also need zstd, so that one needs to be updated too.

electricbrass:
> Giving an entire condition string requires CMake 3.22

We will likely be increasing the minimum to 3.24 soon anyway.

CMakeLists.txt
chewi: Do they? I can't see any sign of that.

electricbrass: Yes, its a cpptrace dependency. The cpptrace cmake already statically links it into the cpptrace binary, so it doesn't show up in the lists of linked libraries for any of the odamex targets.

chewi:
> We will likely be increasing the minimum to 3.24 soon anyway.

Okay, in that case, I've just pushed it with the new syntax. It seems to work as intended on 4.3.1. I guess there's no harm in merging before bumping the minimum since this was already broken anyway.

CMakeLists.txt
chewi: I see. Fixed!

electricbrass (approved):

Issue #1519, opened 2025-12-27
[BUG] Crash in P_SmartMove when targeting disconnected player


electricbrass:
**Describe the bug**
In Eviternity II map14, a player became a spectator and a few seconds later their client crashed.

According to the crash dump, it happened in P_SmartMove because `target->subsector` was null when calculating `on_lift`. It appears that `actor` was a cyberdemon and `target` was a player. `target->player` was also null.
<img width="788" height="276" alt="Image" src="https://github.com/user-attachments/assets/36e3f34a-acb7-4dee-972e-50b6049a49c3" />

It seems likely that the cyberdemon was still targeting the spectator, but it's not definitely what happened.

**Build that the bug occurred in**
12.0.0

**To Reproduce**
Uncertain, have not been able to reproduce

**Expected behavior**
No crash

**Screenshots, NetDemos, & Crash Dumps**
[odamex_g_41152_20251226T220110.zip](https://github.com/user-attachments/files/24352104/odamex_g_41152_20251226T220110.zip)

**Additional context**

electricbrass:
Revisiting this after #1587, it looks like a player spectating just before crashing was most likely a coincidence. The player being targeted is in an almost entirely destructed state, matching what would be expected of a player actor left behind by a player who disconnected from the server, rather than a spectator.

electricbrass:
It seems like the destructor of AActor setting its `self` `szp` to null is supposed to avoid situations like this, but `AActor::Destroy` only calls the destructor in 2 cases:
1. The refcount was non zero, so it was placed into `LingerDestroy`, and a later `DThinker` to be destroyed cleaned it up later
2. `DObject::Inactive` is set to true, which only occurs when shutting down the application

PR #1701, opened 2026-04-19
[BUGFIX] - Read access error on nullptr within P_SmartMove

endless-r0ad:
This bug was found from one of @nikovaan dmp files from a testing session. The crash happened immediately after a CL_LoadMap or CL_ResetMap. Not sure if more research needs to be done as to why target->subsector was null here. But this would at least fix the read violation

image

### DIFF
--- a/common/p_enemy.cpp
+++ b/common/p_enemy.cpp
@@ -486,7 +486,7 @@ bool P_SmartMove(AActor* actor)
 	int dropoff = 0;
 
 	/* killough 9/12/98: Stay on a lift if target is on one */
-	bool on_lift = co_staylift && target && target->health > 0 &&
+	bool on_lift = co_staylift && target && target->health > 0 && target->subsector &&
 	               target->subsector->sector->tag == actor->subsector->sector->tag &&
 	               P_IsOnLift(actor);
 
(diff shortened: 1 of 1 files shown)


electricbrass:
Here is more about why this happens:
https://github.com/odamex/odamex/issues/1519

electricbrass:
I'll have to take another look at the code but i think this fix will just result in another crash later in the movement code, because its all expecting a valid mobj by this point. We might want to try adding the mo->destroyed check mentioned here? 
Screenshot_20260419-132114

endless-r0ad:
Ok yep totally missed that other issue, will try to look into adding the mo->destroyed check

endless-r0ad:
@nikovaan found a way to reliably reproduce the error on odasrv by joining the game, waiting for survival match to start, have an enemy target you, and then disconnect. Most recent commit I believe is what you also had in mind as well @electricbrass 

image

electricbrass:
> Most recent commit I believe is what you also had in mind as well @electricbrass

Yep, I think this looks good!

electricbrass:
Oh actually, it probably needs to be `if (target && target->WasDestroyed())`, since target could be null

electricbrass (approved):

Issue #1697, opened 2026-04-18
[BUG] Instantly-moving ceilings desync


CurrentResident:
**Describe the bug**

*Please note that the original description of this bug was incorrect*.

Some sectors move instantly by design / special command.  Sectors that do this should only produce a single `SVC_MovingSector` message from the server.  For instantly-moving floors, this works by setting the floor mover object's status to `destroy`, which allows it to continue existing until checked, transmitted, and properly destroyed in `SV_DestroyFinishedMovingSectors`.  This allows the final state of the floor to be transmitted just before the mover is destroyed.

The above function also checks for ceilings.  However, the ceiling mover object does not set status to `destroy` in most cases where motion stops - it simply performs an immediate Destroy action, preventing `SV_DestroyFinishedMovingSectors` from doing its work and preventing the final ceiling state from going out to clients.  For instant-mover ceilings, clients never see the motion.

In the attached example, this causes an "invisible shield" sector to not lift for the clients although it has for the server.  This leads to a startling desync where weapons fired across the sector appear to hang in midair even though the server processes the projectiles freely flying through the sector and ultimately striking obstacles on the other side.

**Build that the bug occurred in**

Protobreak commit 073b30bf50b711c47edf05d8e7283c149de189f7.  [Build](https://github.com/odamex/odamex/actions/runs/23720461265)

**To Reproduce**

- Wad:  2vydotwad.wad
- Map MAP03

1. Start recording a netdemo.
1. Make your way to line 1089, a switch.
2. Notice the two square-shaped platforms with invisible walls, each containing a gang of inactive monsters.
3. Activate the switch.
4. Notice the two groups of monsters go active.
5. Run into one of the platforms.  Notice the view briefly bob like you're no-clipping.
6. Fire rockets or plasma through the boundaries of the platform.
7. Notice the predicted projectiles appear to freeze in the horizontal plane, only to teleport and explode elsewhere as though they had successfully flown through the wall.
8. Replay the netdemo, watch the messages at the moment of activation.  Look for a `SVC_Switch` message for line 948 with no activation and no special value.
9. Also notice the lack of `SVC_MovingSector` messages, which would have triggered the "invisible shield" sectors to lift.

**Expected behavior**

- The invisible shield sectors should lift.
- The projectiles should fly naturally without appearing to stop at the boundary of the platform.

**Screenshots, NetDemos, & Crash Dumps**

Initial report:  https://discord.com/channels/236518337671200768/1495076326204375090/1495077105463988404

[...]

PR #1706, opened 2026-04-20
Fix moving ceiling desync at end-of-motion final state

CurrentResident:
Defer the server-side destruction of ceiling movers until after it's had the chance to send final state to clients.

This updates ceilings to match the logic and approach used for floors, which do not exhibit the bug.

Addresses https://github.com/odamex/odamex/issues/1697

### Testing

1. Launch server and load a map that has an instant-raise invisible shield sector.
2. Connect to server
3. Disable `cl_predictsectors`
4. Go to the invisible shield
5. Trigger the raising of the shield
6. Fire rockets at where the shield should have been.
7. If the bug is in effect, the rocket will exhibit stationary "hang time" at the shield boundary.

### 13.0

For 13.0, protobreak will be updated so that the final-state message will be sent as Reliable.
Between the logic updates to make that happen and the coupling of floor and ceiling mover states into a grand unified `MovingSector` message, the changes are complex enough to be worthy of protocol-breakage.  Therefore addressing reliability is deferred to 13.0

### Before

https://github.com/user-attachments/assets/ad43b8a1-e968-43bb-9526-f3ef9b0d8cd3

### After

https://github.com/user-attachments/assets/32ecde7f-c526-4ba9-bb24-51d654a90d44


electricbrass (approved):
Looks good!

PR #1710, opened 2026-04-21
[BUGFIX] No need to increment gametic if netdemo is paused

### LINKED ISSUE #1516: [BUG] Server time increases for players during netdemo while netdemo is paused
**Describe the bug**
While paused in netdemos, the playtime of players will continue to increase during that time. The times will reset to correct at the start of the next map.

12.0.0 Windows x64 Client

**To Reproduce**
1. Start a netdemo (any)
2. Pause during it
3. Bring up the in-game scoreboard, and observe the playtimes

**Expected behavior**
For the playtimes to pause for players during a netdemo while said netdemo is paused

https://github.com/user-attachments/assets/08be3693-888f-4d16-a03e-56da89e4749a

**Additional context**
N/A

endless-r0ad:
Noticed that gametic was still incrementing while the game is in a paused state. I cant see any reason why this should be happening? For netdemos the gametic is overwritten anyway each tic in netdemo.readMessages(). Fixes #1516

### DIFF
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -627,7 +627,7 @@ void CL_StepTics(unsigned int count)
 
 		G_Ticker ();
 		
-		if (!paused && !netdemo.isPaused())
+		if (!netdemo.isPaused())
 			gametic++;
 
 		if (netdemo.isPlaying() && !netdemo.isPaused())
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -626,7 +626,10 @@ void CL_StepTics(unsigned int count)
 		OInterpolation::getInstance().ticGameInterpolation();
 
 		G_Ticker ();
-		gametic++;
+		
+		if (!paused && !netdemo.isPaused())
+			gametic++;
+
 		if (netdemo.isPlaying() && !netdemo.isPaused())
 			netdemo.ticker();
 	}


electricbrass:
I'm a bit worried by this change. To my knowledge, no other port stops gametic when paused, and this affects so many different things that it's hard to predict exactly what effects this will have. I know at least that some animations will stop (though I suppose that could be considered an improvement to some).

Instead I would probably put the check here:
https://github.com/odamex/odamex/blob/1d1547d0a1cf24324450f42016e0d40ddcbd1e62/client/src/cl_main.cpp#L2121-L2128

Another question then is if this should only affect netdemos or singleplayer pausing as well. Obviously it doesn't make sense to keep incrementing the time while a netdemo is paused, but for singleplayer, it could be argued that paused time is still time in the game, and so the player time should keep incrementing.

electricbrass:
Though of course, how many players actually ever look at the play time on the scoreboard in singleplayer, so maybe it doesn't matter much either way at all.

rakohus:
Does this affect menu and/or console when the gametic is paused?

I don't think this should apply to single player given what vanilla and other ports do. I understand the reasoning behind stopping it during netdemos, but is there a problem being solved?

electricbrass:
> Does this affect menu and/or console when the gametic is paused?

Not as far as I can tell. Other than that any console commands that print messages outside of the console will stay on screen until unpausing.

electricbrass:
Oh should also check how it behaves when moving forward in a netdemo by 1 tic at a time. I think that sets it to unpaused for the duration of that tic and so there shouldn't be any issue?

endless-r0ad:
Ill plan to proceed with just the netdemo change after testing tic fast forward - it seems to work on first test. I had originally added singleplayer pause as well just because it made sense to me (the game is paused why is gametic increasing?) but I understand now there could be a lot of risk involved.

electricbrass (approved):

PR #1711, opened 2026-04-22
Fix:  CPU spin and stop streaming console input when server (or client) is pushed to a terminal background

CurrentResident:
There are two parts to this fix:

1. Ignore `SIGTTIN`.  This signal is sent to background processes by the terminal when input becomes available for reading.  Unfortunately, the input is not actually supplied to the program until it's brought to the foreground, so the default POSIX action is to completely suspend the background process and wait for the user to decide to bring it to the foreground.  Since that suspension is a tremendous disruption to Odamex, we change the action from Stop to Ignore.  When this happens, the program-side input read operations fail, and stdin / std::cin are marked as failed.
2.  Update the input-reader thread's main loop to allow the thread to exit if the file or input stream goes EOF, bad, or failed.  The only exception to this is FIFOs / named pipes, which we intentionally want to allow to be reopened to work with successive opens+writes from a user perspective.

### Testing

On my RHEL 10 box, launched odasrv in a terminal under the various conditions, and verified that the CPU usage remains "normal" in all cases:

- `./odasrv` (input normal)
- `./odasrv` + ctrl-z + `bg`  (input terminates)
- `./odasrv &` (input terminates)
- `./odasrv -confile cmds.txt`  (runs commands from text file, then input terminates)
- `./odasrv -confile ./thinger-fifo`  + `cat version > ./thinger-fifo` in another terminal, along with other commands up to `quit`
- `./odasrv -confile ./thinger-fifo &`  `cat version > ./thinger-fifo` in same terminal, along with other commands up to `quit`
- `./odasrv -confile ./thinger-fifo` + ctrl-z + `bg`  (input works via `cat` to fifo as shown above)

### DIFF
--- a/common/m_consolecommandstream.cpp
+++ b/common/m_consolecommandstream.cpp
@@ -23,6 +23,7 @@
 #include "m_consolecommandstream.h"
 
 #include <condition_variable>
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <memory>
@@ -43,7 +44,6 @@
 
 #elif defined UNIX
 
-#include <filesystem>
 #include <signal.h>
 
 #endif
--- a/common/m_consolecommandstream.cpp
+++ b/common/m_consolecommandstream.cpp
@@ -41,6 +41,11 @@
 //  destructed.  On Debug builds, this raises an Abort signal.  By switching to
 //  SRW Locks, which don't even have a destructor, we can destruct at any time.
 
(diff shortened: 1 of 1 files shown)


CurrentResident:
Well darn it, I assumed that because Windows has named pipes that they'd support `is_fifo`...

TIL.

bcahue (approved):

Issue #1589, opened 2026-02-04
[FEATURE] Colorized server console


electricbrass:
**Is your feature request related to a problem? Please describe.**
The client console helpfully color codes everything, making it easier to see at a glance what's going on. The server console on the other hand is just a wall of monochrome text.

**Describe the solution you'd like to see**
Instead of stripping color codes for serverside printing and ignoring print levels, convert the color codes to ANSI escapes.

**Additional context**
We probably want a way for users to disable the color, as well as disabling it automatically when running on older Windows versions (support on Windows is only on 10 or later) or when piping server output.

electricbrass:
https://fmt.dev/12.0/api/#terminal-colors-and-text-styles

{fmt} has support for colors.

electricbrass:
Most of what I wanted is in #1716. I'll leave this open for the remaining missing parts.

There are 2 remaining improvements:
1. Converting ZDoom color codes to ANSI escapes.
2. Only coloring the line starting *after* the timestamp.

PR #1716, opened 2026-04-26
[FEATURE] Colored server console output

electricbrass:
This is a minimal version of what was requested by #1589

Server console output is now colored based on print level, similarly to the client console. In this PR, only the print level is considered, and the entire line, including the timestamp, gets that color. Odamex's ZDoom color escapes still get stripped before serverside printing, so there isn't yet any finer grain control over coloring only part of a line.
image

Color printing is enabled by default and is controlled by the new `log_color` cvar. Setting this to 0 disables the color, and setting it to 2 enables color for the log file as well. 

Color is automatically disabled when:
- Output is being piped or redirected to a file
- Windows only:
  - Running on an older Windows version before support for ANSI color escapes was implemented
- Unix only:
  - The `NO_COLOR` environment variable is set
  - The `TERM` environment variable is not set or is set to `dumb`

Two additional changes have been made as well:
1. Disbling quick edit mode with `SetConsoleMode` now is only attempted if stdin is an actual console. This means Odasrv no longer fails to start on Windows when trying to pipe the output. This has been broken since 10.4.0.
2. Odasrv sets the window title on non-Windows systems now. This was previously Windows only.

Before:
image
After:
image

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -3,3 +3,6 @@
 
 # ACS formatting changes
 caa159e47d0a4fc77c63aa4fd141b7b0946439e6
+
+# server i_main indentation update
+4b0fa19800182b0e94d01ca98ad2781aa0f368ea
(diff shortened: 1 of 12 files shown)


electricbrass:
Also removed a bit of dead code that's been unused and mostly useless for years.

bcahue (approved):

PR #1722, opened 2026-04-27
Segregate wad creation and dependent copy

CurrentResident:
### Summary

Rather than have the one odawad target that we conditionally update to copy, we now have three targets: odawad to create it, and one per executable to copy from the generated file to the executable target directory.  This allows for both per-executable post-build copying and wad rebuild AND copy if any of the wad's actual sources changed.

### Testing

1. Generate and build all executables, verify wad is copied.
2. Generate all, build only one executable, verify wad is copied.
3. Generate one, build it, verify wad is copied.
4. After executables are built, touch wad/wadinfo.txt, run build, verify wad is rapidly generated and then copied to the executable directories for the configured executables.


electricbrass (approved):
Looks good to me

PR #1724, opened 2026-04-27
[BUGFIX] Something for friendlies?

electricbrass:
I'll be honest, I haven't looked enough to see what this does other than verifying that it only affects friendlies. It's a couple lines from MBF that we're missing. We have the rest of the code a bit later down in A_Chase, which never gets run without adding these lines.


bcahue (approved):
None of our demos break due to this, I say let it ride

PR #1728, opened 2026-04-28
[BUGFIX] Odalaunch update checker

electricbrass:
The automatic update checks in odalaunch have been disabled for the last 7 years because of the website moving to HTTPS. This PR restores that functionality using the wxWidgets' newer webrequest api, which handles HTTPS properly. It has also been change to read the latest releases from https://api.github.com/repos/odamex/odamex/releases/latest.

Builds off of some changes from #1725


electricbrass:
Hm, looks like this requires a wxWidgets newer than what's being used on our Fedora 28 and Ubuntu 20.04 builds. It requires at least 3.1.5. This might have to wait for now...

electricbrass:
The update checker is now disabled when using a version of wxWidgets less than 3.1.5. In that case, it matches the current behavior: the checkbox to enable/disable automatic update checking is not shown, and clicking the check for updates button opens the Odamex website.

bcahue (approved):

Issue #1729, opened 2026-04-28
[CHANGE] Allow disabling sprees/multikills for singleplayer while keeping them enabled in multiplayer and allow disabling on a server


electricbrass:
**What is the existing behavior? Please describe.**
1. Right now, the only control for sprees and multikills is a pair of clientside settings that controls whether they are displayed. This means if one does not like them in singleplayer, but does in multiplayer, they have to change settings each time they switch between them.

2. A server has no way to disable sprees and multikills, even if it is trying to be configured for a more vanilla experience. I would really like to be able to disable these on my servers, especially because I feel the current implementation for co-op is not very good (see #1730).

**What is the desired behavior? Why are you requesting the change?**
Add separate settings for displaying sprees/multikills and actually tracking them. The former could be the same as the current settings, while a pair of new cvars could be introduced for the latter. Clientside, the new cvars could control sprees/multikills in singleplayer, and serverside they would obviously control multiplayer.

**If applicable, describe alternatives you've considered**
1. Always automatically disabling them in singleplayer.

2. None

**Additional context**
I hadn't paid much attention to this feature while it was in development. IMO these should have been there before this was ever in a release.

electricbrass:
Having these enabled by default also feels like it goes against Odamex's usual vanilla-by-default stance, yet only changing the default would mean completely missing out on a cool exclusive feature for those who don't know the setting exists, due to the lack of finer control.

bcahue:
Indeed this should be a client side variable (`cl_showofflinesprees`), and actually was intended to be included in the initial implementation, and this was also intended to be disabled by default (meaning, sprees and multi kills are disabled for single player and vanilla demos)

PR #1738, opened 2026-05-16
[CHANGE] Add more Spree and Multi Kill controls

bcahue:
Added 4 vars, 2 clientside and 2 serverside:

### Client
`cl_showofflinesprees` - Show sprees in non-network games, like in single player or vanilla demos. Is superseded by `cl_showsprees`. Defaults to 0.

`cl_showofflinemultikills` - Shows multi kills in non-network games, like in single player or vanilla demos. In fact, multi kills aren't really viewable outside of vanilla demos due to Odamex's inability to play deathmatch offline. Is superseded by `cl_showmultikills`. Defaults to 0.

### Server
`sv_showsprees` - Configures the server-wide disablement of Sprees. If this is 0, sprees aren't shown to any player, even if the player enabled them in the client -- because not only is it disabled in the client, sprees aren't broadcasted at all. Defaults to 0.

`sv_showmultikills` - Configures the server-wide disablement of Multi Kills. If this is 0, multi kills aren't shown to any player, even if the player enabled them in the client. Defaults to 0.

Addresses #1729

### DIFF
--- a/client/src/cl_cvarlist.cpp
+++ b/client/src/cl_cvarlist.cpp
@@ -448,6 +448,10 @@ CVAR(cl_showsprees, "1", "Show killing sprees for the display player.", CVARTYPE
 
 CVAR(cl_showmultikills, "1", "Show multi kills for the display player.", CVARTYPE_BOOL, CVAR_CLIENTARCHIVE)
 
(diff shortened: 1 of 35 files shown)


electricbrass:
The `sv_` variables should probably be added to the sample server configs. Also should they be on by default? That seems the opposite of most of Odamex's vanilla by default settings.

bcahue:
Done, I set the `sv_showsprees` and `sv_showmultikills` as defaulting to 0, and also added them to the `/config-samples` directory, enabled if the config is "new-school" or has a modern game mode (ctf, tdm, lms). Coop/horde is always 0 in these configs.

electricbrass (approved):

PR #1747, opened 2026-06-06
Add "auto" option to hud_scaletext

TehRealSalt:
Lets you set `hud_scaletext 0` to automatically scale chat & mid-prints to the screen resolution, similar to how `con_scaletext` works. Just found myself wanting this, so I poked around to figure out how to do it.

image
(Using "auto" on a resolution that would give me x5, if it were otherwise allowed.)

### DIFF
--- a/client/src/cl_cvarlist.cpp
+++ b/client/src/cl_cvarlist.cpp
@@ -576,8 +576,8 @@ CVAR(			hud_scale, "1", "HUD scaling",
 CVAR(			hud_scalescoreboard, "0", "Scoreboard scaling",
 				CVARTYPE_FLOAT, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE)
 
-CVAR_RANGE(		hud_scaletext, "2", "Scaling multiplier for chat and midprint",
-                CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 1.0f, 4.0f)
+CVAR_RANGE(		hud_scaletext, "2", "Scaling multiplier for chat and midprint (0 = auto).",
+                CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 0.0f, 4.0f)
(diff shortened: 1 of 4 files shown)


client/src/cl_cvarlist.cpp
```diff
@@ -576,8 +576,8 @@ CVAR(			hud_scale, "1", "HUD scaling",
 CVAR(			hud_scalescoreboard, "0", "Scoreboard scaling",
 				CVARTYPE_FLOAT, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE)
 
-CVAR_RANGE(		hud_scaletext, "2", "Scaling multiplier for chat and midprint",
-                CVARTYPE_BYTE, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 1.0f, 4.0f)
+CVAR_RANGE(		hud_scaletext, "2", "Scaling multiplier for chat and midprint (0 = auto).",
+                CVARTYPE_INT, CVAR_CLIENTARCHIVE | CVAR_NOENABLEDISABLE, 0.0f, 4.0f)
```
electricbrass: Why was this changed from byte to int?

electricbrass:
> (Using "auto" on a resolution that would give me x5, if it were otherwise allowed.)

Also, if 5x or higher is desired, the max could probably be raised. I'd guess it's been a while since it was put at 4x and that that might not be enough for some of today's monitor resolutions.

client/src/cl_cvarlist.cpp
TehRealSalt: I was comparing this & `con_scaletext`'s definition, didn't really mean to commit it. That one uses int, so I was seeing if it made any observable difference. I'll amend to undo that

electricbrass: I'm not sure that it actually makes any difference anyway since these are just clientside. It's mostly (entirely?) just to indicate that it can be safely truncated to a byte when transmitted to/from the server.

electricbrass (approved):

PR #1750, opened 2026-06-08
[FEATURE] Bash completion

electricbrass:
I was making a bash completion script for myself and realized that other people might want it too. Unfortunately, there's not a very good way to export the completion from the flatpak (it would be a ridiculously easy way to escape the sandbox), but hopefully somebody finds this useful.

There might be a powershell one coming sometime too ;)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,9 @@ endif()
 # WAD building
 add_subdirectory(wad)
 
+# Bash completion
+add_subdirectory(completion)
+
 # Disable the ag-odalaunch target completely: -DNO_AG-ODALAUNCH_TARGET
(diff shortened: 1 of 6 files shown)


electricbrass:
Oh also, this is fully standalone and does not depend on anything from https://github.com/scop/bash-completion, just to be a bit more portable.

electricbrass:
Decided to add an extra command to the flatpak wrapper script that writes out the completions + some wrapper functions (needed to make the completions work correctly, plus nicer to use anyway) to stdout like a number of projects have been doing recently. Now flatpak users can opt-in to the completions more easily if they want.

bcahue (approved):

PR #1751, opened 2026-06-09
Default build and launch templates for vscode people

endless-r0ad:
These are my profiles that I always copy over to a new branch if I am using vscode. Figured it would be convenient to store these in the source for vscode users. Includes a compound to debug the client and server at the same time. It requires the person to have cmake installed but they should already have that if they followed the build instructions

image


endless-r0ad:
Took some time to get debugging to work on my old mac to make these templates comprehensive, the wiki has been updated. I tested all of these templates on mac sonoma and linux. If someone could test on windows that would be appreciated since my laptop broke. Here are the build commands also

image

The build commands require the odamex dependencies to be installed (which is the first thing the user would be doing)
The launch commands require the odamex dependencies and CodeLLDB extension only

electricbrass:
> If someone could test on windows that would be appreciated since my laptop broke.

Looks like LLDB support for windows is still experimental. I might also add " (LLDB)" to the names of the debug tasks. The other issue I see on Windows is that msbuild will put the build executables in different paths from where ninja puts them (it puts different cmake build types in different directories). It probably should be something like ` "program": "${command:cmake.launchTargetDirectory}/odamex.exe"` (and the linux/osx ones would be worth changing the same way) instead.

endless-r0ad:
Will put these examples in the dev wiki

PR #1760, opened 2026-06-16
Fix client/server deadlock when quit is called on MacOS

endless-r0ad:
When quitting the client or server on Mac, the call to join here deadlocks, requiring the user to force quit. I couldn't find a better solution so I rolled it back to 12.1 logic which was detaching the thread


common/m_consolecommandstream.cpp
```diff
@@ -174,21 +174,24 @@ namespace {
 #ifdef _WIN32
 				// Windows makes us ruthlessly kill the thread.
 				if (TerminateThread(m_thread.native_handle(), 0))
+#elif defined(__APPLE__)
+				// join() deadlocks on mac, someone else can figure out why
+				m_thread.detach();
 #else
```
CurrentResident: This change leaves the conditional aspect of the `_WIN32` case dangling, thus can't compile.  Honestly, I wish I hadn't written it that way to begin with.  It's going to be ugly no matter what, it should have been "intuitive" rather than "clever."

Consider changing this to be more like:

```c++
#if defined(__APPLE__)

m_thread.detach();

#elif defined (_WIN32)

if (TerminateThread(...))
{
    m_thread.join();
}
else
{
    m_thread.detach();
}

#else

if (pthread_cancel(...) == 0)
{
    m_thread.join();
}
else
{
    m_thread.detach();
}
#endif
```

Kinda gross, but it's obvious at a glance what's conditional and what isn't.

CurrentResident (changes requested):

common/m_consolecommandstream.cpp
endless-r0ad: Thanks, I totally missed the if case in win32, hopefully it works now

electricbrass (approved):

PR #1761, opened 2026-06-17
Friendlies optimization

CurrentResident:
This PR adds a static vector of `AActorPtr`s that track the friendly mobjs and avoids doing an expensive search for friendlies if the vector is empty / there are no friendlies.

Additionally, there's a *slight* modernization of `szp` that I started in on back when I was still using an `unordered_set` for this.  Rather than throw it out entirely, I decided to include it here to see if it sticks or faces push-back.  I'm fine either way, but wanted to see if any of it is palatable.


electricbrass:
> Additionally, there's a _slight_ modernization of `szp` that I started in on back when I was still using an `unordered_set` for this. Rather than throw it out entirely, I decided to include it here to see if it sticks or faces push-back. I'm fine either way, but wanted to see if any of it is palatable.

This all looks good to me. I'm all for modernizing things more.

bcahue (changes requested):
What do we think about adding to the friendly list on a new AActor() constructor (where it sets up other information) and removing from the list when an AActor is Destroy()'d? We may have to check for health against AActors in the new list -- unless we add another hook in P_KillMobj to detect when an actor is dead and remove them from the list, but then we run into the problem of Arch-Viles reviving -- but it'd make the bookkeeping much easier.

CurrentResident:
> What do we think about adding to the friendly list on a new AActor() constructor (where it sets up other information) and removing from the list when an AActor is Destroy()'d?

I would be all onboard with that if the friendliness of a mobj held constant across its object-level lifetime.  A pretty good middle-ground would be to add access methods for friendliness, and update the list and other info there (i.e. ownership, special visual effects).  Thoughts?

> We may have to check for health against AActors in the new list -- unless we add another hook in P_KillMobj to detect when an actor is dead and remove them from the list, but then we run into the problem of Arch-Viles reviving

It could just be me being unfamiliar with the monster business logic, but shouldn't monsters have to consider their potential target's corpse state and/or health already, and decide what to do without having special considerations in the bookkeeping structures?  I'd rather we not have to adjust the friendly vector when mobjs go back and forth between dead and not-dead - it really should be just a shortcut to the mobjs that have the is-friendly attribute, regardless of anything else.

Regarding Archies resurrecting them, I suggest we update those locations in the code to call the friendly-setter method one way or the other (adding or removing friendliness) and contain the mid-lifetime bookkeeping adjustments to that method.  Or, please let me know if I'm missing the point here.  :)

bcahue:
> I would be all onboard with that if the friendliness of a mobj held constant across its object-level lifetime. A pretty good middle-ground would be to add access methods for friendliness, and update the list and other info there (i.e. ownership, special visual effects). Thoughts?

Yes, this would make it cleaner. I just wanted to imitate the structure of the list that other ports that do MBF friendlies do, by adding to the list as soon as the thinker is minted (P_AddThinker in DSDA/prBoom) and removing/changing if something happens to the thinker that changes its state or friendliness (P_UpdateThinker in P_KillMobj/P_HealChase (arch revive)).

> It could just be me being unfamiliar with the monster business logic, but shouldn't monsters have to consider their potential target's corpse state and/or health already, and decide what to do without having special considerations in the bookkeeping structures? I'd rather we not have to adjust the friendly vector when mobjs go back and forth between dead and not-dead - it really should be just a shortcut to the mobjs that have the is-friendly attribute, regardless of anything else.

Sorry, I should be clear for why I wanted the list like that. In MBF, when a monster dies they're removed from the friendly/hostile mobj list so that the friendly enemy check is as lean as possible and iterates 1 at a time to keep the CPU happy. When an enemy/friendly is found on these lists (targeted by something), they are moved to the back of the list so the next check doesn't pick them up. This makes it necessary to update the list when something happens to the mobj that takes it "out of play" or "put into play." For this particular change it isn't needed but I wanted to explain myself for future optimizations.

> Regarding Archies resurrecting them, I suggest we update those locations in the code to call the friendly-setter method one way or the other (adding or removing friendliness) and contain the mid-lifetime bookkeeping adjustments to that method. Or, please let me know if I'm missing the point here. :)

Yep, spot on!

CurrentResident:
Okay, thanks for explaining!  I'm onboard with consolidating the manipulation of the friendlies list to the `AActor` class and simple, straightforward APIs, including a tiny update to `AActor::Destroy`.

If I understand correctly, you're asking that the vector be narrowed in scope down from "friendlies" to "living friendlies."  ~~Would it make sense in that case that we don't consider corpses to be MF_FRIEND at all, and that if the corpse comes to life again, that's when it adopts the friendliness (or lack thereof) of its resurrector, regardless of whatever it was before?~~

Found my answer.  No, we should care whether corpses are MF_FRIEND.

----

Okay, I'll update `AActor` to consolidate the special knowledge about the friendlies vector, but I'll leave the question for what updates we make to the vector's content itself, if any, down the road.

bcahue (approved):

bcahue:
Looks good, thank you!

PR #1774, opened 2026-06-25
Roll NETDEMOVER to 4 and allow developers to attempt loading old version netdemos

CurrentResident:
The update to allow truncated netdemos to be played back also changed the netdemo's header format in an incompatible manner due to fields being relocated to different offsets.  This updates the netdemo code so that it is capable of reading the header of the current format (4) and the previous format (3).

By default, any non-current format is rejected.  It also allows a developer to enable `TRY_LOADING_OLD_NETDEMOS` to proceed with the decode attempt anyway.

### DIFF
--- a/client/src/cl_demo.cpp
+++ b/client/src/cl_demo.cpp
@@ -45,6 +45,9 @@
 EXTERN_CVAR(sv_maxclients)
 EXTERN_CVAR(sv_maxplayers)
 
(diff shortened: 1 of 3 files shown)


bcahue (changes requested):
Can we add to LatestDemoVersion to add the last supported version of NETDEMOVER 3, which might be 12.2.1 but we can change it if we release another since this is protobreak.

CurrentResident:
Sure, will do.

bcahue (approved):

PR #1775, opened 2026-06-25
Don't hit with BFG tracer according to sv_unblockplayers and sv_unblockfriendly

endless-r0ad:
If the BFG ball was allowed to pass through a player because of sv_unblockplayers or sv_unblockfriendly, the tracer should follow the same rules. Current day if the ball passes through a teammate the tracer will still hit and negatively effect their movement, this fixes that


electricbrass (commented):
I would probably just put these in p_local.h instead of a new file.

endless-r0ad:
> I would probably just put these in p_local.h instead of a new file.

Thanks that is way better, added in latest commit

electricbrass (approved):

PR #1781, opened 2026-06-28
[FEATURE] Create "cvardocjson" command

bcahue:
The `cvardocjson` command is a new command for Odamex to print out a json file containing a dictionary of cvars, their names, defaults, min and max's (if a number) help text and flags, if any. It is heavily inspired by the `cvardoc` command, hence its name.

The reason to do this is for external programs to have a dictionary of cvars, for example a future version of Odalaunch, so cvars on servers can have descriptions rather than just the name.

This would help any effort to create a "create server" dialog, and let server hosts make better judgements about what to configure. Or something that'd open an Odamex configuration file, to better know what its reading.

The code is a little funky because `jsoncpp` only seems to allow you to serialize in alphanumeric order only, and to get a custom order that makes a little more sense, you seem to have to craft the file yourself.

I also created a json schema file for this document. It should be updated if the json schema of this command is ever changed.

### DIFF
--- a/common/c_doc.cpp
+++ b/common/c_doc.cpp
@@ -230,32 +230,16 @@ static CvarView GetSortedCvarView()
 }
 
 /**
- * @brief Serialize a single JSON value to a compact, one-line string with no
- *        trailing newline, for embedding in hand-ordered output.
- */
-static std::string JSONScalar(const Json::Value& value)
-{
-	Json::FastWriter writer;
-	writer.omitEndingLineFeed();
-	return writer.write(value);
-}
-
-/**
- * @brief Render cvar information as a JSON object string.
+ * @brief Render cvar information as a JSON object.
  *
- * This is because jsoncpp serializes objects in alphabetical order,
- * so to get a custom order, we must do it by hand.
- *
- * @param cvar Cvar to read.
- * 
- * @param indent Indentation string for the object's braces. Fields are
- * indented one further level (3 spaces).
- * 
- * @return The serialized object, or an empty string for cvars that have no
+ * @param out Output value to write to. Set to null for cvars that have no
  * representable type.
+ * @param cvar Cvar to read.
  */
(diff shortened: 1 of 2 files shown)


electricbrass:
Is there a reason the order of the fields matters? Wouldn't this just be for other programs to consume? With human readable output being for `cvardoc` or being produced by whatever reads this. It seems like there's not much benefit to the extra complexity here, for something that JSON explicitly defines as unordered.

On the other hand, if you really care about the order, I wonder how hard it would be to migrate all the existing json code to nlohmann, which does have `nlohmann::ordered_json`, and is generally much nicer to use.

bcahue:
It was just for the readability of the exported file. It for sure isn't necessary and serializing how `jsoncpp` would like it would save a ton of lines of code, so I can make edits to do that. Migration to `nlohmann` is probably in the cards but I wanted to keep this PR simple. 😛

electricbrass:
> Migration to `nlohmann` is probably in the cards but I wanted to keep this PR simple. 😛

Yea I didn't mean that for this PR, more so just wondering.

electricbrass:
Just had a thought: It feels like number should be integer. Seems a bit odd that number doesn't include float. I know it's consistent with what cvardoc already does, but it think it would be worth changing that too.

bcahue:
The reason I split them is because "integer" and "number" are different types in the [JSON Schema](https://json-schema.org/understanding-json-schema/reference/numeric) where integer is a whole number (or with a fractional 0 like 1.0) and number is a float. I was going for maximum compatibility with the latest JSON schema as possible, and inferring the types as they would be inferred in JSON (when in Rome...).

bcahue:
And after looking at it again it doesn't seem that I actually made them represent JSON types. Whoops. I'll fix that.

electricbrass (approved):

PR #1793, opened 2026-06-29
Add tntem command to server

### LINKED ISSUE #1791: [CHANGE] Allow TNTEM cheat to be used directly from the server in netplay
Current behaviour requires sv_allowcheats to be enabled, a player to rcon login, and then enter `tntem` in console. Not `rcon tntem`, as that will fail.

Allow entering `tntem` directly from the server, or a player with rcon login to do `rcon tntem`.

**If applicable, describe alternatives you've considered**
Having some kind of "kill monsters" command on server that's not tied to the state of sv_cheats would also work. In some cases, this might even be preferrable.

electricbrass:
Fixes #1791

### DIFF
--- a/common/m_cheat.cpp
+++ b/common/m_cheat.cpp
@@ -250,6 +250,13 @@ END_COMMAND(mdk)
 
 #endif
 
+#if SERVER_APP
+BEGIN_COMMAND(tntem)
+	const int kills = P_Massacre();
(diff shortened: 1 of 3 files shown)


bcahue (approved):

bcahue:
I wish tntem was a server command so clients can clean the monsters up with 1 command instead of potentially thousands. Alas, outside the scope for this. Approved!

PR #1797, opened 2026-07-01
Fix incorrect weapon autoswitch from pistol or fist

### LINKED ISSUE #1712: [BUG] Misbehaving weapon autoswitch from pistol or fist due to weapon or ammo pickup
**Describe the bug**

If the player is wielding a pistol or fists (or any weapon that has the `WPF_AUTOSWITCHFROM` flag) and picks up a weapon that is high priority in the PWO, but then picks up a weapon that is lower-priority (but auto-switchable-to) *while the pistol or fist is still being lowered*, then the lower-priority weapon is ultimately brought up.

---

Additionally, there's a problematic condition being checked:

```c++
player.ammo[weaponinfo[player.readyweapon].ammotype] != ammo
```

where the RHS `ammo` is of type `ammotype_t` - the above expression compares an ammo quantity to an ammo type.  In theory this could cause a player running that's around with the pistol out with 0 ammo to not switch when picking up ammo for another weapon that they actually have.

----

One of the 13.0 test builds.  The problem very likely exists in 12.2.

**To Reproduce**

1.  Set PWO so that SSG outranks RL.
2. Ensure PWO is enabled server-side, if using a server.
3. Start a map that spawns SSG and RL nearby.
4. With the pistol out, run and pickup the SSG and RL in rapid succession such that the SSG has not yet started to bring up when the RL is picked up.
5. Observe that the RL comes up.

**Expected behavior**

The SSG should come up.

Forthcoming.

**Additional context**

The bug appears to originate in `P_GiveAmmoAutoSwitch` in **p_interaction.cpp**.  The complex check at the bottom considers neither PWO nor `pendingweapon`.

Consider referring to `P_CheckSwitchWeapon` in **p_pspr.cpp** for how to check preferences and take `pendingweapon` into account.  FWIW, I suspect that just calling `P_CheckSwitchWeapon` would not be adequate here because of the vanilla Doom "always switch" condition that gets checked first.

endless-r0ad:
This fixes #1712  by prioritizing pendingweapon instead of readyweapon, as @CurrentResident  described. This ensures on quick successive weapon pickups starting from pistol or fist, this autoswitch logic only happens on the first weapon pickup (when pendinngweapon is wp_nochange). Any subsequent pickups while the pistol/fist is lowering won't hit this logic and end result will abide by PWO rules. 

Easiest way to test this fix is on map01 with PWO on- start with pistol and run over all 3 weapons on the ledge, confirm the correct weapon comes up according to your PWO settings. 

Also added a fix that lets WPSW_PWO_ALT (attack cancels PWO setting) override the autoswitch logic here. To test this, start on the map01 ledge with the "attack cancels PWO" setting. Start with pistol, hold down fire, run over all 3 weapons - note you will keep firing now whereas in 12.2.1 it will autoswitch


drivingspectrum (dismissed):

drivingspectrum:
Looks good.

Thinking about it now though, maybe "ammotype_t ammo" in the function should be renamed to "ammotype_t ammotype" to make it clearer.

drivingspectrum (approved):

PR #1798, opened 2026-07-01
[FEATURE] Odalaunch upgrades

bcahue:
This started out a while ago with a small scope and gradually grew (as is tradition) and now I think it's enough changes to make a PR. I wanted to give the wx launcher a more premium feel. Here's the summary:

First, the permanent player/server variable display is replaced by popovers, if the system supports it (i.e. if wx was compiled with wxUSE_POPUPWIN which is the default in setup.h). If the system does not support it, it falls back on the older style player/server variable display.

Popovers are an at-a-glance view of server variables that replace the verbose permanent lists of players/variables before them. How they work is, when you mouse over either a server's `Server Name` or `Players` field, a small popover window displays depending on which field you moused over -- `Server Name` contains server metadata, and `Players` contains gameplay data, score, and the player list.

Popover showing server metadata:
image

Popover showing gameplay variables and players, first for Deathmatch:
image

Now for Horde:
image

And finally CTF:
image

This allows for more interactivity with the launcher, as well as getting more important stats to eyeballs in an efficient way.

Second, I added a Server Details dialog that can be activated by either middle-clicking a server in the list or using the context menu on the server list:
image

This window will have everything the popovers had, with some additional information, like every cvar transmitted to the client via the server browser. This window stops all timers and blocks input on the main window until this window is closed.

image

At the top of this window, we have the MOTD display. This doesn't show if the server transmits an empty MOTD.

Below that on the left, we have the Server Metadata static box which largely replicates what is shown in the metadata popover (hovering over `Server Name`) with a few additions to encompass all details shown in the browser.

To the right of that, we see the Gameplay Variables static box which indicates the specific type of game that is being played. It points out variables that modify gameplay considerably, like rounds, score, and even modifiers like weapon strength (sv_weapondamage) and monster strength/health (sv_monsterhealth/sv_monsterdamage).

Below that we see the Players table, which is the same Player table that shows up in the permanent list, and player popovers, the only change is a visual indication of which slots are open to players and those that are not.

Below that, a Refresh/Join control. Refresh will refresh the server every 10 seconds and update its information. If the server is full, the Join button is disabled unless Refresh is on, then it turns into a "Join when slot free" toggle button. When a refresh indicates the server has an open slot, the launcher closes this window and launches the game.

To the left of this is the centerpiece of this window, the Gameplay Variables window. This is a wxCollapsablePane with a wxPanel, to show every cvar the server sent to the client. Cvars are separated into categories -- created manually but in the future can be part of the cvar signature -- to classify types of cvars. In order to inform the user on which cvar has which effect, we utilize the new `cvardocjson` JSON document to pull in cvar details and parse the help text in order to show it as a tooltip:

image

In Odalaunch, both `odamex_cvardoc.json` and `odasrv_cvardoc.json` are consumed even though only the latter is needed for this. The former might be useful if we wanted to, say, put in a user config analyzer/modifier in Odalaunch. The latter still being useful if we wanted to create a server launcher. Either way, both are read on program launch into a `std::unordered_map` for quick lookups based on variable name. If no JSON files are found, odalaunch doesn't care -- it will happily launch without loading the cvar descriptions -- you can tell this happened in the launcher if any server variable is not underlined.

This window has a ton of optimizations in it to keep it rather instantaneous in opening. The XRC is loaded on launch, and the window is simply hidden on close and reshown in order to keep window opening snappy. The Server Variables list is double-buffered, and the window launches on middle-click down instead of completing a middle click. Also, the Server Variables panels/controls are all rendered on idle after the window loads to keep the window opening snappy and still be faster than lazy loading them.

I tested this with Windows and the Flatpak which has wxWidgets 3.3, and it looks good in both. Could use a MacOS test drive.


odalaunch/src/dlg_main.cpp
electricbrass: Any chance you feel like fixing my hack and making this properly parse the json now that Odalaunch is linking against jsoncpp? I had just done this because it was one tiny use and didn't seem worth adding the library but its probably better to do it properly now

bcahue: Will do.

bcahue:
Converting to draft until I can fix defaults on aircontrol/gravity

endless-r0ad:
Tested on macOS Sonoma - works great!

electricbrass:
wxWidgets 3.3.3 just released and has a couple things that might be useful either for this PR or a future one:
- Add dark mode support to wxMessageBox and other dialogs in wxMSW.
- Support switching between light/dark modes in wxMSW.

bcahue:
I'd have to test it, but older versions iterations of dark mode had the wxAdvancedListCtrl that drives everything super laggy, and until that's fixed -- or fix it, as SLADE has dark mode enabled on a list control and performance is great there -- I don't think we should add dark mode on wxMSW.

bcahue:
Marking this as ready. I put the gravity/air control fields in gameplay variables in the server details window and player popups, and I also added colored percentage indicators on all % variables to let you know at a glance if its higher or lower than default.

I also added min/max/default values to the Server Variables window. Since the entire control was double buffered, this created a lag issue, so I made the expand panels double buffered instead, which fixed it, but you may notice a slight flicker of the resizer since I don't have a way to control its painting. It's not too noticeable but it's there, and I think its preferable to scroll lag for panels with a ton of content.

And one more change, I sorted those servers that say "<< NO RESPONSE >>" to the bottom of the list always. I don't think anyone has any need to have those towards the top of the list.

drivingspectrum (approved):

PR #1801, opened 2026-07-02
Apply bossaction linespecials using a copy of line 0

CurrentResident:
This brings Odamex in line with other source ports' interpretation of the UMAPINFO standard, as discussed in the Discord oda-development channel.

More urgently, it resolves a crash experienced in NT2F map22, where a bossaction has a linespecial of `27074`, which calls for a change of floor texture.  Because there is no *explicit* line associated with that special (nor does the standard allow for one), it's a matter of interpretation as to which sector's texture to source for the target tagged sector.  After discussing with Elf Alchemist, the common but undocumented approach is to use a copy of line 0 as the special's starting point.  So now we do that here, and this should avoid any additional undefined behavior owing to linespecials that require information from actual, real lines.

### DIFF
--- a/common/p_enemy.cpp
+++ b/common/p_enemy.cpp
@@ -3608,7 +3608,9 @@ void A_BossDeath(AActor *actor)
 					continue;
 				}
 
-				line_t ld;
+				// De-facto UMAPINFO standard interpretation: activate the bossaction linespecial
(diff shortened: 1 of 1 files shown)


electricbrass:
Lets toss this into stable since we're gonna get 1 more 12.x release out to fix the TX texture crash

Issue #1028, opened 2024-09-13
[BUG] 2-Sided Midtex always has a slight gap between it and the upper bound


rakohus:
**Describe the bug**
2-Sided Midtex always has a slight gap between it and the upper bound of where it should be. It is very noticable when people stack 2-sided linedefs with midtex.

**Build that the bug occurred in**
Odamex 10.6

**To Reproduce**
Run test wad

**Expected behavior**
No gap between midtex and where it should be. In this wad's case, the ceiling.

**Screenshots, NetDemos, & Crash Dumps**
[bug1028.zip](https://github.com/user-attachments/files/16987742/bug1028.zip)

electricbrass:
It looks like this was introduced in 0.7.0. Though there was an almost imperceptible gap in 0.6.x (and possibly earlier). This smaller gap is also present in ZDoom 1.23b33.

electricbrass:
`git bisect` indicates it originates from this commit:
https://github.com/odamex/odamex/commit/a67b11cd752a1113c5518805237e2a369947e957

Edit: I did it again and this time it gave me a different commit?? https://github.com/odamex/odamex/commit/3f745d9588582338fe94811c0820d42f75d9ae56

rakohus:
FINALLY FIXED THANK YOU @bcahue

PR #1806, opened 2026-07-03
[BUGFIX] Midtex Gap

### LINKED ISSUE #1567: [BUG] Skyboxes leave a midtext gap-style missing row of pixels at the bottom
**Describe the bug**
At the bottom edge of skyboxes where they meet the map horizon, there is a bottom row of pixels similar to the #1028 midtex gap issue. The differences here being that the gap is at the bottom instead of the top, and it leaves a HOM instead of vision to the back sector.

12.1.0

**To Reproduce**
Observe a map with a skybox.

**Expected behavior**
For the joint between the skybox and map horizon to be seamless without a HOM.

https://github.com/user-attachments/assets/aa5bed8c-a461-4ee9-ad5b-adc7863f24c5
Video (cybercrime3.wad pictured). You can briefly see the menu graphics and Railgun particles in the HOM gap.

**Additional context**
N/A

bcahue:
This PR was meant for the midtex gap, but the unlocking of knowledge on why it failed in the way it did allowed me to fix other bugs. This PR fixes #1567, #1028, and also fixes a bug I found with WADs with missing TX_ markers -- the -1 in W_CheckNumForName was messing it up a bit so I added a guard for it.

Now for the info you've all been waiting for! The midtex gap was caused by 4 things that I found:

1. R_BlastMaskedSegColumn had a legit off-by-one FRACUNIT when calculating dcol.yl.
2. MAX(mceilingclip[dcol.x], 0)) does the same thing mceilingclip[dcol.x] + 1) does (i.e. make it 0, never make it -1 which is what that array starts at) but does it in a way that allows every column on the screen to be drawn.
3. sprtopscreen/dcol.texturefrac use centeryfrac instead of centery which is what R_PrepWall uses, which can cause the midtexture to draw higher or lower a tiny (but noticeable, at least the debugger says) amount.
4. This had a lot to do with it as well, but R_PrepWall was creating scale1/scale2 values using a float, while being truncated to a fixed-point number to get the colfrac. This allowed the render function R_BlastMaskedSegColumn to become out of sync with what R_PrepWall prepared for midposts and what was drawn for R_BlastMaskedSegColumn, creating a gap. I tried an expensive calculation to match these up but it was too slow so instead I simply copied wallscalex when it was available to use for drawing midtex. It's separated by whether or not the texture has Y scaling, that should be tested.

I tried to also apply these changes to the Sprite blaster/particle draw function as well as they all had these foundational issues but less chance for them to pop up I suppose. Should make the sprite more accurate at least.

Also fixed is the Skybox HOM issue, now that I know floorclip should be 0 instead of -1 to end the clip.

Also mentioned at the beginning, I fixed an issue with TX_ markers as maps without TX_ markers had texture 3 overwritten, and luckily I had a test WAD where an animated sky frame was texture 3 so it flashed a texture and gave a huge clue.

### DIFF
--- a/client/src/r_plane.cpp
+++ b/client/src/r_plane.cpp
@@ -792,7 +792,7 @@ void R_DrawSkyBoxes()
 			else
 			{
 				ceilingclip[i] = pl->top[i];
-				floorclip[i] = pl->bottom[i];
+				floorclip[i] = pl->bottom[i] + 1;
 			}
 		}
 
--- a/client/src/r_plane.cpp
+++ b/client/src/r_plane.cpp
@@ -802,6 +802,7 @@ void R_DrawSkyBoxes()
 		ds_p->x2 = viewwidth - 1;
 		ds_p->silhouette = SIL_BOTH;
 		ds_p->midposts = NULL;
+		ds_p->midscales = NULL;
 		ds_p->curline = NULL;
 
 		// [RK] Allocate full width clip arrays.
--- a/client/src/r_segs.cpp
+++ b/client/src/r_segs.cpp
@@ -45,6 +45,7 @@
 // a pool of bytes allocated for sprite clipping arrays
 Pool<tallpost_t*> masked_midposts_pool(4096);
 Pool<int> sprclip_pool(4096);
+Pool<fixed_t> midscales_pool(4096);
 
 // OPTIMIZE: closed two sided lines as single sided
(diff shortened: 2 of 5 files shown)


electricbrass (approved):
Tested with a few wads, couldn't find any issues. Amazing work finally figuring this out.

PR #1809, opened 2026-07-03
A couple bash completion fixes + windows installer fix

electricbrass:


electricbrass:
This also includes a powershell completion script. Unlike the bash completion, it isn't installed at all because I'm just not satisfied enough with it to feel like it should be included with odamex, but it'll be here in the repo for anyone who wants it at least.

bcahue (approved):

Issue #631, opened 2022-03-22
[BUG] Gamma correction doesn't work with colored lighting


electricbrass:
**Describe the bug**
Gamma correction has no effect on sectors with colored lighting.

**Build that the bug occurred in**
10.0.0 (release version) and g354f1

**To Reproduce**
adjust gamma correction level

**Expected behavior**
It to get brighter or dark everywhere when adjusting gamma.

**Screenshots, NetDemos, & Crash Dumps**
![Screenshot 2022-03-21 230752](https://user-images.githubusercontent.com/84749759/159399994-2de66872-ed94-4915-b3d6-11bf283de790.png)

**Additional context**
map04 of odahorde is a good place to see this

leximayfield:
Thanks for the report.  I'm aware of the bug, but it is something that we need to take care of.

My memory from having run into the bug in the past is that it's only wrong so long as the level is loaded, and upon map load the gamma of the new lights is correct...unless the gamma is changed again.  There might also be an 8 vs 32-bit discrepancy.

bcahue:
Good to close this?

electricbrass:
Yep, looks to be fixed now

PR #1810, opened 2026-07-03
Fix colored lighting gamma refresh

rakohus:
This fixes gamma correction for sectors using colored lighting in 32-bit color mode, fixing issue #631. Previously, colored-light shademaps were created and cached on map load. While in game, if player changes gamma or F11 was used, the colored lighting was not refreshed, but would appear correct after restarting the map. This refreshes cached colored-light shademaps during gamma updates. It doesn't rebuild the 8-bit colormap.

### DIFF
--- a/client/src/v_palette.cpp
+++ b/client/src/v_palette.cpp
@@ -773,6 +773,7 @@ void BuildDefaultShademap(const palette_t* pal, shademap_t& maps)
 	}
 }
 
+static void V_RefreshSpecialLights();
 
 //
 // V_RefreshColormaps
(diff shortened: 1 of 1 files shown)


client/src/v_palette.cpp
```diff
@@ -988,21 +991,23 @@ fargb_t V_HSVtoRGB(const fahsv_t &color)
...
+		BuildLightRamp(*maps);
 
 	const argb_t* palette_colors = V_GetDefaultPalette()->basecolors;
 
 	// build normal (but colored) light mappings
 	for (unsigned int l = 0; l < NUMCOLORMAPS; l++)
 	{
 		// Build the colormap and shademap:
-		palindex_t* colormap = maps->colormap + 256 * l;
+		palindex_t* colormap = build_colormap ? maps->colormap + 256 * l : NULL;
```
electricbrass: Use `nullptr` instead of `NULL` here. Looks good otherwise.

electricbrass (changes requested):

electricbrass (approved):

PR #1817, opened 2026-07-05
[BUGFIX] More MBF-like targeting, with some performance improvements

bcahue:
This started with converting the real MBF friendly/hostile monster targeting. Horray! However, this did not alleviate the lag issues that occurred when spawning a friendly in during a slaughter map. I launched the same map with DSDA under MBF21 complevel, and these functions run as part of normal monster checking there -- and to my surprise, no lag at all!

So at that point, this became sort of a side quest for me to try and improve performance across the board and try to make the display functions take up top CPU which it should in the quest for great framerates, and lower the gamesim to the bottom of the chart.

When doing a vTune hotspot profile of Odamex, where typically you should see the display functions taking up most of the CPU, we saw gamesim functions take up most of the CPU:

image

Thats a ton of CPU time used for the gamesim, so the display struggles to find time to create frames for the tic before its time to render the next frame -- your bad framerate.

Drilling down to the bad friendly targeting, we see this:

image

This is an indication you should probably be checking the Memory Access / Microarchitecture Exploration analysis as well to check for memory binding and cache usage. A lot of times, these functions have high times because of low CPU cache utilization or high memory binding. It takes a little more digging to find out if there's bad memory locality, or bad speculation. It's also more involved to fix with not a lot to show for it.

For that particular problem, it was indeed bad caching.

Anyway, I'll go thru what I found and what I did:

1. Friendly Targeting -- This was death by 1000 cuts, the one thing that improved the frames significantly was reordering the variables in AActor so when functions compare those variables, it's much more likely to be a CPU cache hit and thus, very fast. I also made the lists of actors that are friendly or hostile intrusive doubly linked lists as they are in DSDA/PrBoom+, this way we can more easily keep track of head/tail of lists, and move items from head to tail at extreme speed (not sure if o(1) but it has to be close).
The only other thing that was going to help was to calm down Blockmap, which I did by emulating what DSDA/PrBoom+ has for this. To keep it quick I limited the blockmap count size to 9 blocks which covers all the Doom monsters (keep it quick with a fixed array), anything bigger is simply an adhoc heap allocation.

2. Excessive iteration -- We iterate every tic for things to set an effect to, and more than once a tic, in the case of scrollers. These iterate through the entire actor list every time for this, and in maps with 20000+ monsters, that is not a small chunk of CPU. If `co_helpfriends` is 1, it will also iterate the full actor list to find friends, albeit not every tic. So instead, we put these into their own list. An effects list for AActor pointers, and a vector of scroller pointers. Also, P_HelpFriend now uses the friendlies list to check for friends in danger. It's done this way so AActors doesn't need to be type checked to give an effect, and the same for scrollers, as I'll get to later.

3. `blockmapdist` short circuit on PIT_CheckThing. Surprisingly vanilla compatible.

4. Actor slabs -- In order to get better locality from memory when iterating huge numbers of actors, it makes sense to allocate actors from the same location in memory, preferably on the same memory page. An ActorSlab is just a group of bytes enough to hold 256 actors, put in a vector to keep locality. AActor's void* operator new(size_t count) is overloaded here to provide a means to load actor slabs (or put it on the heap if it's not really an actor) and allocate memory from said slabs. Actors allocate from the slabs instead of cold memory, and delete simply frees up their slot for the next monster. In startups/teardowns like new maps, it means the price is only paid once for actor allocations, then they're essentially free. The net result for this was, it simply moved the "fire" elsewhere, but after doing this, Actor iterations between "cold" actors became super fast, so I imagine it had some good effects -- combined with the actor threaded list, and actors that are iterated when they are nearby tend to fall within cache lines.

5. DObject::IsKindOf type checking storm from constant iterations -- This was running EVERY time we have to check for actors of a certain type -- doors, ceilings, floors, friendlies, scrollers, you name it. With 20000 monsters (and many more map objects) it was hitting some high CPU, and every hit was a cache miss. The steps I took in item 2 solved this immediately.

6. Sprite Improvements + Prefetch -- After being satisfied in the performance of oku2v31.wad, I checked nuts.wad and saw a frame drop when tons of sprites were on screen. The profiler was telling, much of R_ProjectSprite was hitting huge amounts of cache misses, and high CPU. First thing I did was hoist some calculations into sprite setup instead of sprite projection, then I cached the last spritedef encountered, thinking huge groups of monsters would use that locality due to being the same. Then I created a smarter clipping algo to disqualify sprites that will always be behind the player. Then I changed the R_GenerateVisSprite world space calculation to use less division. Then I prefetched the next sprite when the current one is projecting, to try and keep it within cache lines. There's probably more improvements to be done here, but they'd likely need some galaxy brain intervention. 😄

Here's what vTune looks like now, same scene (oku2v31.wad, friendly on field):

image

Much better! Top 3 of these are renderer functions (FixedDiv is used a lot there), the 4th is the function that adds sprites for projection, and the highest usage there is simply iterating all the actors, with no memory binding -- meaning there isn't much more we can do there, aside from parallelizing.

Oh yeah, I also added `co_zdoomfriendtargeting` to emulate ZDoom friendly targeting (or lack thereof), which is a release valve to not use the MBF targeting (and is much cheaper).


rakohus:
It's probably worth adding `co_zdoomfriendtargeting` to the sample configs.

common/actor.h
```diff
@@ -682,9 +797,13 @@ class AActor : public DThinker
...
-		// this actor can inhabit
-		std::vector<AActor*>  m_next;
-		std::vector<AActor**> m_prev;
+		// this actor can inhabit - point at the inline arrays unless the
+		// actor spans more than INLINE_BLOCKS blocks
+		int m_capacity;
+		AActor**  m_next;
+		AActor*** m_prev;
+		AActor*   m_inlinenext[INLINE_BLOCKS];
+		AActor**  m_inlineprev[INLINE_BLOCKS];
```
electricbrass:
```suggestion
		std::array<AActor*, INLINE_BLOCKS>  m_inlinenext;
		std::array<AActor**, INLINE_BLOCKS> m_inlineprev;
```

electricbrass (dismissed):

common/actor.h
bcahue: Done.

bcahue:
Also added the co_zdoomfriendtargeting var to the menu and the sample configs (set to 1 only on ZDoom configs).

drivingspectrum (approved):

PR #1820, opened 2026-07-06
Fix missing intermission when ExitLevel coincides with a netdemo snapshot-write

CurrentResident:
This also allows the other `gameaction`s to work if they are commanded on the same tic that a netdemo snapshot is written.  I suspect this was a leftover from when the `writeSnapshotData()` function was initially created and based on the SaveGame function.


bcahue (approved):

bcahue:
Brilliant find!

PR #1832, opened 2026-07-12
Fix for player spawns that touch a blocking line

### LINKED ISSUE #1825: [BUG] Xibalba-.wad broken spawns
**Describe the bug**
https://web.archive.org/web/20210301215416/https://odamex.net/bugs/show_bug.cgi?id=1125

Bringing this bug back from the dead as it is still prevalent in 12.2.1, stable and protobreak

**To Reproduce**
- Run odasrv with sv_gametype 1 and the attached xibalba-.wad
- Connect with client and join in
- Kill yourself with the rocket launcher in the map over and over (or some other way)
- You will only respawn at 2 deathmatch spawns on the map

**Expected behavior**
The other spawns should work

**Additional context**
It only seems to occur when connected to a server running this wad, strangely enough when connecting as a spectator, you may end up in 1 of the other spawns

I couldn't find this wad on the internet so I had to upload it, apologies to the author for the lack of text file if there was one with it.

[XIBALBA-.zip](https://github.com/user-attachments/files/29845908/XIBALBA-.zip)

.

endless-r0ad:
Fixes #1825

This change skips the PIT_CheckLine part of the overall "is this spawn obstructed? if so mark invalid and check the next one" check during player spawn.  PIT_CheckLine checks for blocking linedefs when a thing is moving (during spawn the player hasnt moved yet)

Tested with the below spawns. Note the top spawn in the image mimics the xibalba spawns that dont work. Currently only 3 of these spawns work on stable (the ones without the x) -
test-spawns

This allows players to spawn into a wall and possibly get stuck (the bottom spawn), however imo that example is clearly the mapper's fault. Spawns that are just grazing the wall like xibalba will now work

This is a slight change to vanilla behavior but not sure its worth making a co_ cvar for it?


electricbrass:
Is this bit of the mbf21 spec at all relevant here:
> Depending on how your port has modified spawning code, you may have a lingering bug related to the validcount variable. Basically, in vanilla doom, this variable is incremented in the wrong position in P_CheckPosition. The explanation for why this is a problem is complicated, but ripper projectiles (and possibly other cases) will expose this bug and cause desyncs. If your port strives for demo compatibility, I recommend adding an extra validcount increment in P_CheckPosition before running the P_BlockLinesIterator, as was done in dsda-doom (wrap it in a compatibility check for mbf21 if necessary): [code](https://github.com/kraflab/dsda-doom/blob/5ebd1c7860fae536dc46e5a310acb98ca59a3165/prboom2/src/p_map.c#L916-L920). Without that fix, vanilla heretic demos go out of sync, and the same will be true for mbf21 demos with certain dehacked features in play. A port like crispy doom wouldn't be affected because it hasn't modified the spawn code in a way that introduces the bugged validcount. Ports like pr+ and ee do have this problem, for example.

endless-r0ad:
> Is this bit of the mbf21 spec at all relevant here:
> 
> > Depending on how your port has modified spawning code, you may have a lingering bug related to the validcount variable. Basically, in vanilla doom, this variable is incremented in the wrong position in P_CheckPosition. The explanation for why this is a problem is complicated, but ripper projectiles (and possibly other cases) will expose this bug and cause desyncs. If your port strives for demo compatibility, I recommend adding an extra validcount increment in P_CheckPosition before running the P_BlockLinesIterator, as was done in dsda-doom (wrap it in a compatibility check for mbf21 if necessary): [code](https://github.com/kraflab/dsda-doom/blob/5ebd1c7860fae536dc46e5a310acb98ca59a3165/prboom2/src/p_map.c#L916-L920). Without that fix, vanilla heretic demos go out of sync, and the same will be true for mbf21 demos with certain dehacked features in play. A port like crispy doom wouldn't be affected because it hasn't modified the spawn code in a way that introduces the bugged validcount. Ports like pr+ and ee do have this problem, for example.

Wow great catch.. I think so. Seems like the one increment we do on line 1014 should not be happening if it's a spawn. As far as adding the extra validcount increment, I think we may want to add that fix on a new PR for mbf21/hexen compatibility

drivingspectrum:
Tested and it is fixed on my end, awesome work!

Edit: I updated the bug report with an image showing multiple players connected and in different spawn spots, so I wonder if this is a deeper issue?

electricbrass:
> This is a slight change to vanilla behavior but not sure its worth making a co_ cvar for it?

I definitely think a cvar for this would just be confusing. It would be hard to know what maps it matters for and unclear to users what it actually does.

However, this should at least only happen when not playing back demos, and even then, it seems to me like there could be old vanilla maps this change might break, with situations exactly as described where there are spawns inside walls. I wonder what port xibalba was designed for and what the difference there is exactly that allowed it to work. And I wonder if maybe we could somehow be smarter about detecting cases where the spawn isn't truly overlapping the line.

endless-r0ad:
> Tested and it is fixed on my end, awesome work!
> 
> Edit: I updated the bug report with an image showing multiple players connected and in different spawn spots, so I wonder if this is a deeper issue?

Note the comment I left on the issue for more detail, but I think the existing logic that allowed this to happen is correct.

endless-r0ad:
> > This is a slight change to vanilla behavior but not sure its worth making a co_ cvar for it?
> 
> I definitely think a cvar for this would just be confusing. It would be hard to know what maps it matters for and unclear to users what it actually does.
> 
> However, this should at least only happen when not playing back demos, and even then, it seems to me like there could be old vanilla maps this change might break, with situations exactly as described where there are spawns inside walls. I wonder what port xibalba was designed for and what the difference there is exactly that allowed it to work. And I wonder if maybe we could somehow be smarter about detecting cases where the spawn isn't truly overlapping the line.

Fix for demo playback added in latest commit. I was wondering the same thing about the port, the port must have fixed this before the wad was made. We could definitely have a smarter check but imo we should not override the mapper's spawn placement, even if it seems terribly placed like in a wall.

drivingspectrum:
I did a little more digging and found the wad with its text file in the wayback machine
https://web.archive.org/web/20220821081313/https://www.wad-archive.com/wad/Xibalba-1

It was released on the 10/5/11 for skulltag or zdaemon

electricbrass:
> we should not override the mapper's spawn placement

I would agree (or rather, I would love to make it error out and force the mapper to fix it), but changing vanilla behavior is something that needs to be done very carefully.

electricbrass:
> It was released on the 10/5/11 for skulltag or zdaemon

Well, it looks like zandronum at least is not skipping this check. Haven't checked yet to see how the spawns actually behave there.

electricbrass:
It looks like there is an MBF change that might be the proper fix for this:
https://github.com/fabiangreffrath/woof/blob/a62e4e175208fd104ae0c63585288cba19850427/src/p_map.c#L410
https://github.com/fabiangreffrath/woof/blob/a62e4e175208fd104ae0c63585288cba19850427/src/p_map.c#L363

We should test this and see if it solves the issue

electricbrass:
Well, maybe. Zandronum still appears to be doing something different, as it does not have this change.

electricbrass:
Just tested and the MBF change does allow these spawns to work. I wonder what zandronum is doing though

Edit: nevermind I tested wrong, it seems to make the middle of the 3 crossed out spawns to work but not the top one (didn't test the other).

electricbrass:
I think we're probably safe to enable that MBF change always (except for demo playback of course), though if we wanted to be extra cautious we could tie it to co_mbfphys. I'll make a PR for that (i have some ideas on doing it without needing a global variable like in woof) and keep investigating what's going on here.

endless-r0ad:
Feel free to close the PR and re-implement - my change fixes the spawn logic only for issue #1825, the MBF changes seem to include getting a moving player unstuck from a wall

endless-r0ad:
This is still a simple fix for #1825

bcahue:
Can we bump to protobreak?

electricbrass:
I'm still against this change, but unfortunately have not made much progress with investigating the reason this map works in the ports that it does

PR #1836, opened 2026-07-15
Allow IDDT during netdemo playback

CurrentResident:
All client-side cheats had been effectively disabled during netdemo playback.  This PR specifically allows the activation of IDDT, IDMUS, and IDMYPOS during netdemo playback (or pause).

It also moves the client-side-only code to a new cl_cheat.cpp + .h pair of files in client/src.  This was actually necessary to be able to naturally reference the `netdemo` instance.

Here's what it looks like with `netdebug_automap` enabled - note the netid numbers on each mobj:

Odamex_COOP_20260715_092645


client/src/st_stuff.cpp
```diff
@@ -547,15 +549,18 @@ bool ST_Responder(const event_t& ev)
 		{
 			if (cheat::AddKey(&cheat, static_cast<byte>(ev.data1), &eat))
 			{
-				if (cheat.DontCheck || cheat::AreCheatsEnabled())
+				if (   cheat::AreCheatsEnabled()
+				    or cheat.DontCheck
+				    or (cheat.AllowInNetdemoPlayback
+				        and (netdemo.isPlaying() or netdemo.isPaused())))
```
electricbrass: Not necessarily specific to this PR, but should we add a netdemo.isActive or something so we don't need to keep remembering check for both playing and paused everywhere?

electricbrass (dismissed):

client/src/st_stuff.cpp
CurrentResident: Yes, I totally agree.  Will add it.

electricbrass (approved):

PR #1838, opened 2026-07-16
[BUGFIX] Odalpapi NET_ReportError never includes file/line/function on GCC/Clang

electricbrass:
Fixes the errors always saying `[:0] BufferedSocket::()`.


drivingspectrum:
Nice catch!

drivingspectrum (approved):

PR #1839, opened 2026-07-16
Fix extra light and weapon raise speed issue when spying

### LINKED ISSUE #313: [BUG] Sector light raising permanently while spynexting player firing weapon
**Describe the bug**
The overall light level increases during spynext when a player fires, indicating that the sector light level upon weapon flash does not return to its normal level.

Odamex v0.9.0 (g749b6-6576)

**To Reproduce**
Spynext a player and have them fire weapons, the surrounding light level should stick to brighter than default.

**Expected behavior**
The light level returning to normal after the momentary muzzle flash.

I made a clip of this during the stream of a LMS event - https://clips.twitch.tv/SteamySlipperyOcelotMikeHogu-b10j3rnSC1TyfNGz

**Additional context**
No!!!!!!!!!

endless-r0ad:
This fixes #313 as well as a weapon raise/lower speed bug discovered during testing that happens when spying

Note consoleplayer on left, spying player on right-
Before

https://github.com/user-attachments/assets/e5913f48-1dac-45d2-964e-82e309beae22

After

https://github.com/user-attachments/assets/0dc81ca5-0104-44f5-9e7e-dbd1630ed31a


rakohus:
Will this also observe the weapon change speed on something like hordamex? I know the speed is modified there and it's a fairly popular wad

electricbrass:
Will this also observe the weapon change speed on something like hordamex?

Pretty sure it should, though testing to be sure would probably be best.

I wonder though if this might have any adverse effects on players that aren't spying, because in that case, CL_PredictSpying will not run. Because of the way the sprite offsets interact with weapon switching, I do wonder if this could cause desyncs (though in this case I assume very much temporary ones) with players being in the wrong psp state. It's probably not a big deal, since with spying we already see the same type of desync and it doesn't appear to cause any real issue, so changing it from too fast switching while spying to too slow switching while not spying might not really matter.

endless-r0ad:
> Will this also observe the weapon change speed on something like hordamex? I know the speed is modified there and it's a fairly popular wad

I tested and it seems the same display issue is there as well but its less noticable since the speed is faster. The fix resolves the slight issue there as well

endless-r0ad:
> Will this also observe the weapon change speed on something like hordamex?
> 
> Pretty sure it should, though testing to be sure would probably be best.
> 
> I wonder though if this might have any adverse effects on players that aren't spying, because in that case, CL_PredictSpying will not run. Because of the way the sprite offsets interact with weapon switching, I do wonder if this could cause desyncs (though in this case I assume very much temporary ones) with players being in the wrong psp state. It's probably not a big deal, since with spying we already see the same type of desync and it doesn't appear to cause any real issue, so changing it from too fast switching while spying to too slow switching while not spying might not really matter.

I dont think it should effect other players because the only time svc_playerstate is sent from the server is during spying (sent every tic) or when a client uses a cheat, svc_playerstate is sent to all players to let them know what player has what cheat

bcahue:
I think we'll also need this bumped to protobreak

PR #1842, opened 2026-07-18
[FEATURE] Recursively stack skyboxes (portals), implement ZDoom stacked sectors

### LINKED ISSUE #1572: [FEATURE] Thing 9081 Skybox Picker support
Now that Skyboxes have been implemented, I think a good next step would be the Skybox Picker.

https://zdoom.org/w/index.php?title=Classes:SkyPicker

The Skybox Picker (present in Zandronum and ZDaemon) allows individual sectors to choose their own skyboxes. There are some neat use cases that will be showcased below.

Support for actor 9081 Skybox Picker.

N/A

**Additional context**
https://github.com/user-attachments/assets/9542d794-965b-446a-acc5-d8669b04f91f
MAP05 of Rampage CTF has a skybox where the red and blue sides each view a different digital sky matching their team's colour.

https://github.com/user-attachments/assets/6616f74d-368e-44cb-b626-5881fcb5a11a
This is the ZDoom version of CQ3 v2 (WIP). Giving a Thing tag to a Skybox Viewpoint will normally nullify it. However, if it's targeting a Skybox Picker of the same tag, you can have a Skybox Viewpoint set within another skybox to create 'skybox-ception'. This allows for effects like rotating skyboxes.

bcahue:
We did it Odamex! We have portals! 🎉🎊🥳

Well, not quite...

What I am PR'ing today is an update to the skybox behavior. Now we are able to recursively draw skyboxes as long as the amount of skyboxes to draw is lower than the `r_portalrecursions` variable (default 16, max 64). This essentially marks skybox visplanes as additional views, almost as if its an unrendered backsector that need to have BSP rendered before they are displayed. If the render depth of these skyboxes reaches the value in `r_portalrecursions`, the remaining skybox visplanes get rendered as sky instead (and thus stopping the recursion).

Sound familiar? That's the same ideology behind line portals (with some more code for proper display, as segs are portals now) linked portals (lots more code for interactivity, since things need to see and path thru the portals and have it be seamless).

I also went the extra mile and implemented the stacked portal things from ZDoom, as they are another view-only skybox-like feature that mimics 3d floors. This means that sectors can have a different floor and ceiling skybox. [I've included a test WAD for this to demonstrate Odamex's skybox capabilities.](https://wads.doomleague.org/skybox_test.zip)

This resolves #1572 and puts the map display on par with ZDaemon and Zandronum.

However, I went off the campsite a bit here trying to make the test wad in that ticket work. I had to add CLIENTSIDE script flags and ACSe P-Code 276 (SetActorAngle). This adds a bit of a wrinkle because we don't transfer TIDs for most actors that come over the network, so CLIENTSIDE scripts won't have a ton to work off of when it comes to manipulating actors like this WAD. Maybe worth breaking protocol for, but beyond the scope of this PR.

That being said, the wad works fully with Odamex, ZDaemon and Zandronum now, and I verified it works online as well.


electricbrass:
> Maybe worth breaking protocol for, but beyond the scope of this PR.

Definitely seems worth it imo, but agreed, not in this PR. Since its just 1 extra field it should be doable in 12.3 though.

electricbrass (dismissed):

bcahue:
Converting to draft since I'm close to implementing stacked sectors (view only) from ZDoom.

electricbrass (approved):

PR #1849, opened 2026-07-19
Remove fossilised protocol guards and set protocol version to 1

drivingspectrum:
Also completes an outstanding TODO.

I was also going to remove the VersionRevision field, however I am unsure how this is handled in the recent odalaunch upgrades PR as protobreak is behind stable right now.


bcahue (approved):
Seems it makes current protobreak servers respond with << NO RESPONSE >> but I imagine that will change once servers integrate these changes.

drivingspectrum:
Huh weird yeah, I'm sure I mentioned in the message that the dev servers needed to be updated if it was merged, ah well.

PR #1851, opened 2026-07-20
[FEATURE] Mouse Driven Menu

### LINKED ISSUE #1397: [FEATURE] Mousewheel should scroll the console
Make scrolling console easier. Allow mousewheel to scroll up and down.

See title of request

**Additional context**
Some newer players mentioned this to me in a game and I agree it would be a good feature.

bcahue:
This really took too long for us to adopt, so I'm glad to add it! This PR introduces a way to navigate the current menu system using a mouse only. Left click will change a toggle, or drag/set a slider, while right click mimics the ESC key (go back 1 level). Scroll wheel will scroll menus as you'd expect.

The real thing enabling this is the creation of "input modes" to determine when UI should receive input, and marshal said input. Previously, Odamex would discard mouse input for everything except the game.

I also created a function for Odamex to determine the x/y of where a user clicked a window transformed into the scaled/emulated surface it came from. I'm not entirely sure if it's 100% right but it seems to work alright.

I also took it upon myself to make the console part of this UI input mode to marshal mouse events to `C_Responder()`, and implement scrolling and pasting with mouse in the console. Right click will paste there. This resolves #1397.

The actual menu part is pretty simple, instead of a conventional "hittest" function like you'd expect in a window system, the menu and options simply put the x/y/height/width of an option in an array on creation, and enumerate these on menu click to see if any x/y of the click landed on any of these menu items.

I also added a "yes/no" mouseable button set to the "exit game" screen, ala UZDoom, to allow using the mouse there.

I stopped short of creating an onscreen keyboard (again ala UZDoom) but with these functions an enterprising dev would be able to whip that up in a short amount of time using these "psuedo-hittest" functions and the DIGFONT or similar to make the keyboard graphics. I also chose not to add a "back" button as we don't have a graphic for it.

This should tide us over until we can implement a real UI system in Odamex.

### DIFF
--- a/client/sdl/i_input.cpp
+++ b/client/sdl/i_input.cpp
@@ -424,7 +424,7 @@ static void I_ApplyInputMode(EInputMode mode)
 	else if (mode == INPUT_MODE_UI)
 		input_subsystem->grabInputForUI();
 	else
-	input_subsystem->releaseInput();
+		input_subsystem->releaseInput();
 }
 
 
@@ -859,6 +859,8 @@ static int I_GetEventRepeaterKey(const event_t* ev)
 		button == OKEY_LSHIFT || button == OKEY_LCTRL || button == OKEY_LALT ||
 		button == OKEY_RSHIFT || button == OKEY_RCTRL || button == OKEY_RALT ||
 		button == OKEY_NUMLOCK)
+		return 0;
+	else if (button >= OKEY_MOUSE1 && button <= OKEY_MWHEELRIGHT)
 		return 0;
 	else if (button >= OKEY_HAT1 && button <= OKEY_HAT8)
 		return button;
--- a/client/sdl/i_input.cpp
+++ b/client/sdl/i_input.cpp
@@ -55,9 +55,20 @@
 bool tab_keydown = false;	// [ML] Actual status of tab key
 #endif
 
+EXTERN_CVAR(ui_mouse)
+
 static IInputSubsystem* input_subsystem = NULL;
 
 static bool nomouse = false;
+
+enum EInputMode
+{
+	INPUT_MODE_RELEASED,
+	INPUT_MODE_UI,
+	INPUT_MODE_GAME
+};
+
+static EInputMode current_input_mode = INPUT_MODE_RELEASED;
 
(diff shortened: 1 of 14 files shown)


electricbrass:
Scrolling with trackpads seems to be backwards.
I noticed you avoided one of my biggest complains about the menus in some other ports: clicking the window to focus it doesn't interact with the menu!!! (the other one being the atrocious onscreen keyboard and input boxes of gz/uz) Thank you, I hate so much when I click to focus a window and then it does something in the menu and I'm not quite sure what.
On the exit buttons, I personally prefer Woof's approach of just making left click exit and right click go back, but if we are gonna have the buttons, it feels strange to have them still appear when navigating with the keyboard, yet not have them be selectable with the keyboard.

electricbrass:
> I stopped short of creating an onscreen keyboard

I would probably try and clean up our usage of SDL_StartTextInput/SDL_StopTextInput so that theyre actually called when text input fields gain/lose focus, and just let the on screen keyboard be handled by whatever is on the system tbh. (Right now we just start text input at startup and never stop it, so the onscreen keyboard is just shown right when starting and then never again). Though maybe a built-in one as a backup wouldn't be a bad idea.

electricbrass:
Okay, slightly clearer versions of the main 2 issues I pointed out:
1. Scrolling does not respect OS settings for scroll direction. By default, on my setup at least, this makes it so that trackpad scrolling is backwards, while mouse scrolling is correct. I believe we might be using a flag to tell SDL to ignore these because we do want to ignore that for in-game inputs.
2. The yes/no exit buttons are only interact-able with the mouse, yet are still shown even when `ui_mouse` is off. They should not be shown when `ui_mouse` is disabled or they should be able to be used from the keyboard, or both.

bcahue:
I'll take a look. It does look like we check for SDL_MOUSEWHEEL_FLIPPED and reverse it if it's there. I'll gate that to only do that if the UI input mode is not on.

bcahue:
All suggestions should be implemented now.

electricbrass:
Seems good to me now!

electricbrass (approved):

PR #1853, opened 2026-07-22
[BUGFIX] `exec` and `-config` not searching relative paths

electricbrass:
Before 11.1.0, `exec` and `-config` would search the CWD if the path given was a relative path. 11.1.0 unintentionally changed this so that it would require an explicitly relative path starting with `./` or `../`. This PR restores the old behavior and also makes `M_FindUserFileName`/`M_GetUserFileName` and `OWantFile` constructors perform tilde expansion, which fixes most cases of tildes not being expanded in file paths.


electricbrass:
There's a couple other miscellaneous bug fixes in here now too

bcahue (approved):

Issue #1477, opened 2025-12-06
[BUG] Teleport to line with same tag desyncs in multiplayer


electricbrass:
**Describe the bug**
EV_LineTeleport calculates the point on the linedef where the thing crossed in order to calculate how far along the destination linedef it should be teleported to. This works in singleplayer. In multiplayer, the client may receive the server's `svc_activateline` message *after* the client has already teleported the thing. This results in the calculations using the thing's position ending up completely wrong, and the thing can get teleported to completely the wrong location clientside.

**Build that the bug occurred in**
11.2.0, various 12.0.0 builds, presumably has existed since at least 10.0

**To Reproduce**
Go to map01 of this wad in multiplayer
[keyscrolltele.zip](https://github.com/user-attachments/files/24003792/keyscrolltele.zip)
The red key should loop on the conveyor forever, and it will be doing that serverside. But clientside it will teleport to the space to the player's left, and continue teleporting farther and farther down the hall each time the server tells it to teleport.

**Expected behavior**
Teleport to the correct location!

**Screenshots, NetDemos, & Crash Dumps**
11.2.0 netdemo
[Odamex_COOP_20251206_173850_keyscrolltele.wad_MAP01.zip](https://github.com/user-attachments/files/24003823/Odamex_COOP_20251206_173850_keyscrolltele.wad_MAP01.zip)

**Additional context**

PR #1856, opened 2026-07-22
Fix teleport overshoot via duplicated teleport actions

CurrentResident:
The client is allowed to predict mobj teleportation via line activation.  However, the client also performs the mobj teleportation when the ActivatedLine message comes in from the server.  For many (or most?) teleport handler actions, this is fine because the teleport destination is an absolute position via TELEPORTMAN; Back-to-back activations simply set the mobj to the destination point.

For line-to-line teleporters, this actually can cause essentially a duplicated teleportation where the mobj's relative offset from the triggering line is preserved WRT the destination line.  For a back-to-back activation, this essentially results in a 2x overshoot of the destination, resulting in the mobj wandering far off of the destination for several tics until a regularly-scheduled UpdateMobj puts the mobj back to where the server says it really belongs.

This PR addresses the issue by monitoring when a line activation results in the teleportation of the activator, and when it happens, follow up with an immediate UpdateMobj to force the mobj to proceed into the tic with a synchronized position, correcting any overshoot immediately.

Before:  (notice the monsters "flicker" into existence)

https://github.com/user-attachments/assets/09f05093-f5da-43ad-b92b-b416e224d8c3

After:

https://github.com/user-attachments/assets/9a6ec21e-d405-4505-b560-aceb1f8de382


bcahue (approved):

electricbrass:
Does this fix #1477?

CurrentResident:
> Does this fix #1477?

It sure does.  That's very much the case that I was looking to address with this!  There are still other teleport + prediction-related cases, but this PR addresses the one specifically described in the linked issue.

CurrentResident:
~~OH!  No it does not.  I just tried the demo map.  Let me look closer.~~

Helps to be checked out to the correct branch!

CurrentResident:
The person responsible for hiring the person responsible for sacking Resident has now been sacked.

Okay, I eat my boot.  I was checked out to the wrong branch when I tried out the red key demo wad.  It does work.

https://github.com/user-attachments/assets/c493978a-6399-4042-afb8-b6130d7f9fca

PR #1857, opened 2026-07-24
Fix for P_JustTeleported being true on odasrv startup

endless-r0ad:
Seeing a crash on odasrv when both of these are nullptr

Screenshot from 2026-07-24 09-25-00

### DIFF
--- a/common/p_map.cpp
+++ b/common/p_map.cpp
@@ -192,7 +192,7 @@ void P_ClearJustTeleported()
 
 bool P_JustTeleported (AActor* thing)
 {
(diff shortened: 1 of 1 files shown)


CurrentResident:
Oof!  Thanks for the fix.  What wad were you using where the crash happened?

endless-r0ad:
> Oof! Thanks for the fix. What wad were you using where the crash happened?

Xibalba wad https://github.com/user-attachments/files/29845908/XIBALBA-.zip

bcahue (approved):

PR #1867, opened 2026-07-30
Implement MBF's tmunstuck thing

electricbrass:
This allows players to escape from being stuck inside of blocking/one-sided lines in certain cases.

MBF implements this using a global variable for passing extra information into PIT_CheckLine. To avoid that, I made the blockmap iterators into variadic templates that allow passing in additional arguments to the PIT_* functions. Because of this, I also took the opportunity to clean up many of the existing globals in the same way.

This is a partial fix to #1825, but more investigation is required to figure out the complete solution for it. See comments on #1832

I also am using this PR somewhat to test how the clang-tidy bot behaves on real PRs, so there's some changes and code cleanup related to that too.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -52,6 +52,8 @@ Checks: |
   -readability-named-parameter,
   # we mostly use lowercase
   -readability-uppercase-literal-suffix,
+  # simplified form can be less clear sometimes
+  -readability-simplify-boolean-expr
 CheckOptions:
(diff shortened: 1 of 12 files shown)


electricbrass:
The last clang-tidy warning seems to be a bug in clang tidy 21, I don't get it locally with 22 and what it's saying is definitely wrong.

drivingspectrum (approved):

PR #1870, opened 2026-07-31
Submodulize dependency libraries and build as static libs

CurrentResident:
### Overview

Substantial changes to the build process:

- **Switch dependencies over to being git submodules.**
  - It used to be a mix of binary downloads and submodules, but now it's all submodules.
- **Use CMake's FetchContent module to pull the submodules into the build and override `find_package`**.
  - Please see `libraries/CMakeLists.txt`, which does everything.
  - Due to how FetchContent works, some libraries need to be exposed via `FetchContent_MakeAvailable` before others because they are also dependencies of these libraries.  For example, zlib is used by libpng, SDL, wxWidgets, and its find module must have been completely overridden by the time the other libraries' CMakeLists are pulled in.
- **Build the dependencies during the build phase**
- **Disable a variety of warnings in dependency libraries that are now treated as errors.**

### ACTION REQUIRED

Developers will need to run a `git submodule update --init --recursive` when checking out this branch or any other that has merged this in.

### Action requested

Also, please feel free to try building it yourself and adding comments, questions, concerns, objections, etc.  Any insight as to whether the debug symbol files need to be extracted on MacOS would be welcome too!


.github/workflows/release.yml
```diff
@@ -295,56 +289,6 @@ jobs:
...
-            cp "${found}" "${DEPS_PREFIX}/licenses/${dest}"
-          fi
-        }
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external" "LICENSE.gme.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/ogg" "LICENSE.ogg-vorbis.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/vorbis" "LICENSE.ogg-vorbis.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/opus" "LICENSE.opus.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/opusfile" "LICENSE.opusfile.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/wavpack" "LICENSE.wavpack.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/xmp" "LICENSE.xmp.txt"
```
CurrentResident, line 338: OOPS!  Add the license copying back in.

.github/workflows/macos-rc.yml
```diff
@@ -125,56 +119,6 @@ jobs:
...
-            cp "${found}" "${DEPS_PREFIX}/licenses/${dest}"
-          fi
-        }
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external" "LICENSE.gme.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/ogg" "LICENSE.ogg-vorbis.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/vorbis" "LICENSE.ogg-vorbis.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/opus" "LICENSE.opus.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/opusfile" "LICENSE.opusfile.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/wavpack" "LICENSE.wavpack.txt"
-        copy_license_glob "SDL2_mixer-${{ env.sdl2_mixer_version }}/external/xmp" "LICENSE.xmp.txt"
```
CurrentResident, line 168: Same thing here.

CMakeLists.txt
```diff
@@ -20,6 +22,10 @@ project(Odamex VERSION 13.0.0)
 
 include(CMakeDependentOption)
 
+if (APPLE)
+    enable_language(OBJC OBJCXX)
+endif()
+
```
CurrentResident, line 30: Add a comment explaining that SDL needed this to build on MacOS.

libraries/CMakeLists.txt
```diff
@@ -146,26 +23,406 @@ if(BUILD_CLIENT OR BUILD_SERVER OR BUILD_TESTS)
...
+
+    if (BUILD_CLIENT)
+
+        # ======================================
+        # PNG
+        # ======================================
+        if (USE_INTERNAL_PNG)
+            set(FETCHCONTENT_SOURCE_DIR_PNG "${CMAKE_CURRENT_SOURCE_DIR}/libpng")
+            FetchContent_Declare(
+                PNG
```
CurrentResident: PNG needs to also be built if `BUILD_LAUNCHER`.

libraries/CMakeLists.txt
```diff
@@ -146,26 +23,406 @@ if(BUILD_CLIENT OR BUILD_SERVER OR BUILD_TESTS)
...
+    # ======================================
+    # zlib
+    # ======================================
+    if (USE_INTERNAL_ZLIB)
+        set (FETCHCONTENT_SOURCE_DIR_ZLIB "${CMAKE_CURRENT_SOURCE_DIR}/zlib")
+        FetchContent_Declare(
+            ZLIB
+            URL "${FETCHCONTENT_SOURCE_DIR_ZLIB}"
+            OVERRIDE_FIND_PACKAGE
+            )
```
CurrentResident: ZLIB needs to also be built if `BUILD_LAUNCHER`.

packaging/flatpak/net.odamex.Odamex.yml
```diff
@@ -92,14 +42,15 @@ modules:
       - -D BUILD_LAUNCHER=1
       - -D BUILD_TESTS=0
       - -D BUILD_MASTER=0
-      - -D BUILD_TESTS=0
       - -D USE_INTERNAL_LIBS=0
+      - -D USE_INTERNAL_CURL=1
```
electricbrass, line 46: We want to be using the flatpak runtime libraries in the flatpak whenever they're present (except for SDL mixer). See USE_INTERNAL_LIBS being disabled for the flatpak.

CMakeLists.txt
electricbrass, line 19: If we are using SYSTEM with add_subdirectory, this needs to be 3.25. And FIND_PACKAGE_ARGS (see other comments) will require 3.24.

ci/ubuntu-jammy.Dockerfile
```diff
@@ -10,8 +10,8 @@ ENV TZ=US \
 # Packages - first the majority of them, then cmake
 RUN set -x && \
     apt update && \
-    apt install -y git g++ ninja-build libsdl2-dev libsdl2-mixer-dev \
```
electricbrass, line 13: This and the fedora dockerfile are to make sure compatibility is maintained with the versions of the system libraries on these. I suppose the question that needs deciding then is are we planning drop officially supporting building with those or not. If not, we should not change this here.

libraries/CMakeLists.txt
```diff
@@ -1,136 +1,13 @@
...
-      --config ${CMAKE_BUILD_TYPE} --target install
-      --parallel ${PARALLEL_PROC_COUNT} RESULT_VARIABLE _RESVAL)
-    if(_RESVAL GREATER 0)
-      message(FATAL_ERROR "Library ${_LIBRARY} did not build correctly.")
-    endif()
-  endif()
-endfunction()
+include(FetchContent)
+
+set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)   # option() honors normal variables
```
electricbrass, line 10: This is already NEW because of the minimum cmake being over 3.13, so it doesn't need to be set.

libraries/CMakeLists.txt
```diff
@@ -146,26 +23,406 @@ if(BUILD_CLIENT OR BUILD_SERVER OR BUILD_TESTS)
...
+
+if (BUILD_TESTS)
+    # ======================================
+    # GTest
+    # ======================================
+    set (FETCHCONTENT_SOURCE_DIR_GTEST "${CMAKE_CURRENT_SOURCE_DIR}/googletest")
+    FetchContent_Declare(
+        GTest
+        URL "${FETCHCONTENT_SOURCE_DIR_GTEST}"
+        OVERRIDE_FIND_PACKAGE
```
electricbrass, line 36: Could we also add FIND_PACKAGE_ARGS to most of these dependencies (haven't looked yet to see which ones exactly it does or does not make sense for)? Right now FETCHCONTENT_TRY_FIND_PACKAGE_MODE will be all or nothing, instead of allowing OPT_IN to have mixed usage where it makes sense, and I also think that having the default to be to search for system libraries first makes sense.

CMakeLists.txt
electricbrass, line 49: Does USE_INTERNAL_LIBS get used anymore now?

.gitmodules
```diff
@@ -34,3 +34,15 @@
...
+	path = libraries/sdl
+	url = https://github.com/libsdl-org/SDL.git
+[submodule "libraries/sdl_mixer"]
+	path = libraries/sdl_mixer
+	url = https://github.com/libsdl-org/SDL_mixer.git
+[submodule "libraries/wxwidgets"]
+	path = libraries/wxwidgets
+	url = https://github.com/wxWidgets/wxWidgets.git
+[submodule "libraries/libdwarf-lite"]
+	path = libraries/libdwarf-lite
```
electricbrass, line 47: Should we go ahead and make fmt a submodule too? I don't actually know why that one wasn't already.

cmake/modules/CreateDebugSymbolFile.cmake
```diff
@@ -0,0 +1,39 @@
+function(create_debug_symbol_file)
```
electricbrass, line 1: Does this affect the Flatpak builds? Flatpak already strips debug symbols and provides separate .Debug bundles containing them.

electricbrass (changes requested):

.gitmodules
electricbrass: This would be one to leave FIND_PACKAGE_ARGS off of. We definitely want to be getting exactly the right version on this one.

cmake/modules/CreateDebugSymbolFile.cmake
electricbrass: And does it hurt the stack traces that some errors print?

PR #1871, opened 2026-07-31
[FEATURE] Compile cvardocjson into Odalaunch instead

bcahue:
This PR introduces a way for the CI/CD pipelines (and local builds) generate cvardocjson files to consume in Odalaunch.

This PR got away from me since I operated under the assumption that all of our cvars were literal, when in fact we had 1 (snd_musicsystem) that created different defaults depending on platform.

Rather than create an exception system for this, since there was only 1 cvar that fit this criteria, I just made the decision that all cvars should be literal and the same across platforms -- and changed snd_musicsystem to have an `MS_AUTO` default which selects a smart default for the platform you are on (matching old functionality -- old configs will have the actual subsystem selected, new configs will use AUTO which selects those same subsystems.

I also made new command flags: `--cvardocjson` and `--cvardoc`, which will launch Odamex/Odasrv, dump the files requested before any subsystems are loaded or the ROOT user check is done, then exits. This ensures a WAD is not required to dump cvar data. I also made --version use this "info dump" paradigm.

The short of it is that the cvardocjson gets compiled into Odalaunch now instead of depending on the odamex/odasrv_cvardoc.json files being in the same directory as Odalaunch. A CMake script runs after building the client/server to generate cvar documentation, and always saves these json files in odalaunch/res/ before compiling them into the build directory, where CMake adds it to the Odalaunch sources.

This gives less freedom for the user to sub their own, but creates a much more user friendly file layout with static files coming down the pike, and always marries the version or branch's cvar documentation to the actual cvars, so there's no chance for drift. With no json files to delete, we ensure the documentation is always included and Odalaunch never enters a degraded mode where it has no information on cvars.


client/sdl/i_music.cpp
```diff
@@ -179,6 +179,39 @@ void I_SetMusicVolume (float volume)
...
+#endif
+}
+
+//
+// I_ResolveMusicSystem
+//
+// Turns whatever snd_musicsystem holds into a music system we can actually
+// start.
+//
+MusicSystemType I_ResolveMusicSystem(int musicsystem_type)
```
electricbrass: Should be constexpr

client/sdl/i_music.cpp
```diff
@@ -179,6 +179,39 @@ void I_SetMusicVolume (float volume)
 		musicsystem->setVolume(volume);
 }
 
+//
+// I_GetDefaultMusicSystem
+//
+// The music system to use when the user has expressed no preference, which
+// depends on what the platform does best.
+//
+MusicSystemType I_GetDefaultMusicSystem()
```
electricbrass: should be constexpr (and a comment to change that to consteval when it gets into protobreak)

electricbrass (changes requested):
Should add the new cli options to the tab completions. They should go in the `exclusiveopts` table in the bash one, and should have `-Exclusive` in the powershell one.

electricbrass:
All the misc-include-cleaner warnings can be ignored, after #1867 those will stop happening.

client/sdl/i_music.h
```diff
@@ -57,7 +62,7 @@ EXTERN_CVAR(snd_musicsystem)
 // [ML] Keep track of the currently loaded music lump name
 extern std::string currentmusic;
 
-void I_InitMusic(MusicSystemType musicsystem_type = (MusicSystemType)snd_musicsystem.asInt());
+void I_InitMusic(MusicSystemType musicsystem_type = I_ResolveMusicSystem(snd_musicsystem.asInt()));
 void STACK_ARGS I_ShutdownMusic(void);
```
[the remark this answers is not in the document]
electricbrass: Hmm, I'm not sure what needs to be done to fix this. I think this is caused by doomtype.h not being included in this file at all. Does I_ShutdownMusic even need STACK_ARGS anymore? Or does addterm require that?

CMakeLists.txt
```diff
@@ -210,6 +233,33 @@ if(NOT BUILD_CLIENT AND NOT BUILD_SERVER AND NOT BUILD_MASTER AND NOT BUILD_LAUN
...
 endif()
 
+# The "cvardocs" target refreshes the committed documentation on demand.
+if(TARGET odamex OR TARGET odasrv)
+  add_custom_target(cvardocs COMMENT "Refreshing cvar documentation")
+  add_custom_target(cvardocs-check COMMENT "Checking cvar documentation is up to date")
+
+  foreach(_cvardoc_target odamex odasrv)
+    if(TARGET ${_cvardoc_target})
+      add_custom_command(TARGET cvardocs POST_BUILD
```
electricbrass: Just curious, why are these custom commands using standalone cmake scripts instead of just functions in modules included from `cmake/modules`? That seems simpler but maybe there's something I'm not seeing.

bcahue: I was kinda throwing stuff at a wall trying to get these scripts to work but that does seem like the proper spot to place it, I'll change it

odalaunch/src/cvardoc_data.h
```diff
@@ -0,0 +1,31 @@
...
+//-----------------------------------------------------------------------------
+
+#pragma once
+
+#include <cstddef>
+
+// These variables get filled in by cmake on build
+extern const char ODAMEX_CVARDOC_JSON[];
+extern const size_t ODAMEX_CVARDOC_JSON_LEN;
+extern const char ODASRV_CVARDOC_JSON[];
```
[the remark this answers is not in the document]
electricbrass: you can put 
```
// NOLINTNEXTLINE(modernize-avoid-c-arrays)
```
on the line before the declaration to silence this one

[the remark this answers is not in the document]
electricbrass: Oh wait this is the externs
In that case easier to wrap with 
```
// NOLINTBEGIN(modernize-avoid-c-arrays)
...
// NOLINTEND(modernize-avoid-c-arrays)
```
Alternatively, keep the arrays internal, and create std::string_views that are what get `extern`'d instead (though I think jsoncpp doesnt have string_view support).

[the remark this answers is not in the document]
bcahue: I'm trying to convert the generator into string_view to silence this, figured its better than a linter disable, and added benefit of not having to drag around the length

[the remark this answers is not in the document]
electricbrass: > though I think jsoncpp doesnt have string_view support

Except now i see theres also a size, so presumably wherever this is being used could use a string_view data + size pair.

client/sdl/i_music.cpp
bcahue: Completed

client/sdl/i_music.cpp
bcahue: Completed

electricbrass (approved):

CMakeLists.txt
bcahue: This is completed, I still had to call a cmake script because I can only attach scripts to post_build commands, as the variables aren't there anymore once configure is over. I still made it a little elegant though, turning it into 1 module file that can operate as an import or standalone script file depending on how its called.

PR #1873, opened 2026-08-01
[BUGFIX] 12.2 compatibility fix and various ednum conflict fixes

electricbrass:
1. mobjtype_t and odamobjinfo have been reordered to match previous 12.x versions, restoring compatibility between up to date clients and older servers
2. If a thing has been defined in dehacked to use an ednum belonging to a team start, coop start 5-64, horde spawner thing, polyobject, sector silencer, or any of the skybox/stacked sector related things, the dehacked thing takes priority and is spawned in place of the built-in mobj type.
3. Avatars are now spawned on Hexen format maps
4. A crash when P_IsUnderDamage is called and a sector's ceilingdata is a door has been fixed.

The rest of the changes are cause I decided to replace `clamp` with `std::clamp`


common/p_setup.cpp
```diff
@@ -960,6 +960,13 @@ void P_LoadThings2 (int lump, int position)
 		P_SpawnMapThing(*mt, position);
 	}
 
+	// Sort by player number if starts are not in order
+	std::sort(playerstarts.begin(), playerstarts.end(), [](const mapthing2_t& p1, const mapthing2_t& p2){
+		return P_GetMapThingPlayerNumber(p1) < P_GetMapThingPlayerNumber(p2);
+	});
+
+	P_SpawnAvatars();
```
CurrentResident, line 968: 👍

common/info.h
```diff
@@ -1543,6 +1543,12 @@ enum mobjtype_t: int32_t {
 	MT_EXTRALIFE,
 	MT_RESTEAMMATE,
 
+	// TODO: 13.0.0, delete these and uncomment the earlier ones
+	// for 12.3 they have to be here because modifying internal
+	// mobjtype nums breaks version compatibility
+	MT_UPPERSTACK,
+    MT_LOWERSTACK,
+
```
CurrentResident, line 1551: Is there a compelling reason that they should be defined up at line 1493 instead of here?  Could we just leave them here as their permanent home?

common/p_spec.cpp
```diff
@@ -198,14 +198,13 @@ int P_ArgToCrush(byte arg)
...
 	int dir = 0;
-	for (seclist = actor->touching_sectorlist; seclist; seclist = seclist->m_tnext)
+	for (const msecnode_t* seclist = actor->touching_sectorlist; seclist; seclist = seclist->m_tnext)
 	{
-		if ((cr = (DCeiling*)seclist->m_sector->ceilingdata) && cr->m_Status == 2) // Down
+		const DSectorEffect* cr = seclist->m_sector->ceilingdata; // Crushing ceiling
+		if (cr && cr->IsKindOf(RUNTIME_CLASS(DCeiling)) && static_cast<const DCeiling*>(cr)->m_Status == 2) // Down
 		{
-			cr->m_Crush > NO_CRUSH ? dir = 1 : dir = 0;
+			static_cast<const DCeiling*>(cr)->m_Crush > NO_CRUSH ? dir = 1 : dir = 0;
```
CurrentResident: Totally random thought...  Is there such a thing as a floor crusher?

common/p_mobj.cpp
```diff
@@ -2936,26 +2936,23 @@ void P_ResolveStackLinks()
 }
 
 //
-// P_IsSpawnThing
+// P_IsPlayerSpawnThing
 //
 // Returns true if the mapthing2_t is a spawn
 //
-bool P_IsSpawnThing(const mapthing2_t& mt)
+bool P_IsPlayerSpawnThing(const mapthing2_t& mt)
```
CurrentResident, line 2943: Note to self, or whoever merges from stable to protobreak:  there's a `const bool` that should be replaced with a call to this function.

common/info.h
electricbrass: Organization, puts them next to the other things they're related to, but I supposed they could just stay here

common/p_spec.cpp
electricbrass: Yep, Floor_RaiseAndCrush. I don't believe floor crushers exist at all on doom format maps, at least in any way that we support. So adding a check here for them could be done without affecting compatibility.

common/info.h
CurrentResident: I think organization is a good reason.  The question is resolved as far as I'm concerned.

Tangentially, since the introduction of these enumerals is a breaking change, would it have made more sense if the originating PR went into protobreak instead?  If not (or even if it is), would it be worth considering further reorganization in protobreak so that more related types can be grouped together into a set of reserved blocks of enumeral ranges?

----

Meta question - Since I initiated this comment thread, shouldn't I be able to mark it resolved?  Is there a github project setting that would allow me to do so?

electricbrass: > Tangentially, since the introduction of these enumerals is a breaking change, would it have made more sense if the originating PR went into protobreak instead?

Probably

> If not (or even if it is), would it be worth considering further reorganization in protobreak so that more related types can be grouped together into a set of reserved blocks of enumeral ranges?

Basically everything else is already grouped in a way that makes sense. For cases like this would it would pretty much need going through all things zdoom adds and deciding which might be relevant for us to add in the future and leaving some spaces in case of that, which would also require further reworks because right now other areas assume that all the internally defined mobjtypes are in one of two contiguous ranges. All the ones in positive range are essentially fully set in stone and will never change. The negative part, where these are, will almost certainly continue to have additions over time, but they are still relatively rare, and most will likely be in new categories that can just go at the end. So would it be better to have a way of introducing these without it being a breaking change? Yes. Is it worth it right now? Not sure.

> Meta question - Since I initiated this comment thread, shouldn't I be able to mark it resolved? Is there a github project setting that would allow me to do so?

I would have thought so. No idea if there's a setting.

electricbrass:
I think the remaining clang-tidy warnings are outside the scope of this PR

PR #1874, opened 2026-08-01
Include flatpak debug bundles in artifacts and releases

electricbrass:
Adds Flatpak bundles containing the net.odamex.Odamex.Debug package in all places where we publish Flatpak bundles. These can be installed to get debug symbols for the Flatpak.

I had been waiting for https://github.com/flatpak/flatpak-github-actions to do a new release but with the looming release of 12.3 I figured we should go ahead and do it.

### DIFF
--- a/.github/actions/odatests-notify/action.yml
+++ b/.github/actions/odatests-notify/action.yml
@@ -17,7 +17,7 @@ runs:
   using: "composite"
   steps:
     - name: Checkout odatests-results branch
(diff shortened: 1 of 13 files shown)


electricbrass:
Also fixes a bug where the redistributable version in the iss script was not quoted properly, preventing the windows installer from being built.

bcahue (approved):

Issue #1814, opened 2026-07-04
[BUG] Opening PPSS2 via drag-and-drop in explorer causes a fatal error


nikovaan:
**Describe the bug**
Opening PPSS2 (Pacifist Paradise: Secret Santa 2) via drag-and-drop using explorer.exe in windows 11 causes a fatal error. Opening it via odalaunch or by connecting to a server running the wad works just fine.
Error message: `Action RandomJump on state 3068 expects no more than 0 nonzero args (8 found). Check your DEHACKED.`
When checking the DEHACKED lump CODEPTR list, Frame 3068 is listed as SpawnObject and not RandomJump.

**Build that the bug occurred in**
Tested in 12.2.0, 12.2.1, and 13.0.0 test build gremotes/pull/1811/merge-0-g16d0f-1 (https://github.com/odamex/odamex/actions/runs/28667935256)

**To Reproduce**
Drag-and-drop PPSS2 to odamex.exe.

**Expected behavior**
Launching the wad normally.

**Screenshots, NetDemos, & Crash Dumps**
<img width="552" height="132" alt="Image" src="https://github.com/user-attachments/assets/71e01dd1-eda9-4a24-a5a2-84e9af63f3d6" />

**Additional context**
wad available from https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/ppss2

bcahue:
Debugging this a bit, seems like the WAD is being parsed as a DEH patch?

Issue #1813, opened 2026-07-03
[BUG] Loading Corruption.wad only works via FLTK launcher


drivingspectrum:
**Describe the bug**
Corruption.wad will only load when using the FLTK launcher, it will throw an R_Textures error when:
* dragging and dropping onto the odamex executable
* wad ccmd
* -file

**Build that the bug occurred in**
Tested on 12.2.1

**Expected behavior**
It should load with any method

<img width="352" height="133" alt="Image" src="https://github.com/user-attachments/assets/eeeea7fb-f7e6-4279-a375-94a888289174" />

drivingspectrum:
Looks like this issue is related to issue #1737, which will also load using the fltk interface

electricbrass:
OH I know what this is. Corruption looks like a valid iwad so it tries to load it as an iwad

rakohus:
> OH I know what this is. Corruption looks like a valid iwad so it tries to load it as an iwad

Yes, this is the same issue as judgment.wad (which I guess we never put on the tracker), where it contains all 6 of the designated IWAD lumps found in w_ident.cpp.

Issue #1737, opened 2026-05-15
[BUG] Launching We Have Doom At Home via drag and drop in explorer causes a fatal error


nikovaan:
**Describe the bug**
Trying to launch We Have Doom At Home (whdah.wad) causes a fatal error that just says Sound if you try to launch it by dragging whdah.wad to odamex.exe.

**Build that the bug occurred in**
12.0.0, 12.1.0, 12.2.0

**To Reproduce**
Drag and drop whdah.wad to odamex.exe in explorer.

**Expected behavior**
Wad would be launchable via drag and drop in explorer.

**Screenshots, NetDemos, & Crash Dumps**
<img width="349" height="121" alt="Image" src="https://github.com/user-attachments/assets/59e26ee2-1668-41ff-a5a6-10c05d14db73" />

<img width="976" height="331" alt="Image" src="https://github.com/user-attachments/assets/b847c662-a053-484c-9207-2e8c83baf42f" />

No crash dump was generated.

**Additional context**
OS: Windows 11
wad available at: https://www.doomworld.com/forum/topic/154723-we-have-doom-at-home-mbf21umapinfo-rc4/
The version of wad used was RC4.

All other methods of starting the wad work fine.
List of methods tried:
Using odalaunch.exe -> start single player
Starting odamex.exe and giving it doom 2 and whdah as files
Starting from cmd using `.\odamex.exe -iwad .\DOOM2.WAD -file .\WHDAH.wad`
Starting a server with whdah.wad and starting client with doom 2 and connecting to server
Loading up a netdemo for whdah.wad by dragging and dropping the netdemo in explorer

~~The mojibake is making me wonder if this is somehow related due to my extremely specific language settings.
Windows display language set to English (Finland)
Language for non-unicode programs Japanese (Japan)
Installed language packs are English (Finland), Japanese and Finnish~~
Edit: had a different person with (probably) less insane language settings test and it causes an error there as well, so it shouldn't be due to that.

bcahue:
This is also parsing a WAD as a DEH patch similar to #1814

PR #1882, opened 2026-08-07
[BUGFIX] Many WAD loading fixes

bcahue:
This PR, I tried to fix as many WAD loading issues as I could find. #1737 is fixed as is #1813 and #1814. The following changes were made:

1. Introduced a "Standalone IWAD test" that tests suspected standalone IWADs. These suspected IWADs pass the preliminary check (checking if it has the lumps an IWAD would have, like STDISK, PLAYPAL, etc) are checked to see if graphics exist for every sprite defined (no sprite would be a fatal error), and also a patch for every texture defined. All the IWADs I have found meet this criteria. If the suspected IWAD fails the test, it is reloaded with a "fallback" IWAD. Fixes corruption.wad and any other WAD with all the lumps that fit the shape of a standalone IWAD.

2. I discovered drag-n-dropped WADs were being run as dehacked patches. This is because dropped WADs weren't having their file type detected as what happens when specifying -wad, -bex and -deh. As such, the file was being added once to wadfiles, and twice to patchfiles, causing wads dropped onto the executable to be put into the DEH parser, with unexpected results. I fixed the root cause of this by having files as bare arguments go through a "classifier" to run as the resource Odamex believes it to be (wad/deh/bex) based on heuristics like file extension, but I also added a WAD check to the DEH parser to prevent users from doing things like `-deh mywad.wad` and having unpredictable Odamex behavior from that.

3. If more than 1 file was dropped into the executable, each of those files would run twice -- and four times in the case of DEH parsing. Fixed as part of the efforts above.

4. When having a `wad` command in a .cfg file (on a server but this would happen on the client as well), when the cfg is loaded, it is loaded before the command line options, which results in servers loading the correct WAD, then rebooting into DOOM2.WAD and not into the WAD that was specified in the config file. Rather than moving around the execution sequence (command line options then config file) I made the concept of "startup WADs" where WADs that are executed as part of the startup sequence (using command line flags of -file, +wad, etc and config commands like `wad`) are all put into a list, deduped by path (or just WAD name if that's all you specify) and executed in order (same order, cfg file then command line flags). This fixes all instances where Odamex/Odasrv boot up to the desired WADs, then it reboots and loses the PWADs that were loaded. Maplists aren't affected because they evaluate after the startup sequence.

5. G_LoadWad used to load WADs as case sensitive, when M_BaseFileSearchDir would search for wads using case-insensitive comparisons. The net effect of this was that this was causing an unnecessary WAD reboot when, for example, you'd do `wad doom2` it'd cause it to load twice. The fact that Odamex loads WADs as case insensitive is unchanged -- it'd be impossible to reconcile this with Windows users in the case of a linux server hosting case sensitive wads. This of course still leaves the "bug" open in Linux where it will simply load the first instance of a WAD it finds if you try to specify a case sensitive name for it, and 2 of them with the same case insensitive name exist in the same directory.

Also fixed a bug in G_LoadWad where, if `::wadfiles` was < 2, which is true when Odamex can't find any WADs to load, it would underflow and skip a WAD reboot and move forward with loading with no WADs causing Odamex to error with an obtuse error, rather than a friendly "buy an IWAD please" error.

I tried to make the server loading match up with the client loading exactly (save for the server only stuff like `lastmap`). I've tried this with many different configs, +wad and -file combinations, and can't get it to break, which was my goal to get WAD loading as bulletproof as possible.


common/d_main.cpp
```diff
@@ -1031,27 +1105,152 @@ static void AddCommandLineOptionFiles(OWantFiles& out, const std::string& option
...
+
+		std::string ext;
+		if (!M_ExtractFileExtension(arg, ext))
+		{
+			if (type != OFILE_WAD)
+				continue;
+		}
+		else
+		{
+			bool matches = false;
```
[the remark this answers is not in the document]
bcahue: Its not right? I can't make that const because its edited below.

bcahue:
This should be ready, the last clang-tidy suggestion isn't possible to make.

common/d_main.cpp
electricbrass: Could simplify to
```c++
if (std::ranges::none_of(exts, [&](const auto& fileext){ return iequals(ext, fileext); })
    continue;
```

client/src/d_main.cpp
```diff
@@ -892,14 +892,14 @@ void D_DoomMain()
 		const char* skipParams[] = {
 		    "+connect", "+demotest", "+map",      "+netplay",  "+playdemo",
 		    "-connect", "-file",     "-playdemo", "-timedemo", "-warp",
+		    "+wad"
```
electricbrass: Should this also have `-map` and `-wad`? And `-netplay` and `-netrecord`?

bcahue: I agree with `-map` `-wad` and `-netplay` but `-netrecord` is sticky as its checked every time CL_LoadMap is called to record (or split) a demo, so it seems like it'd be helpful to pick WADs here, as a `-netrecord` without a `+connect` doesn't really make sense, and if you specify `+connect` it'd skip the bootwindow anyway

electricbrass: Ah yea, makes sense

bcahue: Implemented

common/d_main.cpp
bcahue: Smart! Thank you! Implemented.

electricbrass (approved):

Issue #1416, opened 2025-11-10
[CHANGE] Support [hideauthorname] in MAPINFO TitlePatch


electricbrass:
**What is the existing behavior? Please describe.**
Currently Odamex recognizes `TitlePatch = "<patch>"`, but will error out on `TitlePatch = "<patch>",[hideauthorname]`.

**What is the desired behavior? Why are you requesting the change?**
The newer form with hideauthorname should be recognized. At the moment, Odamex doesn't actually make use of the author field anywhere, so even just throwing away the value would be fine. I don't know of any wads currently making use of this that are otherwise supported by Odamex, but it's a small enough change that I don't see much reason not to make it preemptively.

**Additional context**
Putting any non-zero value for [hideauthorname] should prevent the author name from being displayed in intermissions.

Issue #399, opened 2021-06-06
[FEATURE] Support for dynamic LANGUAGE tokens


Ch0wW:
**Is your feature request related to a problem? Please describe.**
Odamex's current implementation of `LANGUAGE` does not allow the possibility to add custom tokens that some PWADs are using (for instance: SIGIL). As a result, some ZMAPINFO tokens won't work properly because of that.

**Describe the solution you'd like to see**
Have a support for a dynamic support of `LANGUAGE` tokens.

PR #1889, opened 2026-08-09
[BUGFIX] Mapinfo Fixes

### LINKED ISSUE #808: [BUG] MAPINFO unable to use LANGUAGE lump entries for strings
**Describe the bug**
Odamex will use the literal string, when ZDoom will parse strings that start with "$" as LANGUAGE lump strings. As a result, Odamex loads the map name as string literals like "$MAP_NAME1"

Odamex 10.3.0 gb28e-1

**To Reproduce**
Load a wad with a MAPINFO lump only that uses LANGUAGE lump string tokens, such as 12Days.wad (https://www.doomworld.com/idgames/themes/xmas/12days-v4)

**Expected behavior**
For Odamex to check the LANGUAGE list if it encounters a MAPINFO string token that begins with $, and fail back to using the string literal if the LANGUAGE lump does not exist. Boom ports such as DSDA-Doom handle this.

N/A

**Additional context**
N/A

bcahue:
This is yet another PR that got away from me...

What started as a fix to IWADs displaying "Level X: " in console/automap instead of the proper level name, expanded into a cornucopia of additions to ZDoom MAPINFO that makes Odamex display things just a wee bit better.

As for the original fix, the reason that was displaying bad map names for IWADs was because the map name lookup was using an OLumpName, which cut off the characters at 9 chars, cutting off the leading space needed to transform it into the mapname we know and love in `G_InitLevelLocals()`.

On top of that, I added the following features and fixes:

1. `$Name` LANGUAGE lookups - Now ZDoom's old and new MAPINFO can use $ as a lookup for map name, cluster enter and exit text, episode names, skill names, and nightmare confirm text.
2. Created a universal helper for LANGUAGE lookups and "maybe lookups", the later being a lookup that will simply return the token given to it if it can't find the name in a lookup. I also went ahead and replaced every instance I could find using a language lookup to use these 
3. Emulated support for ZDoom's `\c` color escapes. ZDoom supports these in MAPINFO text. They're still stripped from player names or player chat messages.
4. Fixed a typo in `V_GetTextColor` where CR_RED was set as "e" instead of "g", seems like a copy/paste typo.
5. Implemented lookup for ZDoom MAPINFO Episode name.
6. Implemented showing map authors, including the original authors of the IWAD maps:
  a. For WADs that resolve UMAPINFO as their mapinfo provider, custom patches that replace title (CWILV*) patches make it so that authors are never shown, as they may be shown as part of the title patch.
  b. For WADs that resolve MAPINFO/ZMAPINFO as their mapinfo provider, patches provided with TitlePatch always show their author unless `hideauthorname` is specified as part of the definition.
  c. IWAD authors have been put into the MAPINFO lumps that come with odamex.wad. However, these authors are only shown if playing an IWAD map. If it's been replaced by a PWAD map, the author is not shown (unless it resolves an author from the PWAD MAPINFO lump.

Map authors are shown during intermission, in console and on the automap. There's several display options for map authors in the Automap menu, I suggest trying them all, but it defaults to "fade with map name."

Fixes #808, #399 and #1416

image

image

https://github.com/user-attachments/assets/e768629c-9bbe-4349-bb68-3b2f73e7b1f5

Test wad: https://wads.doomleague.org/mapinfoauthortest.zip


common/g_mapinfo.cpp
```diff
@@ -526,7 +535,7 @@ void MIType_$LumpName(OScanner& os, bool newStyleMapInfo, void* data, unsigned i
 	{
 		// It is possible to pass a DeHackEd string
 		// prefixed by a $.
-		const OLumpName s = GStrings(OStringToUpper(os.getToken().c_str() + 1));
```
electricbrass, line 529: This one is explicitly a lump name, it should probably stay OLumpName

common/g_mapinfo.cpp
```diff
@@ -551,7 +560,7 @@ void MIType_MusicLumpName(OScanner& os, bool newStyleMapInfo, void* data, unsign
 	{
 		// It is possible to pass a DeHackEd string
 		// prefixed by a $.
-		const OLumpName s = GStrings(OStringToUpper(musicname.c_str() + 1));
```
electricbrass, line 554: Same for this one

electricbrass (dismissed):

common/g_mapinfo.cpp
bcahue: Whoops got carried away

common/g_mapinfo.cpp
bcahue: Whoops got carried away

common/g_mapinfo.cpp
electricbrass: lol so did I back when turning these into OLumpNames

bcahue:
Converting to draft until I can get the sigil wads loading with Odamex. They give the following error: 
image

bcahue:
Oops I'm dumb, that's the error it gives when loading it with DOOM2.WAD. I still want to give it a better error than that.

bcahue:
Converting to draft again to implement `author` in intermission and support `hideauthorname` (#1416)

bcahue:
The rest of these suggestions would break conventions with the rest of the code in the files, like the menu and intermission.

bcahue:
I merged in #1904 here to resolve the rest of the m_options.cpp clang-tidy issues. Please merge that in before this.

drivingspectrum (approved):

PR #1890, opened 2026-08-09
[BUGFIX] SDL text input mode enabled all the time

### LINKED ISSUE #1884: [BUG] SDL text input mode is enabled all the time
**Describe the bug**
Currently, SDL_StartTextInput is called when starting the client, and then is left enabled.

Probably all of them ever.

**To Reproduce**
Describe the steps to reproduce this behavior.

**Expected behavior**
Text input mode should be enabled when opening the chat or console, or when selecting a save slot in the save menu or the player name in the player setup menu. When exiting from any of these, SDL_StopTextInput should be called.

**Additional context**
When the operating system has any special handling for text input modes, the current setup causes issues. The two issues examples I would like to see solved by this are:
1. On SteamOS game mode, the on-screen keyboard is brought up immediately and then never gets brought up for any of the actual text input fields.
2. On recent versions of KDE, text input mode causes certain keys to show a popup for alternate characters when held down, and not passing through the key event to the application.

electricbrass:
Fixes #1884

Now text input mode is enable when opening the chat, team chat, console, or full console, when editing a save name, and when editing the player name, and is disabled when leaving/finishing any of those.

### DIFF
--- a/client/sdl/i_input.cpp
+++ b/client/sdl/i_input.cpp
@@ -735,12 +735,14 @@ void I_StartTic()
 
 void I_EnableTextEntry()
 {
-	input_subsystem->enableTextEntry();
+	if (input_subsystem)
+		input_subsystem->enableTextEntry();
 }
 
 void I_DisableTextEntry()
 {
-	input_subsystem->disableTextEntry();
+	if (input_subsystem)
+		input_subsystem->disableTextEntry();
 }
 
 // ============================================================================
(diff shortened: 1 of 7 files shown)


drivingspectrum:
Is this still being worked on?

electricbrass:
Yes, some parts of the input system assume text input is always active and I'm working on fixing that still.

PR #1895, opened 2026-08-10
[FEATURE] Nyan Doom widescreen patches

### LINKED ISSUE #1175: [FEATURE] Support Nyan Doom widescreen assets
Nyan Doom has a system for including widescreen and non-widescreen assets in a single wad. I don't believe it is very widespread (though Woof just added support for it). I don't know if it is worth adding at the moment but if more wads start to use it, it would be needed for loading their widescreen graphics.

Just not doing it

**Additional context**
Here's the [spec](https://github.com/andrikpowell/nyan-doom/blob/master/docs/ws.md)

electricbrass:
Resolves #1175


bcahue (approved):
I don't have a WAD to check this but the implementation code makes sense

PR #1897, opened 2026-08-11
Consolidate Win32 SDK manipulations into header file

CurrentResident:
This PR consolidates the WIN32 API and SDK macro manipulations into a "one true" **win32inc.h** header file, resolving a large number of duplicate macro definition warnings, e.g.:

```
[124/534] Building CXX object client\CMakeFiles\odamex.dir\__\common\i_crash_win32.cpp.obj
D:\Projects\odamex\common\win32inc.h(36): warning C4005: 'WIN32_LEAN_AND_MEAN': macro redefinition
D:\Projects\odamex\common\win32inc.h(36): note: 'WIN32_LEAN_AND_MEAN' previously declared on the command line
```

This involves:

- Removing the related macro definitions from the CMake files,
- adding a few more definitions/checks to **win32inc.h**,
- removing duplicated macros and checks from the source files, and
- forcing the inclusion of **win32inc.h** as a compiler command line option when building for Windows.

I'm flexible on the last bullet point, which was chosen for convenience and simplicity of implementation.  Arguably, adding explicit includes to each compilation unit is higher-quality because it would force inclusion traceability into every source file.  However of all the hidden side-effects buried in the codebase, this one felt like it carries enough of a value-add and low enough of a maintenance burden to come out in favor of taking the easy way out.  I could easily be pushed the other way though.

### DIFF
--- a/cmake/modules/OdamexTargetSettings.cmake
+++ b/cmake/modules/OdamexTargetSettings.cmake
@@ -32,14 +32,14 @@ function(odamex_target_settings _TARGET)
 
   target_compile_definitions("${_TARGET}" PRIVATE $<$<CONFIG:Debug>:ODAMEX_DEBUG>)
 
-  # The Win32 / Win SDK-related definitions and macros are controlled in a header file.
-  # Force that header file into our targets' compile options here.
   if (WIN32)
-      if (MSVC)
-        target_compile_options("${_TARGET}" PRIVATE /FI${CMAKE_SOURCE_DIR}/common/win32inc.h)
-      elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-        target_compile_options("${_TARGET}" PRIVATE --include=${CMAKE_SOURCE_DIR}/common/win32inc.h)
-      endif()
+      target_compile_definitions("${_TARGET}"
+          PRIVATE
+            _CRT_SECURE_NO_WARNINGS
+            WIN32_LEAN_AND_MEAN
+            NOMINMAX
+            NODRAWTEXT
+            )
   endif()
 
   if(MSVC)
--- a/cmake/modules/OdamexTargetSettings.cmake
+++ b/cmake/modules/OdamexTargetSettings.cmake
@@ -30,18 +30,24 @@ function(odamex_target_settings _TARGET)
     target_compile_definitions("${_TARGET}" PRIVATE UNIX)
   endif()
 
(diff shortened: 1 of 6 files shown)


bcahue (dismissed):

electricbrass:
Oops, didn't think about the macro redefinition issue when I added those to the cmake. I wonder if it would be better to keep those and remove the definitions from win32inc.h? I'm not a big fan of having the implicit include everywhere.

Also, this changes odalpapi to no longer work as a standalone library, with the removal of the include there. I don't know how much anyone cares about that but it is something to note.

CurrentResident:
I largely prefer that we keep the special knowledge of WIN32 SDK / API fixups and manipulations in one comprehensive place so that we don't have to go hunting to get a complete picture.  Between the header file and the CMake, the header file gives us the flexibility to deal with the greatest number of issues and do whatever version-based contortions and insert POSIX glue and such.  That's my reasoning for moving the definitions out of CMake and into the header.

> I'm not a big fan of having the implicit include everywhere.

If you prefer, I'm fine with removing the forced include and being explicit in the various files that need it.  If I do that, it will bring in a bunch additional files in this PR and potentially be an ongoing burden.  I might be tempted to even rename the file to help it not be overlooked in the future, maybe something like "OdamexWindows.h" ?  Anyway, please let me know your thoughts.

>  this changes odalpapi to no longer work as a standalone library,

Ah, I didn't realize that was a use case.  I'll fix it.

electricbrass:
I guess I'm thinking about this in how do we minimize the amount of extra macro/function/whatever definitions that are brought in. Including windows.h everywhere adds a lot, but defining a couple macros in cmake for the relatively few spots that use windows.h is then adding exactly 3. It makes the potential for conflicts worse rather than better in a lot of cases.

I was a little iffy on making the cmake change anyway so I do think reverting it is fine, especially since I forgot to consider that it would result in redefinition issues.

> If I do that, it will bring in a bunch additional files in this PR and potentially be an ongoing burden.

Why would that affect many more files? There's clearly a few where windows.h is indirectly included that were causing the issues seen the other day, but those should be very few files overall (and probably could be reduced by cleaning up our headers some).

CurrentResident:
I'm starting to go back on my first thought because it looks like we have indirect inclusions of windows.h somewhere.

CurrentResident:
Ha, I was writing that the moment your comment came in.

CurrentResident:
> Why would that affect many more files?

It wouldn't.  I was misremembering what this PR and the CMake one did...  good morning!  time for coffee!

electricbrass (approved):

Issue #1687, opened 2026-04-13
[BUG] Ad Mortem custom pistol animation not visible when spectating another player mid-animation


nikovaan:
**Describe the bug**
When spectating another player while the player was in the middle firing pistol that had a custom animation, no sprites for the weapon were rendered at all until the player stopped firing.

**Build that the bug occurred in**
https://github.com/odamex/odamex/actions/runs/23720461265

**To Reproduce**
Have not made attempts at reproducing, see the stream clip for reference.

**Expected behavior**
Custom animations showing just fine when spectating another player even if jumping into spectating in the middle of one.

**Screenshots, NetDemos, & Crash Dumps**
Stream clip: https://www.twitch.tv/gobolt/clip/LongCourteousSeahorseDatSheffy-U-SdKOBqYLo7_5yy

**Additional context**
Due to client crash the netdemo for this map was broken. See https://github.com/odamex/odamex/issues/1686 for context.
Wad link: https://www.dropbox.com/scl/fi/naa6f7hzgqoglcew6w53h/AD_MORTEM.wad?rlkey=qe9r1ohuwyq6o1a4w2eaqhaam&dl=0
Wad link valid until I need to free up Dropbox space, but I can re-upload it later if needed.

Issue #744, opened 2022-08-09
[BUG] Automatic coop spy perspective changes don't always show the followed player's HUD weapon sprite


Acts19quiz:
**Describe the bug**
Automatic changes to the player being viewed in Survival modes (Horde depicted), or Coop spy, don't always show the now-followed player's HUD weapon sprite. A minor display bug for sure, but also pretty common.

**Build that the bug occurred in**
v10.1.0 (g0a84-8300)

**To Reproduce**
Die in a Survival mode (Horde depicted). Sometimes, the weapon of the player being spied upon will show correctly, but sometimes it won't.

**Expected behavior**
For the weapon of the player that has been switched to in Coop spy to be shown and positioned correctly.

**Screenshots, NetDemos, & Crash Dumps**
Here is both [a NetDemo](https://github.com/odamex/odamex/files/9293546/Coop_spy_HUD.zip) and [video](https://user-images.githubusercontent.com/63941237/183727334-57da3d53-3606-437a-9dd5-c69e0d291129.mp4) depicting the problem. Required WADs for the NetDemo are DOOM2.WAD and odahorde_220530.wad. The NetDemo accurately depicts, from my point of view, what happened in live play. The bug occurs around the 5:20 mark.

**Additional context**
Sometimes, it seems like the correct weapon is shown, but positioned too far low (like it's been positioned at the bottom of the screen but isn't considering the status bar's placement).

bcahue:
I believe this is because on spynext the weapon sprite starts at 0y and corrects on bob only, so if you're in a firing state, the weapon won't show.

PR #1903, opened 2026-08-16
[BUGFIX] Spy PSprite Improvements

### LINKED ISSUE #735: [CHANGE] Don't play any screen wipe effects when time jumping in a netdemo
Currently, if you have a screen wipe effect turned on, it will play when you time skip forward or backward while playing a netdemo.

Do not play a screen wipe effect.

**Additional context**
Quality of life. When you're skipping through a lot of short bursts, it adds unnecessary time when a screen wipe occurs.

bcahue:
This fixes #735, #744, and #1687 but also improves the look of spying while both playing and within a netdemo. Players in a netdemo that were never spied on in game when they are spied during the demo have even less information and can manifest as players shooting projectiles not belonging to the weapon that they have.

The following has been fixed:

- Spying someone mid-burst drew their weapon lowered until they released their fire button. Fixed by not waiting to populate player's sx/sy.
- ~`netff`/`netrew` lowered everyone's weapon and re-raised it. Fixed by not resetting everyone's psprite on a snapshot load.~ This was fixed by #1819 but was a victim of a bad merge. This fix has been re-implemented.
- Spying anyone in a netdemo other than whoever the recorder was spying showed a default weapon that exhibited strange behavior. Fixed by adding a message that broadcasts when a player switches weapons.
- `netff`/`netrew` were forcing a wipe since they'd reload the level every time. It's a single line to suppress that.

I also added some crash/missing weapon fixes in CL_PlayerState.

I should also say that these changes will not take effect in old netdemos as there is not enough information there, this only affects new netdemos.


endless-r0ad:
just noting the second bullet point was already fixed in #1819, not sure why its not in protobreak

electricbrass:
And it looks like the color stuff that was removed in that PR is still here as well

bcahue:
Strange, let me remove it

drivingspectrum (dismissed):

client/src/cl_demo.cpp
```diff
@@ -1283,6 +1284,9 @@ bool NetDemo::readSnapshot(SnapshotVector::const_iterator snap)
 
 	readSnapshotData(snapbuf);
 	netdemotic = snap->ticnum - header.starting_gametic;
+
+	NoWipe = 2;
```
CurrentResident: Consider making this an enumeration or a function whose name makes it clear what the effect is.

client/src/cl_parse.cpp
```diff
@@ -1740,6 +1740,24 @@ void CL_PlayerMembers(const odaproto::svc::PlayerMembers* msg)
 		if (!p.spectator)
 			p.cheats = msg->cheats();
 	}
+
+	// Our own weapon is driven by prediction and the rollback state, so this is only
+	// here to tell us what everyone else is holding.
+	if ((flags & SVC_PM_WEAPON) && p.id != consoleplayer_id)
```
CurrentResident, line 1772: FYI - I intend to retire the PlayerMembers message because it conflates multiple unrelated interfaces, essentially becoming just a way to sneak data across and avoid creating messages that represent actual application logic intent.  Just want to give a heads up about that.

server/src/sv_main.cpp
```diff
@@ -3638,7 +3643,15 @@ static void SV_SendMonitoredInventoryChanges(player_t& player)
...
-		MSG_WriteSVC(player.client.messenger->ReliableBuf(), SVC_PlayerPsprites(player));
+		// The monitor only latches on a state number change, not every tic, so an idle
+		// weapon costs nothing here and a firing one is a handful of messages a second.
+		// Everyone gets them because a netdemo is only ever recorded from one client's
+		// stream - without this there is nothing to animate the weapon of a player that
+		// the recorder happened not to be spying.
+		const odaproto::svc::PlayerPsprites psprites = SVC_PlayerPsprites(player);
+
+		MSG_WriteSVC(player.client.messenger->ReliableBuf(), psprites);
+		MSG_BroadcastSVC(CLBUF_RELIABLE, psprites, player.id);
```
CurrentResident: I'm not sure how crazy I am about sending this unconditionally because it does scale with player count, and some wads have very many psprites and rapid animations, potentially eating up a bunch of budget that should be going to gameplay-relevant traffic.

odaproto/server.proto
```diff
@@ -257,6 +257,7 @@ message PlayerPsprites
 {
 	int32                 player_tic = 1;
 	repeated PspriteState psprites   = 2;   // psprite[psprite_type] = psprite
+	uint32                pid        = 3;   // whose psprites these are
```
CurrentResident: In the interests of conserving bandwidth, consider either declaring pid as `optional` or using a value of `0` to indicate that the data pertains to the receiving player.

client/src/cl_parse.cpp
```diff
@@ -2511,17 +2529,29 @@ void CL_PlayerState(const odaproto::svc::PlayerState* msg)
 		}
 	}
 
-	statenum_t stnum[NUMPSPRITES] = {S_NULL, S_NULL};
+	struct PspriteUpdate
+	{
+		statenum_t	statenum = S_NULL;
+		fixed_t		sx = 0;
+		fixed_t		sy = 0;
+		bool		positioned = false;
```
CurrentResident: This type uses tabs for both indentation and alignment - Per the standard, we should align using spaces.

CurrentResident (dismissed):

odaproto/server.proto
electricbrass: Does `optional` do anything in protobuf 3? Afaik all fields are implicitly optional

CurrentResident: Yeah, it results in `has_`... APIs for the field.  So it doesn't change the infrastructure, just the interfaces that are available to the caller.

CurrentResident: Honestly, thinking about it more, a value of `0` is probably the way to go here, because capnproto also has a zero-value on-the-wire optimization too.

client/src/cl_demo.cpp
bcahue: This is a global that indicates "how many tics to suppress wipes" since there's checks at the end and the beginning of every new tic to check if they need to begin a wipe. I will create a function to indicate this.

server/src/sv_main.cpp
bcahue: Whoops I didn't think about this before writing that -- in the future, server-side demos will provide all the psprite goodness for each player for everyone to playback correctly with no bandwidth penalty (maybe some disk space penalty 😋) but until then, we'll have to predict when a player is firing when we don't have their psprites -- likely using the players attack sprites since we send those.

We still broadcast every time a player switches pending or ready weapon, so as to not have the spy target in demo who was not observed in game to have the right weapon.

Provides a bit of best-of-both-worlds compromise, but server side demos are going to be needed to make it truly 1:1 with how gameplay looked.

odaproto/server.proto
bcahue: This will be no longer sent. I will revert it to only send to the player whose psprites these belong to as it was before. In order to get a player's psprites now, you either spy them or wait for server side demo support.

bcahue:
All suggestions have been implemented. I have a little test demo here so you can check out the psprite prediction -- at no point was the 2nd, non-consoleplayer spied so everything is local prediction based off player attack sprites.

https://www.dropbox.com/scl/fi/m96k3jklw19myy15guoxh/psprite-test-demo.zip?rlkey=aup7e2fdfdhufnzrb0yfhl2ww&st=xlsdrasm&dl=0

common/p_pspr.cpp
```diff
@@ -445,6 +462,44 @@ bool P_CheckSwitchWeapon(const player_t& player, weapontype_t weapon)
...
 }
 
+//
+// P_WeaponIsRemotelyDriven
+//
+// True when we are a client animating a player whose inputs we never see.
+//
+bool P_WeaponIsRemotelyDriven(const player_t& player)
+{
+	return !serverside && player.id != consoleplayer_id;
```
endless-r0ad, line 473: Probably doesnt matter but should this also check displayplayer_id since those sprites are set by svc_playerstate during spying?

bcahue: Good catch, I will fix this. I still think predicting puff should stay consoleplayer, as under ideal network conditions, the puffs and gun firing would be the same tic, compared to a lag time if they were predicted.

bcahue: After looking at this, changing to displayplayer_id does break this, specifically because its supposed to solve for "do we have this player's inputs" instead of "are we displaying this player," and it would also cause the demo firing prediction to run on the consoleplayer.

Psprite updates are inferred via  `SVC_PlayerState` to whoever is spying on the player, however, this does mean that the prediction is running on the tic the weapon fired on spied players. I can rename the function and try to stop that if we've already received an `SVC_PlayerState`.

bcahue: Done. Confirmed with clumsy that this prediction looks good if we miss packets while spying a player. Also looks great in demos when not spying the player.

server/src/sv_main.cpp
CurrentResident: I'm onboard with this.  Thanks!

CurrentResident (approved):
Awesome, great job!  I'm thrilled about the best-of-both-worlds compromise and how you made the most of prediction!

PR #1908, opened 2026-08-19
Add sender tic and receiver-tic-of-validity to basic packet header

CurrentResident:
This PR does the following:

- Plumbing change:  Adds tic / timing information to the packet header sufficient for performing rollbacks at either end.
- Fix best-effort-only packet sequence reception:  Due to an oversight in my last PR, the sequence number was negated for best-effort-only packets, causing rapid mobj stale-out and loss of credibility on the client side.  This led to substantial desyncs.
- Clarity:  Update `OdaMessenger::NextPacket` to make it obvious that high-priority packets are received first when multiple types of packets arrive.  This was always true due to how the function was being called, but now it's enforced inside the function.

I intend to remove the client-tic number fields from the various rollback-inducing messages that have them and shift over to the received "destination Tic" from the corresponding packet headers.


electricbrass:
~~If you do `.value().sequence` instead of `->sequence` it will throw an exception instead of being undefined behavior when its an empty optional. Not sure if it will satisfy clang-tidy or not~~

Edit: Just checked the docs, clang-tidy treats them the same

common/SequenceReceiver.cpp
```diff
@@ -94,7 +93,7 @@ int SequenceReceiver::NextPacket(buf_t& io_bufferRef)
 				auto& entry = iter->second.bestEffort.front();
 				io_bufferRef.swap(entry.buf);
 				iter->second.bestEffort.pop_front();
-				fetchedPacketSequenceNumber = m_currentSequence;
+				fetchedHeader = entry.header;
```
bcahue: Isn't this already popped and destroyed?

common/OdaMessenger.cpp
```diff
@@ -456,17 +477,21 @@ int OdaMessenger::HandleRetransmissions(int i_currentTic, const netadr_t& i_dest
...
 			}
-			m_noncontiguousRetransmitCount += previousPacketSeq != sendQueueEntry->sequence - 1 ? 1 : 0;
-			previousPacketSeq = sendQueueEntry->sequence;
+			m_noncontiguousRetransmitCount += previousPacketSeq != sendQueueEntry->header.sequence - 1 ? 1 : 0;
+			previousPacketSeq = sendQueueEntry->header.sequence;
 
 			sendQueueEntry->lastRetransmitTic = i_currentTic;
-			const int resendSize = static_cast<int>(m_packet.ReSend(sendQueueEntry->sequence, sendQueueEntry->buf, i_dest));
+			const int resendSize = static_cast<int>(m_packet.ReSend(sendQueueEntry->header.originatorTic,
+			                                                        sendQueueEntry->header.destinationTic,
```
bcahue, line 491: Is header.destinationTic ever assigned? Or is it not needed yet?

bcahue (changes requested):

common/SequenceReceiver.cpp
CurrentResident: Oh shoot, yes it sure is.  Good find, fixing.

common/OdaMessenger.cpp
CurrentResident: Yes, in `OdaMessenger::SendAll`.  All the packet `Send...` calls require it as an argument, and we pass in `m_destinationTic`, which is ultimately set by the client and the server via the `SetDestinationTic` method.  The packet class then sets the destinationTic in the header along with the originatorTic and other info.

As for the need, I intend to replace the pile of `client_tic` fields in various messages with a call to `GetCurrentReceivedLocalTic()`.  Ultimately I did this because as more things get rollback support, we can just use this info for free instead of having to add more repeated fields, and it also provides an actual time point that we can use to compare mobj state validity from server to predicted.  Right now all we have is packet sequence number, which is okay, but only of limited precision because the sequence number rolls only with Reliable packets, and it's possible for multiple best-effort mobj updates to arrive between reliable updates.  In those cases, there's no way to know for certain which of the two is actually more current ("first was earlier" is a reasonable assumption but not necessarily always true), and there's a similar (but more realisitic) situation between best-effort and reliable packets.

common/SequenceReceiver.cpp
CurrentResident: FWIW, the code that results in this block going active is disabled in the Messenger, but it's still being exercised by the unit tester.  Looking more closely, I suspect the reason this landmine didn't blow up on is because the memory block being referenced is still owned by the container and the header's destructor is trivial, so the old values were still there unchanged.

bcahue: Looks good now!

common/OdaMessenger.cpp
bcahue: Sounds good to me.

bcahue (approved):
The rest looks good to me, its hard to get an overall feel for everything netcode and how these changes affect it but for this PR, everything's looking good.

PR #1909, opened 2026-08-20
[CHANGE] Type-safe MAPINFO Container

### LINKED ISSUE #1639: [CHANGE] Update the mapinfo setter mechanism to be type-safe
**What is the existing behavior?**

The mapinfo setters use `void*` data pointers and special-knowledge static casting to get data into their associated application variables/objects.  That association is established by just providing a setter function pointer and the data address in one shot.  (See the various `mapInfoDataContainer` hand-coded assignments,)

The data's type information is lost because its address is accepted as just a `void*`, and the setter function has no choice but to just assume that the hand-coded association is correct and simply cast the pointer to a specific type and start working with the data it finds there.  If those types disagree, then the compiler generates incorrect code without warning or error, and out-of-bounds accesses ensue.

Change the mapinfo setter infrastructure so that a mismatch between setter and data type result in a compilation error, so that out-of-bounds accesses cannot happen in this area.

**Additional context**

This addresses the root cause of https://github.com/odamex/odamex/pull/1638

bcahue:
I started this to resolve #1639 but did not go as far as intended to try to meet @electricbrass in the middle with her enum class changes for bit flags.

The main intention of this was to resolve the fact that the struct MapInfoData had a `void*` and 2 `flags` fields that needed to be cast to its intended type every time. This, of course, can cause issues that don't get noticed until run time, as shown with the fix needed in #1638.

To resolve this, we make this struct have nothing but a name and a `std::function<void(OScanner&, bool)>` with 4 templated constructors that all cover how `mapInfoDataContainer` is initialized today:

```cpp
{ "lightning" }                                               // recognized and no-op'd
{ "border", MIType_Border }                                   // writes global state
{ "levelnum", MIType_Int, ref.levelnum }                      // bound to a field
{ "doublesky", MIType_SetFlag, ref.flags, LEVEL_DOUBLESKY }   // sets a bit in a bit set
```

The templates resolve the type at compile time, and throw an error if there's any mismatched type. This is much preferred to a user finding this issue down the road. No longer needed are the ~50 `static_cast`s to cast the data into its assumed type.

I also changed every `MIType*` function to consume this type if necessary, and have them each output via reference to a type that you'd expect for the parsing function.

Here's some more improvements:

1. In `MIType_SoundName`, `strncpy` no longer copies a max sndinfo token size every time and instead copies the size of the token.
2. In `MIType_Pages`, we now specify the amount of pages in a template variable, so we no longer have to use `flags` as a weird workaround to indicate we want 3 pages or not.
3. `MIType_Bool` and `MIType_SetInt` have been merged into `MIType_Assign<T>`. T is type-checked against the `&out` and `value` variables so an improper type here will never compile.
4. `MIType_Color` no longer modifies raw bytes in a uint8_t in order to record the colors.
5. `MIType_InterLumpName` no longer casts into a `std::pair<OLumpName*, OLumpName*>*`.
6. `MIType_Sky` now sends the sky number (1 for regular sky, 2 for double sky) as a template variable instead of adding it to `flags`.
7. For the flag setters (`MIType_SetFlag`, `MIType_CompatFlag`, and `MIType_SCFlags`), I tried to make it as generic as possible awaiting the arrival of type safe flags. They're a functor, and I did it this way because it allows me to not set the type twice and instead depend on the same type checking everything else does. e.g.:
`{ "islobby", MIType_SetFlag, ref.flags, LEVEL_LOBBYSPECIAL }`
instead of:
`{ "islobby", MIType_SetFlag<levelFlags_t>, ref.flags, LEVEL_LOBBYSPECIAL }`

I also created a new "metadata" flag type for levels, for information about how data was acquired, in order to properly set the "show author" rules. The reason I did this was because that information isn't set by MAPINFO parsing itself, and thus should be separate. I also fixed levelflags2 when enabling bitflag operators to return LEVEL2_HIDEAUTHORNAME, which is a MAPINFO setting.

Please review as this is my first real foray into C++ templates. 😄


common/g_mapinfo.cpp
```diff
@@ -1201,91 +1193,87 @@ template <>
...
+			{ "map07special", MIType_Map07Special, ref.bossactions },
+			{ "baronspecial", MIType_Special<MT_BRUISER>, ref.bossactions },
+			{ "cyberdemonspecial", MIType_Special<MT_CYBORG>, ref.bossactions },
+			{ "spidermastermindspecial", MIType_Special<MT_SPIDER>, ref.bossactions },
+			{ "e1m8special", MIType_Special<MT_NULL, MF3_E1M8BOSS>, ref.bossactions },
+			{ "e2m8special", MIType_Special<MT_NULL, MF3_E2M8BOSS>, ref.bossactions },
+			{ "e3m8special", MIType_Special<MT_NULL, MF3_E3M8BOSS>, ref.bossactions },
+			{ "e4m6special", MIType_Special<MT_NULL, MF3_E4M6BOSS>, ref.bossactions },
+			{ "e4m8special", MIType_Special<MT_NULL, MF3_E4M8BOSS>, ref.bossactions },
+			{ "specialaction_exitlevel", MIType_SpecialAction<11>, ref.bossactions },
```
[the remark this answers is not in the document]
electricbrass: Is it finally time to introduce an enum for doom format line specials? This was just being talked about the other day in the DCPC discord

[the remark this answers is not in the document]
bcahue: I'll see what I can do here

common/g_level.h
```diff
@@ -99,6 +99,14 @@ constexpr static levelFlags_t LEVEL2_COMPAT_CROSSDROPOFF = BIT(18);
 struct acsdefered_t;
 class FBehavior;
 
+// Sector tags the boss actions operate on.
+constexpr int16_t BOSSACTION_TAG = 666;
```
electricbrass: Should be `inline constexpr` when in a header

common/g_mapinfo.cpp
```diff
@@ -1002,9 +991,15 @@ void MIType_SpecialAction(OScanner& os, bool newStyleMapInfo, void* data,
 	action.tag = TAG;
 }
 
+// Ensure the bossaction line specials are valid.
+template <auto SPECIAL, int16_t TAG = 0>
+    requires std::same_as<decltype(SPECIAL), doomLineSpecial_t> ||
+             std::same_as<decltype(SPECIAL), boomLineSpecial_t>
+constexpr auto MIType_LineSpecialAction =
+    MIType_SpecialAction<lineSpecialValue(SPECIAL), TAG>; // NOLINT(modernize-avoid-c-style-cast)
```
electricbrass, line 1018: I assume the NOLINT is suppressing a false positive here? Could we add a comment explaining that

client/src/cl_parse.cpp
```diff
@@ -1689,7 +1689,7 @@ void CL_Print(const odaproto::svc::Print* msg)
 	if (show_messages)
 	{
 		if (level == PRINT_CHAT || level == PRINT_SERVERCHAT)
-			S_Sound(CHAN_INTERFACE, gameinfo.chatSound, 1, ATTN_NONE);
+			S_Sound(CHAN_INTERFACE, gameinfo.chatSound.data(), 1, ATTN_NONE);
```
electricbrass, line 1720: Probably for after this PR, but I wonder if these should be std::string instead of std::array. But that would mean we probably would want to get rid of MAX_SNDNAME entirely, or maybe it would be worth introducing a fix-sized string type kinda like OLumpName but templated on the buffer length.

common/g_mapinfo.cpp
```diff
@@ -1200,92 +1213,89 @@ struct MapInfoDataSetter
...
+			{ "secretnext", MIType_MapName, ref.secretmap },
+			{ "secret", MIType_MapName, ref.secretmap },
+			{ "cluster", MIType_Cluster, ref.cluster },
+			{ "sky1", MIType_Sky<1>, ref },
+			{ "sky2", MIType_Sky<2>, ref },
+			{ "fade", MIType_Color, ref.fadeto_color },
+			{ "outsidefog", MIType_Color, ref.outsidefog_color },
+			{ "titlepatch", MIType_TitlePatch, ref },
+			{ "music", MIType_MusicLumpName, ref.music },
+			{ "nointermission", MIType_SetFlag, ref.flags, LEVEL_NOINTERMISSION },
```
electricbrass, line 1252: If `OFlags::data` is now unused, lets delete it. It was only there as a temporary measure until this PR.

common/g_mapinfo.cpp
```diff
@@ -1002,9 +991,15 @@ void MIType_SpecialAction(OScanner& os, bool newStyleMapInfo, void* data,
 	action.tag = TAG;
 }
 
+// Ensure the bossaction line specials are valid.
+template <auto SPECIAL, int16_t TAG = 0>
+    requires std::same_as<decltype(SPECIAL), doomLineSpecial_t> ||
```
electricbrass: Should we create a concept for this constraint? I could see it being repeated elsewhere later.

common/p_lnspec.h
```diff
@@ -381,6 +381,300 @@ typedef enum {
...
+	ST_TransferSky = 271,
+	ST_TransferSkyFlipped = 272,
+};
+
+constexpr int16_t lineSpecialValue(doomLineSpecial_t special)
+{
+	return static_cast<int16_t>(special);
+}
+
+constexpr int16_t lineSpecialValue(boomLineSpecial_t special)
```
electricbrass: If the concept mentioned in one of my other comments is created, I would probably combine this into a single definition like
```c++
constexpr auto lineSpecialValue(const DoomFormatLineSpecial auto special)
{
    return OUtil::to_underlying(special);
}
```

common/g_mapinfo.cpp
```diff
@@ -693,53 +667,80 @@ void MIType_Sky(OScanner& os, bool newStyleMapInfo, void* data, unsigned int fla
 }
 
 // Sets a flag
-void MIType_SetFlag(OScanner& os, bool newStyleMapInfo, void* data, unsigned int flags,
-                    unsigned int flags2)
+struct MIType_SetFlag_t
 {
-	*static_cast<uint32_t*>(data) |= flags;
-}
+	template <typename F, typename B>
```
electricbrass, line 689: Not sure I understand why these ones aren't just functions?

common/g_mapinfo.cpp
```diff
@@ -1200,92 +1213,89 @@ struct MapInfoDataSetter
...
+			{ "spidermastermindspecial", MIType_Special<MT_SPIDER>, ref.bossactions },
+			{ "e1m8special", MIType_Special<MT_NULL, MF3_E1M8BOSS>, ref.bossactions },
+			{ "e2m8special", MIType_Special<MT_NULL, MF3_E2M8BOSS>, ref.bossactions },
+			{ "e3m8special", MIType_Special<MT_NULL, MF3_E3M8BOSS>, ref.bossactions },
+			{ "e4m6special", MIType_Special<MT_NULL, MF3_E4M6BOSS>, ref.bossactions },
+			{ "e4m8special", MIType_Special<MT_NULL, MF3_E4M8BOSS>, ref.bossactions },
+			{ "specialaction_exitlevel", MIType_LineSpecialAction<S1_Exit_Normal>, ref.bossactions },
+			{ "specialaction_opendoor", MIType_LineSpecialAction<W1_Door_OpenFast, BOSSACTION_TAG>, ref.bossactions },
+			{ "specialaction_lowerfloor", MIType_LineSpecialAction<S1_Floor_LowerToLowest, BOSSACTION_TAG>, ref.bossactions },
+			{ "specialaction_killmonsters", MIType_SpecialAction<BOSSACTION_MASSACRE>,ref.bossactions },
```
electricbrass: This line doesn't have a space after one of the commas, unlike the other lines.

common/g_mapinfo.cpp
```diff
@@ -1154,29 +1146,50 @@ void MIType_MapKey(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
...
+	{
+	}
+
+	// A setter bound to the field it writes.
+	template <typename T, typename... Args>
+	MapInfoData(const char* _name, void (*_fn)(OScanner&, bool, T&, Args...), T& _ref,
+	            std::type_identity_t<Args>... _args)
+	    : name(_name),
+	      fn([_fn, &_ref, _bound = std::tuple<Args...>(_args...)](
+	             OScanner& os, bool newStyleMapInfo) {
```
electricbrass: You can simplify the lambda
```c++
	      fn([_fn, &_ref, &_args...](
	             OScanner& os, bool newStyleMapInfo) {
		      std::invoke(_fn, os, newStyleMapInfo, _ref, _args...);
	      })
```

common/g_mapinfo.cpp
```diff
@@ -1154,29 +1146,50 @@ void MIType_MapKey(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
...
+
+	// A setter that writes to global state rather than to the block being parsed.
+	MapInfoData(const char* _name, void (*_fn)(OScanner&, bool)) : name(_name), fn(_fn)
+	{
+	}
+
+	// A setter written as a generic callable, for setters that accept more than one
+	// destination type (the flag setters).
+	template <typename Fn, typename T, typename... Args>
+	    requires(!std::is_pointer_v<Fn>) &&
```
electricbrass: There's some weird tabs vs spaces stuff going on in both of these constructors.

common/g_mapinfo.cpp
electricbrass: Same applies to the other constructor

common/g_mapinfo.cpp
```diff
@@ -1154,29 +1146,50 @@ void MIType_MapKey(OScanner& os, bool newStyleMapInfo, void* data, unsigned int
...
+	                std::invocable<const Fn&, OScanner&, bool, T&, Args...>
+	MapInfoData(const char* _name, Fn _fn, T& _ref, Args... _args)
```
electricbrass: I feel like 2 constructors aren't required here. Just one that accepts any type of callable should be doable. I don't know off the top of my head the best way to do it but I can mess around with it a bit and see if I come up with anything.

common/g_mapinfo.cpp
electricbrass: Would be even nicer if we had `std::bind_back`, but alas, that is C++23

electricbrass:
Oh also, I might add a call operator to MapInfoData, and move the check for an empty std::function inside of that, instead of relying on ParseMapInfoLower doing the check. Probably would then also make the fn and name members private, with getters instead, to prevent bypassing the check or accidentally modifying them.

common/g_level.h
```diff
@@ -106,19 +106,38 @@ enum class levelflags2_t : uint32_t
...
 
 using enum levelflags2_t;
-constexpr levelflags2_t enable_bitflag_operators(levelflags2_t) { return LEVEL2_COMPAT_CROSSDROPOFF; };
+constexpr levelflags2_t enable_bitflag_operators(levelflags2_t) { return LEVEL2_HIDEAUTHORNAME; };
 using LevelFlags2 = OFlags<levelflags2_t>;
 
+// Map metadata to determine downstream behavior
+enum class metadataflags_t : uint32_t
+{
+	METADATA_AUTHORFROMPWAD     = BIT(0),
```
electricbrass: When making a new enum, we might as well just not include the name in the flags, and get rid of the `using enum`.

common/g_mapinfo.cpp
bcahue: I explained it a bit at the top, essentially it's the only way I can make it so that the type doesn't have to be explicit. If I were to make it a templated function, it would look like this:

`{ "islobby", MIType_SetFlag, ref.flags, LEVEL_LOBBYSPECIAL }`
instead of:
`{ "islobby", MIType_SetFlag<levelFlags_t>, ref.flags, LEVEL_LOBBYSPECIAL }`

It's still doing type checking but now it infers types by the flags, and not needlessly in the function as well.

bcahue: I'll add a comment that explains this

common/g_mapinfo.cpp
bcahue: Yes it was, and I explained what it was seeing in a comment.

common/g_mapinfo.cpp
bcahue: When implemented as shown, my tests started failing for flag setters. `_args` in the lambda needs to be copied rather than passed by reference, since when its time to `std::invoke` the template constructors parameters are gone.

The rest of it is good though, thank you.

common/g_mapinfo.cpp
bcahue: This was doable but I had to convert MIType_InterLumpName to a generic in order to not be the only mapinfo function's signature to have 2 variables passed by reference.

common/g_mapinfo.cpp
electricbrass: Interesting, when I tried it, it works just fine. Plus the copy was breaking enterpic/exitpic, because it was ending up writing intermission scripts to a local copy. Removing the tuple allowed the type system to catch this and is what made it have to be a reference when I tried.

electricbrass: Oh, it's because there was another change I made that I forgot about in the comment: All of args was being passed by value to the constructor, which was an issue. Changed it to accept Args&&... and std:: forward them on. Then it worked. That's also I assume why you had to change interlumpname to not have a second reference argument.

electricbrass: > Removing the tuple allowed the type system to catch this and is what made it have to be a reference when I tried.

Okay I see, the tuple is completely unnecessary in the more generic constructor, but in the one that takes a function pointer, the type of Args comes from the function pointer and so the tuple properly uses reference types when needed because of that. It is annoying that that makes it so that you have to use a function pointer for it to be able to have any of Args be reference parameters then...

electricbrass: So leaving the function pointer constructor pretty much as is probably is the way to go then

electricbrass: Wait, nevermind.

Here's a constructor that works for everything, no changes needed anywhere else.

```c++
	template <typename Fn, typename T, typename... Args>
	    requires std::invocable<const Fn&, OScanner&, bool, T&, Args...>
	MapInfoData(const char* _name, Fn&& _fn, T& _ref, Args&&... _args)
	    : name(_name), fn([_fn = std::forward<Fn>(_fn), &_ref, _bound = std::tuple<Args...>(std::forward<Args>(_args)...)](
	                          OScanner& os, bool newStyleMapInfo) {
		      std::apply([&](auto&... a) { _fn(os, newStyleMapInfo, _ref, a...); },
		                 _bound);
	      })
	{
	}
```

electricbrass: It doesn't have the lifetime issues of my original suggestion and it still works for functions that take multiple reference parameters like MIType_InterLumpName

common/p_lnspec.h
```diff
@@ -24,6 +24,9 @@
 #pragma once
 
 #include "r_defs.h"
+#include "flags.h"
```
electricbrass: Accidental include?

common/g_mapinfo.cpp
electricbrass: Oops, hadn't re-read the PR description when making this comment. For some reason I was thinking there was some way to make this work without them needing to be function objects like that, but I don't think there is.

common/p_lnspec.h
electricbrass: Oh, for to_underlying. A direct include of util.h would probably be better

common/g_mapinfo.cpp
bcahue: Implemented, but I kept MIType_InterLumpName in its current form -- but this keeps us much more flexible when we want to implement more of Z/MAPINFO in the future. Very nice, thank you!

common/p_lnspec.h
bcahue: Done.

common/g_mapinfo.cpp
```diff
@@ -497,11 +497,23 @@ void MIType_LumpName(OScanner& os, bool newStyleMapInfo, OLumpName& out)
...
+enum class interLump_t
+{
+	Enter,
+	Exit,
+};
+
 // Handle lump names that can also be intermission scripts
-void MIType_InterLumpName(OScanner& os, bool newStyleMapInfo, OLumpName& pic,
-                          OLumpName& script)
+template <interLump_t SLOT>
```
electricbrass: I think this should probably be reverted now. At least I thought the old version of this function was more readable, and we can go back to only passing in the vars this function needs instead of the whole level info.

electricbrass: Oh didn't see your comment about keeping it the same

bcahue: I reverted it anyway, passing in the entire level locals is kinda gross lol

electricbrass (approved):

PR #1910, opened 2026-08-20
[CHANGE] Type-safe bitflags

electricbrass:
To help avoid bugs when using ints and enums for flags, this PR introduces a new type: `OFlags`. This is a fully type-safe, constexpr, and zero-overhead bitflag container with no unsafe implicit conversions. Due to limitations of C++ prior to C++26, `OFlags` can only be templated on an `enum class E` for which the function `consteval E enable_bitflag_operators(E)` exists. This function should return the enum variant corresponding to the highest bit. Unscoped C enums cannot be used with `OFlags`.

Because conversion to and from raw integers is required for DeHackEd, binary map formats, and netcode, `OFlags` provides `to_int()` and `unsafe_from_int()` operations, and can also be `static_cast`ed to the underlying type. `unsafe_from_int` does *not* perform implicit conversions from other integer types. It requires a value of the exact underlying type.

Due to MAPINFO touching level and actor flags, there will be conflicts with #1909, but I tried to keep these as minimal as possible. To facilitate this there is currently an additional unsafe `OFlags::data()` which gives direct access to the underlying integer storage. As soon as the work of these 2 PRs are merged, this should be removed in favor of using the type-safe API.

There's also some additional utility stuff related to enums:
- `OUtil::is_scoped_enum`: backport of C++23's `std::is_unscoped_enum`
- `OUtil::Enum`, `OUtil::EnumClass`, and `OUtil::CEnum`: concepts to constrain types to `enum`/`enum class`, `enum class` only, and `enum` only.
- `OUtil::to_underlying`: backport of C++23's `std::to_underlying`

Some other OUtil stuff is now constexpr as well.


common/actorflags.h
```diff
@@ -0,0 +1,208 @@
...
+	MF3_E3M8BOSS		= BIT(14),	// is an E3M8 boss
+	MF3_E4M6BOSS		= BIT(15),	// is an E4M6 boss
+	MF3_E4M8BOSS		= BIT(16),	// is an E4M8 boss
+									// BIT 17 is MF2_RIP -- RESERVED
+	MF3_FULLVOLSOUNDS	= BIT(18),	// full volume see / death sound
+};
+
+using enum mobjflag3_t;
+consteval mobjflag3_t enable_bitflag_operators(mobjflag3_t) { return MF3_FULLVOLSOUNDS; }
+using ActorFlags3 = OFlags<mobjflag3_t>;
```
CurrentResident: Having to supply `enable_bitflag_operators` with the highest-valued enumeral seems like a potential maintenance burden.  I can easily see someone adding a flag and forgetting to update this line.  Is there any way to avoid the need to do this?

Aside from that, I presume this is mitigated by a compile error that results from using an enumeral that is greater than this value...  If so, would it be possible to add a static assert or something that explains exactly to the developer what the solution should be?

common/actorflags.h
```diff
@@ -0,0 +1,208 @@
...
+	MF3_MAP07BOSS1		= BIT(10),	// is a MAP07 boss type 1 (666)
+	MF3_MAP07BOSS2		= BIT(11),	// is a MAP07 boss type 2 (667)
+	MF3_E1M8BOSS		= BIT(12),	// is an E1M8 boss
+	MF3_E2M8BOSS		= BIT(13),	// is an E1M8 boss
+	MF3_E3M8BOSS		= BIT(14),	// is an E3M8 boss
+	MF3_E4M6BOSS		= BIT(15),	// is an E4M6 boss
+	MF3_E4M8BOSS		= BIT(16),	// is an E4M8 boss
+									// BIT 17 is MF2_RIP -- RESERVED
+	MF3_FULLVOLSOUNDS	= BIT(18),	// full volume see / death sound
+};
```
CurrentResident, line 169: FYI - this section uses tabs for both indentation and alignment.

client/src/cl_parse.cpp
```diff
@@ -774,17 +774,17 @@ static void CL_SpawnMobj(const odaproto::svc::SpawnMobj* msg)
 
 	if (msg->spawn_flags() & SVC_SM_FLAGS)
 	{
-		mo->flags = msg->current().flags();
+		mo->flags = ActorFlags1::unsafe_from_int(msg->current().flags());
```
CurrentResident, line 793: One of the changes I've been considering is autogenerating protobuf / capnproto schema enumerations from our codebase.  It's down the road a little ways yet, but do you have a facility (or a suggestion for one) to assign enumerals back and forth from syntactically unrelated enumerations?  (...Besides intermediate integer casts, that is?)

client/src/m_menu.h
```diff
@@ -161,9 +161,8 @@ struct menuitem_t {
 typedef struct menu_s {
 	OLumpName		title;
 	int				lastOn;
-	int				numitems;
 	int				indent;
-	menuitem_t	   *items;
+	std::span<menuitem_t> items;
```
CurrentResident, line 169: As we touch types over time, we should consider changing their alignment tabs to spaces.  Right now this field uses a space among a collection of others that are tabbed.

common/c_effect.h
```diff
@@ -43,13 +43,13 @@
 #define FX_WHITEFOUNTAIN	0x00070000
 
 
-struct particle_s;
+struct particle_t;
```
CurrentResident, line 46: Yay!  Happy to see more `_s` become `_t`.

common/d_dehacked.cpp
```diff
@@ -377,36 +380,46 @@ static constexpr CodePtr CodePtrs[] = {
...
-	int32_t mobjinfo_t::* flags;
 	int32_t dehBit;
-	int32_t internalBit;
+	void (*setter)(mobjinfo_t&);
 } mbf21flagtranslation[] = {
 	// flags2
-	{ "LOGRAV",         &mobjinfo_t::flags2, BIT(0),  MF2_LOGRAV },
-	{ "BOSS",           &mobjinfo_t::flags2, BIT(9),  MF2_BOSS },
-	{ "RIP",            &mobjinfo_t::flags2, BIT(17), MF2_RIP },
+	{ "LOGRAV",         BIT(0),  [](mobjinfo_t& info) { info.flags2 |= MF2_LOGRAV; } },
```
CurrentResident: Is there any way to build this table without having to duplicate the enumeral's integral value here, and potentially have to keep it in sync across multiple files?

-----

I have a similar question about this and the `ClearMBF21Bits` function below.  It would be nice if all the special knowledge needed for these capabilities could be gathered up into a single table or definition.

common/d_dehacked.cpp
```diff
@@ -2656,41 +2654,44 @@ bool CheckIfDehActorDefined(const mobjtype_t mobjtype)
...
-	    mobj.altspeed == NO_ALTSPEED &&
-		mobj.infighting_group == IG_DEFAULT &&
-		mobj.projectile_group == PG_DEFAULT &&
-		mobj.splash_group == SG_DEFAULT &&
-		mobj.ripsound == NULL &&
-		mobj.meleerange == (64 * FRACUNIT) &&
-		mobj.droppeditem == MT_NULL)
+		mobj.spawnhealth == defaults.spawnhealth &&
+		mobj.gibhealth == defaults.gibhealth &&
+		mobj.seestate == defaults.seestate &&
```
CurrentResident, line 2668: NICE!  Love to see reduced reliance on special knowledge literals.

common/farchive.h
```diff
@@ -158,19 +159,23 @@ class FArchive
...
 
-	template <typename EnumeratedType>
-	requires std::is_enum_v<EnumeratedType>
-	FArchive& operator<< (EnumeratedType value)
+	FArchive& operator<< (const OUtil::Enum auto value)
 	{
-		*this << static_cast<std::underlying_type_t<EnumeratedType> >(value);
+		*this << OUtil::to_underlying(value);
+		return *this;
+	}
```
CurrentResident, line 175: How do we know that, say, naive enumerations go through this overload instead of the integral one?  I presume the language favors the implicit conversion of naive enumerations to `OUtil::Enum` rather than integers, but I'm not sure that I understand why.  (I think I would personally have a preference for seeing requirement constraints enforcing it because that would be obvious, as I suspect it would be for other developers as well, but I might just be missing something here)

common/p_enemy.cpp
```diff
@@ -1184,7 +1184,7 @@ AActor::AActorPtr SpawnHelper(const MapThing SpawnPoint, mobjtype_t SpawnType, c
 			SV_SpawnMobj(mo);
 
 			// Spawn a teleport fog if it's not an ambush.
-			if ((SpawnPoint.flags & MF_AMBUSH) == 0)
+			if (not (SpawnPoint.flags & MTF_AMBUSH))
```
CurrentResident, line 1187: 😄

common/p_enemy.cpp
```diff
@@ -944,15 +944,15 @@ void P_NewChaseDir (AActor *actor)
...
+			if ((actor->flags & mask(actor->target->flags)) & MF_FRIEND &&
 			    P_IsFriendlyThing(actor, actor->target) &&
 			    co_friend_distance.asInt() << FRACBITS > dist && !P_IsOnLift(actor) &&
 			    !P_IsUnderDamage(actor))
 			{
 				deltax = -deltax, deltay = -deltay;
 			}
 			else if (actor->target->health > 0 &&
-			         (actor->flags ^ actor->target->flags) & MF_FRIEND)
+			         (actor->flags ^ combo(actor->target->flags)) & MF_FRIEND)
```
CurrentResident, line 955: Can you help me understand `combo` here?  Are we implementing an idiom from one of the C++ standards?  (if so, this is a new one for me!)

common/p_mobj.cpp
```diff
@@ -3638,7 +3633,8 @@ void P_SpawnMapThing (mapthing2_t& mthing, int position)
 		return;
 
 	// check for appropriate skill level
-	if (!(mthing.flags & G_GetCurrentSkill().spawn_filter))
+	// TODO: change type of spawn_filter to MapThingFlags after merging with type-safe mapinfo PR
+	if (!(mthing.flags & combo(MapThingFlags::unsafe_from_int(static_cast<int16_t>(G_GetCurrentSkill().spawn_filter)))))
```
CurrentResident, line 3694: This kind of pattern appears in multiple locations.  It might be really helpful to see some brief examples and HOWTO info in the flags.h header file to help folks understand how to use it and when one combination of special functions and casts and conversions are needed.  I ask because I can imagine myself down the road, getting wrapped around the axle trying to make a new piece of code work using a flagset.

CurrentResident (dismissed):

common/actorflags.h
electricbrass: > Is there any way to avoid the need to do this?

No, not without C++26.

> If so, would it be possible to add a static assert or something that explains exactly to the developer what the solution should be?

I don't know if there's a good way to add a static assert for this, since the exact same limitation prevents the class actually being aware of the individual enum variants in any reasonable way. And function parameters not being constant expressions prevents trying to put this in one the functions too.

common/actorflags.h
electricbrass: Oops, I was fixing that as I was going along and forgot to do some of them

client/src/cl_parse.cpp
electricbrass: We could add a similar function to unsafe_from_int that coverts from an OFlags of a different enum type maybe?

client/src/m_menu.h
electricbrass: Yep, have been doing that. Just forgot to here.

common/d_dehacked.cpp
electricbrass: > Is there any way to build this table without having to duplicate the enumeral's integral value here, and potentially have to keep it in sync across multiple files?

Where is the value being duplicated? There's nothing to keep in sync across files either, this is only relevant to dehacked parsing.


> I have a similar question about this and the `ClearMBF21Bits` function below. It would be nice if all the special knowledge needed for these capabilities could be gathered up into a single table or definition.

Agreed, it used to get that information from the table, but changing these to be different types made it harder to keep that in one place and I'm not super happy with this solution. I would love suggestions on a better approach.

common/farchive.h
electricbrass: This is purely syntactic sugar, it is identical to the old version. There is no implicit conversion here at all since it is a deduced context. Overload resolution always chooses the best match and so because this is a better match than the integral one it will choose this one.

common/p_mobj.cpp
electricbrass: I'll add some extra info on this later today!

common/p_enemy.cpp
electricbrass: It's an idea from the article linked in flags.h. Essentially the idea is that there are a few different conceptual categories here: a flag set, a flag combination, and a flag mask. When just using raw integers/enums, or just defining all the bit operators directly on an enum class, these are all the same type, which opens the door to certain types of bugs.

For example:
```
// pseudocode for conciseness' sake
enum class Flags
{
    Flag1,
    Flag2,
};

OFlags<Flags> flags = OFlags::all_set();
// we want to clear Flag1
// and might forget the ~
flags &= Flag1; // doesn't compile because &= only accepts a mask
flags &= ~Flag1; // operator~ returns a mask and this compiles
// but maybe we really did mean to clear everything except that flag, so 
flags &= mask(Flag1); // now it compiles

// or say we want to set it again now
flags |= Flag1; // does compile
flags |= ~Flag1; // doesn't compile because |= requires a combo or the plain enum class
// but maybe we really did mean set everything except that one, so we can do
flags |= combo(~Flag1);
```

The operators are defined so that all the most common cases will compile, while those that are likely bugs will fail to compile. However, there are some times where we need to escape these restrictions, so `combo` is a function that converts a mask or set into a combo, and `mask` converts a combo or set into a mask. These, in combination with the named functions, provide access to the full set of possible operations, to handle some of these edge cases.

In this particular case, both are instances of OFlags, the flag set type, which is the primary type for storing the flags, but `operator^` expects a set and then a combination of flags to toggle in that set, so we use `combo` to convert the set to a combo. Some of these spots would probablybe cleaner by just not trying to be clever with bit operations tbh, or for these friendliness checks especially, creating a dedicated function, but I was just focused on getting things to compile after changing the types. I think when I come back to this later today I'll see about cleaning up some of the friendliness checks with a pair of functions for the `^` and `&` cases that are used a lot in p_enemy.

common/farchive.h
electricbrass: Just to be slightly clearer, these are all completely equivalent:
```C++
FArchive& operator<< (const OUtil::Enum auto value)
// is the same as
template <OUtil::Enum E>
FArchive& operator<< (const E value)
// which is the same as
template <typename E>
requires OUtil::Enum<E>
FArchive& operator<< (const E value)
// which is the same as
template <typename E>
requires std::is_enum_v<E>
FArchive& operator<< (const E value)
```

(I mean, maybe there's some minor difference I don't know about somewhere in the standard but for anything that matters to us here they're the same)

common/actorflags.h
electricbrass: Also I should mention, it is possible to use only a function declaration with no body, and a decent number of the operations will still be available, only some of them actually rely on being able to call the function.

common/p_mobj.cpp
electricbrass: I'll also flesh out the unit tests a lot more, right now there's not a lot. That should help give an idea of how it works too.

common/p_enemy.cpp
electricbrass: Just realized that the `flags &= mask(Flag1)` case doesn't work because ADL doesn't help here, will fix that later.

Edit: Fixed

common/p_enemy.cpp
electricbrass: I agree with you that `not` is clearer than `!` in positions like this. `(!(` is very easy to overlook the `!`.

common/actorflags.h
```diff
@@ -0,0 +1,208 @@
...
+
+//
+// Misc. mobj flags
+//
+enum class mobjflag_t : uint32_t
+{
+	// --- mobj.flags ---
+
+	MF_SPECIAL       = BIT(0), // call P_SpecialThing when touched
+	MF_SOLID         = BIT(1),
```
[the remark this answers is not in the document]
electricbrass: hm, should I include doomtype.h here or is there a better place to move the definition of BIT to. Probably should also change its current definition from a macro to

```c++
template <std::integral T = uint32_t>
constexpr T BIT(int a)
{
    return static_cast<T>(1) << a;
};
```

common/p_enemy.cpp
electricbrass: Currently, I'm treating mask and combo as unnamed types that are just implementation details of all this, but I'm realizing there's a couple spots where it might make more sense to be able to directly use those types (bossaction_t in particular might make sense to hold a flag combo instead of a flag set), so I'll probably add public names for those too.

common/actorflags.h
CurrentResident: Okay, in that case, I'd like to suggest we either add some sort of intrusive hard-to-ignore thing (e.g. a `LAST_VALUE` sentinel / some other mech), or at least a highly-visible, eye-catching comment within the enum declaration that a completely new developer adding to the end can't miss.

client/src/cl_parse.cpp
CurrentResident: Sounds good to me!

common/d_dehacked.cpp
CurrentResident: > Where is the value being duplicated? There's nothing to keep in sync across files either, this is only relevant to dehacked parsing.

In actorflags.h, in the actual definitions of `mobjflag2_t` and `mobjflag3_t`.  Both the enumeral names and their values (i.e. the `BIT(0)`, `BIT(9),`, `BIT(17),` et al.) are specified / correlated in both files.  Ideally we'd only need to give the values in the enum declaration and no where else.

Let me see if an idea on how to consolidate pops in there.

common/farchive.h
CurrentResident: mmm!  I see.  Would you mind terribly adding a brief comment explaining that any enum should safely resolve to that overload?

common/d_dehacked.cpp
electricbrass: Oh I see. No these need to be separate unless we wanted to reorganized the enums (which would just move the problem to the heretic flags instead of the mbf21 flags) Yes, most are the same, but if you look, some of them are actually different. The values in the enum are internal implementation details while this is a mapping between them and the values defined by the MBF21 spec that are needed when parsing dehacked. And when we add id24 flags, those will probably also go in flags3 since it has the space, and adding a flags4 unnecessarily hurts cache performance, so that will be more flags that require mapping between the internal values and those used in the modding interface.

common/p_mobj.cpp
CurrentResident: Awesome, sounds great!

common/d_dehacked.cpp
CurrentResident: Okay, that is a key piece of info that would be *super* helpful to see documented in the code.  I had no idea.

Hmm, I know this ship has probably long since sailed, but if we are decoupling our flags from the specific values in the standard, then it does make me wonder if a `std::bitset` would've been an option.  And if so, maybe visualizing it in the debugger using the simple serially-numbered bit numbers might have been easy.

> adding a flags4 unnecessarily hurts cache performance

Another tangential, perhaps semi-philosophical thought, surely we're not doing so badly that another word would actually be that detrimental?  As standards evolve, it's basically unavoidable, and repacking the flags kind of just kicks the can down the road, right?  I'm not saying that we should necessarily avoid repacking, or rearranging fields for avoiding implicit padding, but are we actually hurting?

electricbrass: > does make me wonder if a `std::bitset` would've been an option.

Maybe 🤷

> surely we're not doing so badly that another word would actually be that detrimental?

I'm sure it wouldn't really hurt, but we did see from Blair recent optimizations that AActor is very sensitive to the size and layout of its fields. That's not actually the reason for the current setup anyways though. It's because MBF21 defined flags with identical functionality to some of the Heretic flags, which we already had in flags2, so we needed some way to collapse those 3 flags between dehacked parsing and their usage later (hence the 3 gaps in flags3, which now that I think about it, we might as well scoot the flags over to fill those one since theres not really a real purpose served by that).

> Okay, that is a key piece of info that would be _super_ helpful to see documented in the code. I had no idea.

Yea, I think it was a lot clearer in the version of the code code prior to this PR, definitely should add some comments here.

common/p_horde.cpp
```diff
@@ -316,11 +312,7 @@ class HordeState
 		if (G_IsHordeMode() && (!G_IsLivesGame() || playerslives.count > 0))
 		{
 			hordecooldown[hordecoolcount] = wavename;
-			hordecoolcount = ++hordecoolcount;
-			if (hordecoolcount >= HORDECOOLDOWN_SIZE)
-			{
-				hordecoolcount = 0;
-			}
+			hordecoolcount = std::min(hordecoolcount + 1, HORDECOOLDOWN_SIZE);
```
bcahue: If I'm not mistaken, because the array starts at 0, is there a chance hordecoolcount could be 10 and thus out of bounds in the array?

I think this will do what you wanted:

```cpp
hordecoolcount = (hordecoolcount + 1) % HORDECOOLDOWN_SIZE;
```

common/p_horde.cpp
```diff
@@ -292,11 +292,7 @@ class HordeState
 							attempts = 0;
 							break;
 						}
-						horderolodex = horderolodex--;
-						if (horderolodex < 0)
-						{
-							horderolodex = HORDECOOLDOWN_SIZE - 1;
-						}
+						horderolodex = std::max(horderolodex - 1, HORDECOOLDOWN_SIZE - 1);
```
bcahue: Same here, I believe it's going to be checking slot 9 no matter what with this code if `g_horde_cooldown` > 1. Similar fix:

```cpp
horderolodex = (horderolodex + HORDECOOLDOWN_SIZE - 1) % HORDECOOLDOWN_SIZE;
```

common/flags.h
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	[[nodiscard]]
+	constexpr underlying to_int() const noexcept
+	{
+		return m_value;
+	}
+
+	[[nodiscard]]
+	explicit operator underlying() const noexcept
```
bcahue: No `constexpr`?

common/flags.h
```diff
@@ -0,0 +1,475 @@
...
+		return static_cast<const Derived&>(*this);
+	}
+
+public:
+	[[nodiscard]]
+	static constexpr Derived all_set() noexcept
+	requires FlagEnumWithMax<E>
+	{
+		using unsigned_underlying = std::make_unsigned_t<underlying>;
+		return Derived{(static_cast<unsigned_underlying>(enable_bitflag_operators(E{})) << 1) - unsigned_underlying{1}};
```
bcahue: I believe `unsigned_underlying` will give a narrowing conversion error when a int32_t enum checks all_set with the highest bit possible set.

common/flags.h
```diff
@@ -0,0 +1,475 @@
...
+	{
+		// ugh, cast needed cause of integer promotion
+		return Derived{static_cast<underlying>(m_value ^ to_underlying(e))};
+	}
+
+	[[nodiscard]]
+	constexpr Derived toggle() const noexcept
+	{
+		// ugh, cast needed cause of integer promotion
+		return Derived{static_cast<underlying>(m_value ^ all_set())};
```
bcahue: m_value is of type `underlaying` and `all_set()` returns type `Derived`, so I believe this would error. Did you mean to do `all_set().to_int()`?

bcahue (commented):
I went and tried to find as many things as I could. This looks good to me but I have to admit a ton of it flies over my head lol

common/flags.h
electricbrass: Oops, yes this should be constexpr

common/flags.h
electricbrass: Yep, this is exactly why I need to flesh out the unit tests, because there's a bunch of spots like this that have very obvious bugs as soon as it actually gets compiled.

common/flags.h
electricbrass: Oh yes, I need to add an additional static_cast back to underlying after the subtraction.

common/p_horde.cpp
electricbrass: Interesting, I just did the clang-tidy fix for the sequence point error. I guess I should've looked more closely at what it suggested.

electricbrass:
> I have to admit a ton of it flies over my head

I did open this PR just a few hours after saying "i aspire to be as much of a template sicko as gooberman lol"

common/flags.h
electricbrass: In fact, it probably doesn't even compile as-is because I'm using {} instead of ().

bcahue:
One more possible thing, since `operator&` is a bool now, we have code that tries to bit shift via (flag & flag_to_and) >> to_shift. For example, r_things.cpp line 275:

```cpp
dcol.translation = translationref_t(translationtables + (MAXPLAYERS-1)*256 +
    ( (vis->mobjflags & MF_TRANSLATION) >> (MF_TRANSSHIFT-8) ));
```

I believe this would always be set to 0, and I think we'd be able to make the above a compile time error as well.

electricbrass:
Oh good catch. I'm not sure how to make it an error though

At least to make it work properly though we can change `(vis->mobjflags & MF_TRANSLATION)` right there to `(vis->mobjflags & mask(MF_TRANSLATION)).to_int()`

electricbrass:
> I'm not sure how to make it an error though

We could make it return a struct containing a bool, and that type could have an `explicit operator bool`. This would make it work almost all cases where we want it to, and force spots like this to be an error, BUT it would mean we can't assign the result directly to a bool. Maybe that's a tradeoff we want to make?

electricbrass:
If there's anything I want out of C++ profiles, it's a no implicit conversions profile (I assume this has to be converting to an int, right? `bool` doesn't have an operator>>, right?????? please tell me it doesn't)

Though user-defined implicit conversions would be nice to keep sometimes, this PR kinda relies on them.

electricbrass:
> no implicit conversions

And no integer promotion

electricbrass:
> We could make it return a struct containing a bool, and that type could have an `explicit operator bool`. This would make it work almost all cases where we want it to, and force spots like this to be an error, BUT it would mean we can't assign the result directly to a bool. Maybe that's a tradeoff we want to make?

I think I'm going to do this. It would be a useful utility type that could be used in other parts of the codebase to improve type-safety.

bcahue:
My idea was to make operator& return a flag_set instead of a boolean, but that'd probably require the API to change to use a function to compare if a flag is set. I'm not entirely sure if its viable or not.

electricbrass:
Strange, this latest push didn't seem to trigger the clang-tidy workflow for some reason?

common/g_mapinfo.cpp
```diff
@@ -1271,21 +1271,21 @@ struct MapInfoDataSetter<level_pwad_info_t>
...
-			{ "compat_missileclip", &MIType_CompatFlag, &ref.flags }, // original height monsters when it comes to missiles (not implemented)
-			{ "compat_dropoff", &MIType_CompatFlag, &ref.flags, LEVEL_COMPAT_DROPOFF }, // todo: not implemented
-			{ "compat_crossdropoff", &MIType_CompatFlag, &ref.flags2, LEVEL2_COMPAT_CROSSDROPOFF },
-			{ "compat_trace", &MIType_CompatFlag, &ref.flags }, // todo: not implemented
-			{ "compat_boomscroll", &MIType_CompatFlag, &ref.flags }, // todo: not implemented
-			{ "compat_sectorsounds", &MIType_CompatFlag, &ref.flags }, // todo: not implemented
-			{ "compat_nopassover", &MIType_CompatFlag, &ref.flags, LEVEL_COMPAT_NOPASSOVER },
-			{ "compat_invisibility", &MIType_CompatFlag, &ref.flags},  // todo: not implemented
+			{ "compat_shorttex", &MIType_CompatFlag, &ref.flags.data(), static_cast<uint32_t>(LEVEL_COMPAT_SHORTTEX) },
+			{ "compat_limitpain", &MIType_CompatFlag, &ref.flags.data(), static_cast<uint32_t>(LEVEL_COMPAT_LIMITPAIN) },
```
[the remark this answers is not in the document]
electricbrass: This and the other deprecated warnings are intentional

common/actorflags.h
electricbrass: Actually here's a potential issue with making it intrusive. It restricts it to only be usable with our own enum types. Probably not an issue though

electricbrass: I long for the day where this could be a C++26 annotation on the type and it figures out the rest from that automatically... (though that would also have that potential issue)

client/src/m_options.cpp
```diff
@@ -1970,20 +1936,21 @@ void M_OptDrawer()
...
 	OdamexEffect(x1,y1,x2,y2);
 
-	title = W_CachePatch (CurrentMenu->title);
-	screen->DrawPatchClean (title, MENU_CENTER_X - (title->width() / 2), MENU_TITLE_Y);
+	const patch_t* title = W_CachePatch(CurrentMenu->title);
+	screen->DrawPatchClean(title, MENU_CENTER_X - (title->width() / 2), MENU_TITLE_Y);
 
-	y = 15 + title->height();
-	ytop = y + (CurrentMenu->scrolltop * 8);
+	int y = 15 + title->height();
```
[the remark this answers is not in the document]
electricbrass: I'm gonna just leave this for later

electricbrass:
Alright, this should be good to go now. Pretty sure I addressed everything, except for the remaining clang-tidy warnings that I am intentionally leaving for now. I haven't marked these as resolved, because it turns out that if you do that, then push again, it just comments them again. So they'll just have to stay unresolved until right before merging.

common/g_mapinfo.cpp
```diff
@@ -1271,21 +1271,21 @@ struct MapInfoDataSetter<level_pwad_info_t>
...
+			{ "compat_boomscroll", &MIType_CompatFlag, &ref.flags.data() }, // todo: not implemented
+			{ "compat_sectorsounds", &MIType_CompatFlag, &ref.flags.data() }, // todo: not implemented
+			{ "compat_nopassover", &MIType_CompatFlag, &ref.flags.data(), static_cast<uint32_t>(LEVEL_COMPAT_NOPASSOVER) },
+			{ "compat_invisibility", &MIType_CompatFlag, &ref.flags.data()},  // todo: not implemented
 			{ "author", &MIType_Author, &ref },
-			{ "normalinfighting", &MIType_SCFlags, &ref.flags2, LEVEL2_NORMALINFIGHTING, ~LEVEL2_INFIGHTINGMASK },
-			{ "noinfighting", &MIType_SCFlags, &ref.flags2, LEVEL2_NOINFIGHTING, ~LEVEL2_INFIGHTINGMASK },
-			{ "totalinfighting", &MIType_SCFlags, &ref.flags2, LEVEL2_TOTALINFIGHTING, ~LEVEL2_INFIGHTINGMASK },
+			{ "normalinfighting", &MIType_SCFlags, &ref.flags2.data(), static_cast<uint32_t>(LEVEL2_NORMALINFIGHTING), ~static_cast<uint32_t>(LEVEL2_INFIGHTINGMASK) },
+			{ "noinfighting", &MIType_SCFlags, &ref.flags2.data(), static_cast<uint32_t>(LEVEL2_NOINFIGHTING), ~static_cast<uint32_t>(LEVEL2_INFIGHTINGMASK) },
```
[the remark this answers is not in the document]
electricbrass: Hm, I wonder if there's any reasonable way to like force resolve these so that it doesn't bring them back if they get resolved. Without changing the code, since I *want* the warnings for now. If there is, might be a good feature request for clang-tidy-review

bcahue (dismissed):
This looks good here, looks like everything that I wanted to be resolved is resolved. If I had anything for the wishlist, maybe a couple tests that checks out if signed/unsigned flag types can reach the top bit with `all_set()`, and something to test the behavior around `(flags & mask(MF_TRANSLATION)).to_int() >> (MF_TRANSSHIFT - 8)`. But this looks good, and a heck a leap forward towards being type safe. Great work!

electricbrass:
> and something to test the behavior around `(flags & mask(MF_TRANSLATION)).to_int() >> (MF_TRANSSHIFT - 8)`

With the change I just made to the tests, each part of this should be being tested now.

bcahue (approved):

Issue #964, opened 2024-05-24
[CHANGE] Keep spynext focus after warmup


Xenaero:
**What is the existing behavior? Please describe.**
Currently when spectating a player during warmup in duels, the camera will kick you back to your pov once the round begins.

**What is the desired behavior? Why are you requesting the change?**
If you are spectating a player in warmup, keep the focus on the same player once the game switches from warmup/countdown to the live match. It seems like it would be less jarring this way for the spectator.

**If applicable, describe alternatives you've considered**
Could just deal with it and re-F12, but it is inconvenient, and would be jarring for a broadcast.

**Additional context**
Less button presses!!!

PR #1917, opened 2026-08-23
Change to let spectators continue spying through map resets (G_DoResetLevel)

endless-r0ad:


bcahue (commented):
Does this persist in every scenario we have a reset? During `callvote reset`, `g_preroundreset 1`, `sv_countdown` countdown reset, etc

bcahue:
(If so this would fix #964)

endless-r0ad:
> Does this persist in every scenario we have a reset? During `callvote reset`, `g_preroundreset 1`, `sv_countdown` countdown reset, etc

Yes I was testing the with the attached demo for duel countdown and callvote restart. Realized this morning it was not working in a live setting - the last commit makes that work.  Still need to test for g_preroundreset 

[spytest.txt](https://github.com/user-attachments/files/31383776/spytest.txt)
(change to .odd extension)

endless-r0ad:
Confirmed it works on callvote restart, warmup countdown and g_preroundreset - here it is with that  

https://github.com/user-attachments/assets/7f53da50-ce30-4aaf-8845-7c5d4818097e

bcahue (approved):

PR #1918, opened 2026-08-23
[Feature] Add classic Skulltag springpads

rakohus:
Add support for classic Skulltag springpads. This is the only map sector thing that I think exists in classic ST (at least, in the versions we'd be looking to reproduce). Ednum 5068.


electricbrass:
Ignore the info.cpp clang-tidy warnings, protobreak cleans that up

electricbrass (approved):

PR #1922, opened 2026-08-24
Resolve mobj-to-mobj pointers following netdemo/savegame snapshot load

CurrentResident:
This PR includes two changes to resolve netdemo desyncs pertaining to mobj states.

- Resolve `target`, `goal`, and `lastenemy` mobj-to-mobj pointers upon loading a game state snapshot from a netdemo or savegame.
- Workaround a stale server tic check that can cause netdemo desyncs.

The first change involved adding a post-processing step to `DThinker::SerializeAll` after loading all thinkers.

The second change ignores the mobj's recorded server tic number from the snapshot because there are no server packets to compare against when playing back a netdemo.  This workaround is temporary because a subsequent PR is going to incorporate the tic numbers as a new netdemo-specific message.

### DIFF
--- a/common/dthinker.cpp
+++ b/common/dthinker.cpp
@@ -82,6 +82,8 @@ void DThinker::SerializeAll (FArchive &arc, bool hubLoad)
 			arc >> more;
 		}
 
+		P_ResolveMobjToMobjPointers();
+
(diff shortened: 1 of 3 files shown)


bcahue (approved):
Looks good here.

PR #1924, opened 2026-08-25
Ensure mobj "threshold" is synched during UpdateMobj

CurrentResident:
This PR adds `threshold` to the UpdateMobj message, fixing one cause of "warp walk" desyncs.

`A_Chase` uses `threshold` to determine how long a monster must remain tracking a particular target before it's allowed to choose another target or adjust its "mission" in some way.  It is typically set non-zero during `P_DamageMobj`, so that it continues chasing after a player or other monster that very recently damaged it, and that function *does not run on the client side*.  However, `A_Chase` *does run* as a part of monster movement prediction.

Therefore once `threshold` counts down to zero on the client side, it stays that way forever, which allows `A_Chase` to locally predict a premature change of movement direction if the server has a non-zero value for `threshold`.  When that happens, a subsequent UpdateMobj message corrects the misprediction, but `threshold ` may still diverge for a few seconds, allowing a repeated misprediction.  Ultimately this causes the "warp walk" desync symptom.

As for other impacts, `threshold == 0` most of the time anyway, so it's no-impact for the vast majority of cases due to protobuf's zero-value optimization.  Furthermore, the client-side logic that produces the misprediction also locally rolls `rndindex`, triggering incidental downstream mispredictions.

### DIFF
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -1205,9 +1205,10 @@ static AActor* CL_UpdateMobj(const odaproto::svc::UpdateMobj* msg, AActor* mo = 
 		mo->momz = update.mom.z;
 	}
 
-	mo->rndindex = update.rndindex;
-	mo->movedir = update.movedir;
+	mo->rndindex  = update.rndindex;
+	mo->movedir   = update.movedir;
 	mo->movecount = update.movecount;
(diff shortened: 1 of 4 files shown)


bcahue (approved):
Simple change, looks good.

PR #1926, opened 2026-08-25
Change to let the freecam use teleporters 

endless-r0ad:
This adds a slight exception to the NOCLIP rule to let the freecam use teleports and have NOCLIP. Note the call to PIT_CheckLine here is only used to populate ```spechit```


endless-r0ad:
Closing for new direction

PR #1927, opened 2026-08-26
Record packet headers with timing info into netdemos

CurrentResident:
This PR records packet headers into netdemos in sequence with all the messages it already records.  This causes the server tic and client tic-of-validity round-trip timing information to be captured and usable for netdemo replication of all rollbacks and validity checks experienced by the client during the game.

image

In the above image, we see the `msg_header` for a high-priority packet (as indicated by the `flags: 2`) with its sequence number, the server tic at the time the content was generated, and the client tic that was current / valid on the server at the time the content was generated.  The high priority content includes a `svc_servergametic` message (which is slated to be repurposed as a metrics reporting message) and a `svc_updatelocalplayer` message.

Following that, we see a second `msg_header`.  This indicates the reception of a second packet in the frame:

image

This packet indicates that it is reliable (via `reliable_size: 44`) and it is the next packet in the overall reliability-based sequence.  We see from the tic information that it was generated in the same tic as the server and its content is the result of having processed the input that was generated on client tic `214`.

If we take a look at the player attribute info/rollback messages, we see that the embedded client tic exactly matches the tic indicated by the header.

image

image

image

image

In total, not only does this allow us to eventually remove duplicated `player_tic` and `client_tic` fields from multiple messages (in a subsequent PR), but this also allows someone reviewing a netdemo to understand whether packets have gone missing, best-effort packets are received out-of-order, when retransmissions occur, and what server-to-client messages were affected by each.

### DIFF
--- a/client/src/cl_demo.cpp
+++ b/client/src/cl_demo.cpp
@@ -25,7 +25,7 @@
 
 #include "odamex.h"
 
-#include "cl_main.h"
+#include "cl_parse.h"
 #include "p_ctf.h"
 #include "d_player.h"
 #include "m_argv.h"
--- a/client/src/cl_demo.cpp
+++ b/client/src/cl_demo.cpp
@@ -37,6 +37,7 @@
 #include "st_stuff.h"
 #include "p_mobj.h"
 #include "clc_message.h"
+#include "msg_message.h"
 #include "svc_message.h"
 #include "g_gametype.h"
 #include "g_game.h"
@@ -874,6 +875,15 @@ void NetDemo::capture(const std::basic_string<byte>& buffer)
 	}
 }
 
+void NetDemo::capturePacketHeader(const PacketHeaderType& header)
+{
+	if (isRecording())
+	{
+		workingBuffer.clear();
+		MSG_WriteSVCBuffer(& workingBuffer, MSG_Header(header));
+		captured.emplace_back(workingBuffer);
+	}
+}
 
 //
 // writeLauncherSequence()
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -2255,6 +2255,7 @@ MessageResultEnum CL_AcceptNetMessage()
 	{
 		if (netdemo.isRecording())
 		{
+			netdemo.capturePacketHeader(::messenger.GetCurrentReceivedPacketHeader());
 			netdemo.capture(&::net_message);
 		}
 
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -2091,8 +2091,10 @@ bool CL_Connect()
 	else
 	{
 		PrintFmt("Requesting server state...\n");
-		messenger.NextReceivedPacket(::net_message);
-		CL_ParseCommands();
+		if (messenger.NextReceivedPacket(::net_message))
+		{
+			CL_ParseCommands(messenger.GetCurrentReceivedPacketHeader());
+		}
 	}
 
 	messenger.SendAll(gametic, ::serveraddr);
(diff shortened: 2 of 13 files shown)


client/src/cl_demo.cpp
```diff
@@ -874,6 +875,15 @@ void NetDemo::capture(const std::basic_string<byte>& buffer)
 	}
 }
 
+void NetDemo::capturePacketHeader(const PacketHeaderType& header)
+{
+	if (isRecording())
+	{
+		workingBuffer.clear();
+		MSG_WriteSVCBuffer(& workingBuffer, MSG_Header(header));
+		captured.emplace_back(workingBuffer);
```
bcahue, line 884: NetDemo::capture checks buffer size before emplace_back and we should here as well. As far as I can tell, an empty buffer still allocates memory (64 bytes * 2048 max UDP size = 128kb).

odaproto/common.proto
```diff
@@ -11,6 +11,16 @@ message Noop
 {
 }
 
+// msg_header
+message Header
+{
+	int32   sequence        = 1;
+	int32   originator_tic  = 2;
+	int32   destination_tic = 3;
+	uint32  reliable_size   = 4;
```
bcahue, line 20: reliableSize and flags in PacketHeaderType are uint16_t, so it will narrow. Kind of a rock and a hard place since protobuf doesn't have a uint16 but an explicit `static_cast<uint16_t>` in CL_Header is probably warranted because of this (and because of `-Wconversion`).

bcahue (changes requested):

client/src/cl_demo.cpp
CurrentResident: `NetDemo::capture()` is handed a variable-size buffer, which can be zero-length and would make sense to skip in that case.  In this function, we're handed a header which is a fixed-size struct, so it's always going to pack to a non-zero protobuf.

> (64 bytes * 2048 max UDP size = 128kb)

If we're talking about the `std::vector` in `buf_t`, then it's actually only 2KB, but please let me know if I'm overlooking where the 64 byte factor comes from.

odaproto/common.proto
CurrentResident: Will do.

client/src/cl_demo.cpp
CurrentResident: If you're thinking about things like memory space fragmentation, then we can look at shifting the captured buffer deque over to also using the `std::pmr` pooling mech - arguably something we should consider anyway.

bcahue: Ah I missed it was a fixed-size struct, nevermind

bcahue (approved):

client/src/cl_main.cpp
```diff
@@ -2253,10 +2255,11 @@ MessageResultEnum CL_AcceptNetMessage()
 	{
 		if (netdemo.isRecording())
 		{
+			netdemo.capturePacketHeader(::messenger.GetCurrentReceivedPacketHeader());
```
endless-r0ad, line 2258: Realize I'm late to the party here but why capture the header as a separate message from the net_message? In a live setting m_outgoingPacketBuffer has the header+message in the same buffer.

CurrentResident: The main reason is that the header is a statically-sized type and not actually encoded via protobuf, but in order to take advantage of protobuf's string dump feature, we have to get it into a protobuf message.  Because doing that encoding is an essentially different task than just dumping in a bucket o' bytes ala `net_message`, it made sense to carve that off into its own self-contained function.

In other words, the encoding of the header is protobuf in the netdemo, but with live network traffic, it's a direct aligned field-by-field packing for inspectability.

PR #1931, opened 2026-08-28
Freecam fixes and cl_noclip_spectator addition

endless-r0ad:
Addresses the following

- No longer allowed to use in a survival team game (team LMS), to prevent cheating
- Moves to the consoleplayer deathspot when being used in survival
- Added client cvar cl_noclip_spectator, toggles on/off moving through walls for spectators/freecam


endless-r0ad:
Marking as a draft because I see @bcahue has a cool implementation for DeathSpot in the works, Id like to use that instead of what Im doing here

PR #1933, opened 2026-08-29
[FEATURE] Respawn at death spot + many fixes

### LINKED ISSUE #1666: [FEATURE] Server setting for allowing players to respawn where they died in co-op
A number of mapsets are not made with co-op in mind, but are still perfectly playable if players are allowed to respawn where they die in order to avoid soft-locking the whole map.

A non-latched server setting that makes players respawn where they died instead of at player spawn point.

I guess if the ability to have cheats on wasn't latched then we could enable that on the fly and allow people to noclip into a softlocked arena. Or just enable specifically noclip instead of all cheats if the server admin wants more control. 

**Additional context**
Ideally wait for https://github.com/odamex/odamex/pull/913 to be resolved first, as respawning in the middle of an arena with pistol, 50 bullets, and 100 health won't actually accomplish much.

Additionally, this would need to check if where the player would respawn would be an instant death sector, probably if it's a damage floor as sometimes maps have inescapable damage floor pits, and that there isn't level geometry blocking the player from moving if they respawn there or a solid thing that can't be telefragged that'd also block the player from moving.

bcahue:
Fixes #1666

This is yet another PR that got away from me. If this is too much to review I will gladly split it into multiple branches as I got a bit carried away fixing the protobreak testing bugs that gobolt has graciously been finding, as well as going off on other tangents.

### Death Spot Respawn
The main attraction of this PR is the addition of the variable `g_spawnatdeathspot` which enables the user to respawn at the site of their death where P_KillMobj was called, instead of the corpse, as the corpse may have slid somewhere inaccessible or have been deleted.

To do this, I added a DeathSpotManager in order to cover the recording of players' death spots. This is needed online for source of truth bookkeeping, but also on the client to predict messages pertaining to the blocking of the death spot.

I also generalized the P_SpawnPlayer functions to enable the spawning of players no matter their x/y/z (z being beholden to LEVEL_USEPLAYERSTARTZ as always).

Also included in this functionality is the ability to _reject_ a death spot respawn and instead respawn normally. In order to do that, I created a "modifiers" ticcmd that will send a button that is normally sent as a regular ticcmd (as in, it has no functionality outside of indicating it's being pressed as a modifier). I also added some on-screen indicators to both instruct the user that they have options (like respawning, respawning in your death spot, respawning normally, etc) and also, it will change colors of shown buttons if those commands are being input currently.

Deathspot spawning has a series of rules, based on what is blocking the death spot:

- Player (coop modes)
  - If `sv_friendlyfire` is on
      - Player is telefragged
  - else
    - `sv_unblockplayers` is on
      - Spawn normally, no telefrag
    - else
      - Spawn is blocked
- Player (pvp modes)
    - If `sv_friendlyfire` is on
      - Telefrag player
  - else
    - If Player is teammate
      - `sv_unblockplayers` is on
        - Spawn normally, no telefrag
      - else
        - Spawn is blocked
    - else
      - Telefrag player
 - Monster
   - If monster is friendly and `sv_unblockfriendly` is on
     - Spawn normally, no telefrag
   - else
     - Telefrag monster
- Non-sentient being (barrel)
  - Blocked until moved or destroyed
- Avatar (voodoo)
  - Blocked until moved or destroyed
- Instant death floor
  - Blocked until changed (acs, surrounding floor specials)
- No room to stand, crusher/closed door/etc
  - Blocked until the floor or ceiling provides enough room for a player

I believe this checks are enough to handle 99% of use cases to prevent players from spawning into a condition that they cannot survive, and also have it interact with our unblock/friendlyfire cvars as you'd expect. It does blocking prediction clientside, so players are kept abreast of their ability to use their deathspot to respawn.

Test map is here: https://www.dropbox.com/scl/fi/9go4rnbeh3qoi9m2akmdr/deathspottest.zip?rlkey=7qj9oomst7tx71j4qsuohpxia&st=djx3ttwo&dl=0

MAP01 is a boom map with a crusher and instakill floor. You'll need to run a server to see how `g_spawnondeathspot` affects it.
MAP02 is a ZDoom Hexen format map, with an insta kill floor, a crusher, and a scripted switch that will temporarily make the floor in front of the switch a kill floor, then turn it off and spawn an imp. This allows you to exercise the telefrag monster behavior.

### Respawn Prompt

As mentioned above, this PR includes a prompt that comes up if the player is able to respawn. This only shows up in multiplayer, and only if the player has lives left. Here's what they look like:

Normal respawn prompt
image

Deathspot blocked - insta kill floor
image

Died to a crusher, so the prompt that says its blocked/safe to respawn alternates. Notice rshift is green, because I am pressing it.
image

A look at the prompt when spawn delay is on
image

Spawn blocked and `sv_forcerespawn` is on, indicating it will spawn normally soon.
image

Full house - blocked spawn and a `sv_forcerespawn` line telling you if you wait you'll be respawned normally in due time
image

Full house - open spawn and a `sv_forcerespawn` line telling you if you wait you'll be respawned at the death spot in due time
image

Normal spectator prompt
image

Normal freecam
image

### `cl_shotclock` and `cl_shotclocksecondsleft`

I had the idea for this when watching the respawn prompts. When enabled, under the threshold specified in `cl_shotclocksecondsleft`, any timer (including the hud timers) will turn into "shot clocks" as they're seen in professional basketball. These shot clocks, under a certain threshold, grow a tenths place under a certain time, usually 10 seconds. Whole seconds round up as they do normally, but when under the shotclock threshold, they round down to not skip any numbers (i. rounding tenths would make it go 11 (for 1 second) 10 (for 3 tics) then 9.9.

This gives a much better visualization of when 0.0 will be hit and there won't be any time left.

I also added a "monospaced" drawer for text since the hu_font is variable width, and jitters a lot when in shotclock mode.

I also added some cmdlib tests for these.

Here's a look at the round timer in shot clock mode
image

Here's what a spawn delay prompt looks like in shot clock mode
image

### Player Start and Avatar/Voodoo fixes

1. Modulo errors occurred because a player's id was being used as an index into `playerstarts[]`. This caused issues like a player with an id of 5 but the 4th player in server to join gets player 1's spawn, even though a player 4 spawn exists.
2. `EXTRA_COOP_PLAYER_STARTS` was off-by-one, which caused ednum 4061 things to become player 1 starts.
3. Created `dumpspawns` cvar to evaluate voodoo spawn results.
4. Created `P_AvatarBlocksSpot` to make a best effort at not spawning players into voodoo dolls (generally due to map errors). It will do this anyway if there are no other suitable spawns.
5. ALL of these are compatible with vanilla.

### Freecam Fixes

1. Freecam now has its x/y/z interpolated so automap displays properly.
2. Freecam's first tic has interpolation set so it doesn't have potential for a first frame jiggle.
3. Freecam was spawning into the first player thing found, which is an avatar closet on maps with them properly created. Fixed to use actual player spawns.

### Switch Fixes

When playing on the test map, I noticed non-repeatable switches in ZDoom format would flash activated then deactivated, and of course, you couldn't activate it again. I applied a fix to a bug that has likely been in since the map_format changes.

### Other Fixes
1. Rewrote a bit of `G_DeathMatchSpawnPlayer` and `G_TeamSpawnPlayer` to have the compiler not warn about potential null dereferences.
2. `P_GetPlayerStart` will throw an I_Error on the map if it can't find a `playerstart` rather than divide by 0.


common/p_setup.cpp
```diff
@@ -815,6 +874,47 @@ void P_LoadExtendedNodes(int lump, nodetype_t nodetype)
 	}
 }
 
+//
+// P_SetFreecamStart
+//
+// Puts the freecam on a real player start.
+//
+void P_SetFreecamStart()
```
endless-r0ad: @bcahue after looking at both our implementations, imo we should have this common function be something like P_GetFirstAvailableSpawn that returns a mapthing2_t. That way the #ifdef build dependency of cl_freecam can be removed from p_setup and the start position can be set clientside just before addFreecamPlayer

bcahue: Good idea, I have implemented this. I made it return a `std::optional<mapthing2_t>` as maps can be without player starts, and if that's the case it just sends back a `std::nullopt`.

common/doomstat.h
```diff
@@ -167,7 +167,7 @@ struct VoodooStartInfoType
 
 inline constexpr int MAXPLAYERSTARTS = 64;
 inline constexpr OUtil::inclusive_range<int16_t> VANILLA_COOP_PLAYER_STARTS{1, 4};
-inline constexpr OUtil::inclusive_range<int16_t> EXTRA_COOP_PLAYER_STARTS{4001, 4001 + MAXPLAYERSTARTS - 4};
```
electricbrass, line 170: I think changing to exclusive range would be more readable than changing the 4 to a 5.

server/src/sv_mobj.cpp
```diff
@@ -164,7 +172,7 @@ void P_SpawnPlayer(player_t& player, const mapthing2_t& mthing)
 		if (!player.spectator)
 		{
 			M_LogWDLEvent(WDL_EVENT_SPAWNPLAYER, &player, NULL, team, 0,
-			              M_GetPlayerSpawn(mthing.x, mthing.y), 0);
+			              M_GetPlayerSpawn(x >> FRACBITS, y >> FRACBITS), 0);
```
electricbrass: Use FIXED2INT instead

server/src/sv_mobj.cpp
```diff
@@ -61,6 +62,13 @@ void P_SetSpectatorFlags(player_t &player)
 //	between levels.
 //
 void P_SpawnPlayer(player_t& player, const mapthing2_t& mthing)
+{
+	P_SpawnPlayer(player, mthing.x << FRACBITS, mthing.y << FRACBITS,
```
electricbrass: Use INT2FIXED instead

common/g_deathspot.h
```diff
@@ -0,0 +1,214 @@
...
+#include <unordered_map>
+
+#include "m_fixed.h"
+#include "tables.h"
+
+class AActor;
+class player_t;
+struct sector_t;
+
+/// <summary>
```
electricbrass: The rest of the codebase uses the javadoc-like syntax rather than this one

common/d_ticcmd.h
```diff
@@ -96,17 +102,23 @@ struct ticcmd_t
...
 	short	pitch;			// up/down. currently just a y-sheering amount
 	short	yaw;			// left/right
 	short	forwardmove;
 	short	sidemove;
 	short	upmove;
 	byte	impulse;
+	byte	modifiers;
 };
 
+// NOLINTBEGIN(modernize-macro-to-enum,cppcoreguidelines-macro-usage) -
```
electricbrass: Why not just change these into an enum?

common/d_ticcmd.h
```diff
@@ -71,8 +72,12 @@ struct ticcmd_t
 		sidemove = 0;
 		upmove = 0;
 		impulse = 0;
+		modifiers = 0;
 	}
 
+	// NOLINTBEGIN(readability-magic-numbers) - byte offset in a
+	// hand-packed layout, like every other field above
```
electricbrass: This feels like a spot where maybe the warning should just be left for later

common/d_ticcmd.h
```diff
@@ -55,7 +55,8 @@ struct ticcmd_t
 	}
   public:
 
-	static constexpr size_t SERIALIZED_SIZE = 2 + sizeof(short) * 5;
+	// NOLINTNEXTLINE(google-runtime-int) - sizeof of the actual field type
+	static constexpr size_t SERIALIZED_SIZE = 3 + (sizeof(short) * 5);
```
electricbrass: Might as well change those fields to int16_t too, since all the surrounding code assumes that width

common/p_setup.cpp
```diff
@@ -142,6 +141,90 @@ std::vector<mapthing2_t>         DeathMatchStarts;
...
+
+	for (size_t i = 0; i < ::voodoostarts.size(); i++)
+	{
+		const VoodooStartInfoType& voodoo = ::voodoostarts[i];
+		const AActor* mo = voodoo.mobj;
+
+		PrintFmt(PRINT_HIGH, "  [{}] player {} - type {} at {},{} - avatar {}\n", i,
+		         P_GetMapThingPlayerNumber(voodoo.mapThing) + 1, voodoo.mapThing.type,
+		         voodoo.mapThing.x, voodoo.mapThing.y,
+		         mo ? fmt::sprintf("netid %d at %d,%d", mo->netid, mo->x >> FRACBITS,
```
electricbrass: Would be more readable to not mix both format string syntaxes here

electricbrass: Use FIXED2INT

server/src/sv_game.cpp
```diff
@@ -246,28 +247,116 @@ void G_PlayerReborn (player_t &p) // [Toke - todo] clean this function
...
 void P_SpawnPlayer (player_t &player, const mapthing2_t& mthing);
+void P_SpawnPlayer (player_t &player, fixed_t x, fixed_t y, fixed_t startz, angle_t angle);
 
+bool G_CheckSpot (player_t &player, fixed_t x, fixed_t y, fixed_t startz, angle_t angle);
 bool G_CheckSpot (player_t &player, const mapthing2_t& mthing)
 {
-	unsigned			an;
-	AActor* 			mo;
-	fixed_t 			xa,ya;
+	return G_CheckSpot(player, mthing.x << FRACBITS, mthing.y << FRACBITS,
```
electricbrass: Use FIXED2INT

client/src/cl_game.cpp
```diff
@@ -1343,19 +1349,28 @@ void G_PlayerReborn (player_t &p) // [Toke - todo] clean this function
 // because something is occupying it
 //
 void P_SpawnPlayer(player_t &player, const mapthing2_t& mthing);
+void P_SpawnPlayer(player_t &player, fixed_t x, fixed_t y, fixed_t startz, angle_t angle);
 
+bool G_CheckSpot(player_t &player, fixed_t x, fixed_t y, fixed_t startz, angle_t angle);
 bool G_CheckSpot(player_t &player, const mapthing2_t& mthing)
+{
+	return G_CheckSpot(player, mthing.x << FRACBITS, mthing.y << FRACBITS,
```
electricbrass: Use INT2FIXED

client/src/cl_level.cpp
```diff
@@ -709,6 +712,18 @@ void G_DoLoadLevel (int position)
 	// clientside only freecam, added after demo players are added in G_UnSnapshotLevel
 	if (Freecam::allowAdd())
 	{
+		if (Freecam::needPosition())
+		{
+			const std::optional<mapthing2_t> start = P_GetFirstAvailableSpawn();
+
+			if (start.has_value())
+			{
+				Freecam::setStartPosition(start->x << FRACBITS, start->y << FRACBITS,
```
electricbrass: Use INT2FIXED

client/src/cl_mobj.cpp
```diff
@@ -54,6 +56,13 @@ void G_PlayerReborn(player_t &player);
 //	between levels.
 //
 void P_SpawnPlayer(player_t& player, const mapthing2_t& mthing)
+{
+	P_SpawnPlayer(player, mthing.x << FRACBITS, mthing.y << FRACBITS,
```
electricbrass: Use INT2FIXED

common/g_deathspot.h
bcahue: I used XML comments here because they integrate nicely with Intellisense, which is helpful when building the feature. I changed it back to Javadoc style, but the other managers this was modeled after (announcer, spree, etc) still use XML comments

bcahue:
Review suggestions implemented, but it may fail on clang-tidy now since I got rid of a NOLINT on something it did not like.

common/g_deathspot.h
electricbrass: Ah, I have hardly looked at those files and never noticed they used it too. I'm surprised intellisense doesn't work well with both. vscode and zed both seem to support both just as well

electricbrass (approved):

PR #1934, opened 2026-08-29
Linux/BSD install fixes

electricbrass:
Currently, we only install metainfo, icons, desktop files, and mimetypes through a special script for Flatpak builds. These should be installed through a normal `cmake --install`, for users who prefer compiling themselves or for 3rd party packaging.


electricbrass:
I should mention that for now I've just put these in the else blocks after the apple and windows and everything else checks. There's potential for that to mean these get installed on platforms that they don't make sense for, but that's also not anything we officially support. There isn't really a perfect check. (LINUX OR BSD) is overly restrictive (UNIX AND NOT APPLE) is too broad. So I figured this is probably good enough until we find a case where it installs these files when it shouldn't.

bcahue (approved):

PR #1935, opened 2026-08-29
CMake: switch the common library to using PUBLIC sources and consolidate common dependencies

CurrentResident:
The prime driver for this is to let CMake replicate the common sources to the executable targets directly and resolve a dependency issue caused by being explicitly added to the executable targets.

This includes the cleanup items:

- Let `odacommon` bring its dependencies with it.
- Switch from more of the old CMake commands to CMake 3's metatarget based commands.
- Remove direct dependency on X11 libs.

### DIFF
--- a/client/CMakeLists.txt
+++ b/client/CMakeLists.txt
@@ -47,6 +47,9 @@ source_group("Client" FILES ${CLIENT_SOURCES})
 source_group("SDL" FILES ${CLIENT_SDL_SOURCES})
 source_group("GUI" FILES ${CLIENT_GUI_SOURCES})
 
+get_target_property(COMMON_SOURCES odacommon INTERFACE_SOURCES)
+source_group("Common" FILES ${COMMON_SOURCES})
+
 # Client definitions
 include_directories(${COMMON_DIR} ${CLIENT_DIR})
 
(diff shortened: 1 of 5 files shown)


client/CMakeLists.txt
```diff
@@ -150,17 +141,15 @@ if(TARGET SDL2::SDL2 OR TARGET SDL::SDL)
     target_link_libraries(${TEXTSCREEN_LIBRARY} SDL2::SDL2)
   endif()
 
-  if (NOT GCONSOLE)
```
electricbrass, line 153: Did we lose the if not GCONSOLE for textscreen?

CurrentResident: No, it's still there.  See line 122.

This particular block was to make sure that if it's there, it searches the ../common directory for headers, else it just apparently sticks ../common as an include directory for anything else built here.  It's slightly confusing because the logic conflates a couple cases that need to search ../common.

So the check wasn't even really about GCONSOLE at all - more correct would have been `if (TARGET ${TEXTSCREEN_LIBRARY})`

electricbrass: Ah, that was a bit that github had collapsed!

electricbrass (approved):
Love seeing some cleanup and modernization of the cmake

PR #1945, opened 2026-08-31
[BUGFIX] `show_messages` affects chat

### LINKED ISSUE #1921: [BUG] disabling notifications should not disable chat
Pressing F8 disables both notifications and chat messages. It should only disable notifications ("Picked up some bullets.") like it does in vanilla. There can be a situation when player has chat disabled and is not aware of it.

Similar issue fixed in woof: https://github.com/fabiangreffrath/woof/issues/2678

electricbrass:
Fixes #1921

To mute all chat messages, a new `mute_chat` cvar has been added.

I have also changed `printlevel_t` to be an `enum class`, which revealed a couple other bugs that are fixed here as well:
1. The player limit reduced message when lowering `sv_maxplayersperteam` is now sent to the correct player(s) instead of always being sent to player 2. The printlevel is now always `PRINT_HIGH` instead of using the id of the player.
2. Whatever was going on in the client `C_PrintString` where the printlevel was being set to pickup or obituary depending on line wrapping. I haven't looked into what issues this caused exactly, but it was definitely wrong.

### DIFF
--- a/client/src/c_console.cpp
+++ b/client/src/c_console.cpp
@@ -123,17 +123,18 @@ static struct NotifyText
 } NotifyStrings[NUMNOTIFIES];
 
 // Default Printlevel
-std::array<int, 9> PrintColors =
-{	CR_RED,		// Pickup
-	CR_GOLD,	// Obituaries
-	CR_GRAY,	// Messages
-	CR_GREEN,	// Chat
-	CR_GREEN,	// Team chat
+constexpr std::array PrintColors =
+{
+	CR_RED,    // Pickup
+	CR_GOLD,   // Obituaries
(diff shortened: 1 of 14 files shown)


electricbrass:
Converting draft for now, while the existing behavior is inconsistent with vanilla, it is consistent with other popular multiplayer ports, and has been existing behavior in Odamex for a long time.

gera2342:
Doesn't seem to work. I connect with two clients, with mute_chat 0 and show_messages 0. They can't see each other's messages.

Idk why this should be an option, but it's fine with me if it defaults to not muting chat. And maybe also it should print a warning each time someone has their chat disabled and types a message in the chat.

PR #1949, opened 2026-09-02
Debugger pretty printers for OFlags

electricbrass:
It looks like Microsoft's `.natvis` files aren't really sophisticated enough to be helpful here unfortunately (it sounds like they would require manually writing out all the enums), but for gdb and lldb at least, we can get much more readable info:
GDB:
image
LLDB:
image

There isn't any very good way to always autoload these, so I've just left comments in the files on the debugger commands to load them.


electricbrass:
Also, yes the filenames are inconsistent. GDB auto-loading requires the name to be like that while LLDB refuses to load a file with a - in the name.

electricbrass:
I was able to get the lldb one to load automatically in vscode with: `"cmake.debugConfig": { "initCommands": [ "command script import ${workspaceFolder}/tools/debugger/odamex_lldb.py" ] }` in my settings.json. The gdb one I had to copy it to the build directory (plus a second copy names odasrv-gdb.py) and add `add-auto-load-safe-path /home/mia/source/odamex/build/odamex/odamex-gdb.py` and `add-auto-load-safe-path /home/mia/source/odamex/build/odamex/odasrv-gdb.py` to my `~/.config/gdb/gdbinit`

electricbrass:
Figured out a vscode solution for gdb too:
```json
    "cmake.debugConfig": {
        "setupCommands": [
            {
                "text": "-enable-pretty-printing",
                "ignoreFailures": true
            },
            {
                "text": "source ${workspaceFolder}/tools/debugger/odamex-gdb.py",
                "ignoreFailures": false
            }
        ]
    }
```

bcahue (commented):
Put a new line in each script to make git happy and I will approve

bcahue (approved):

PR #1950, opened 2026-09-03
Use tic timing info embedded into packet headers for rollback and validity checks

CurrentResident:
This PR removes an armful of client and server tic fields from a variety of messages and updates their handlers to use the tic numbers associated with their respective packet headers.

Now that the packets have had tic timing information in their headers for a week or two, and established confidence that they correspond exactly to the message-specific tic fields, it is time to start relying on that information and avoid duplicating the tic numbers in multiple spots.

Additionally, this updates the UpdateMobj handler to perform validity checks using the actual tic numbers instead of the reliable sequence numbers, which while that had been adequate in most cases, there were still opportunities for inadvertent use of stale data because successive best-effort updates without an intervening reliable update would still pass the filter.  Now that the real server tic information is used, we can be sure that even the rare cases of out-of-order reception will not pass the filter.


bcahue (approved):
This looks good to me, again how it affects the wider netcode is hard to grasp, but mechanically this seems solid. Good stuff replacing the message tic fields with the header tic fields!

PR #1951, opened 2026-09-03
Feature/ipv6 support

Thunderbolt1003usa:


CurrentResident:
Yo @Thunderbolt1003usa thanks for the contribution!  I want to let you know that we're focusing our heavy-lifting netcode updates on the [protobreak branch](https://github.com/odamex/odamex/tree/protobreak).  It would be a great help if you'd be willing to either rebase your branch on it, or cut a new one based on protobreak plus your changes.

After a brief discussion with the team, it was also brought up that the odalauncher code and banlist code would need to be updated as well, and surely other areas as well.  @electricbrass any more thoughts, questions, suggestions?

Thanks again!

PR #1954, opened 2026-09-03
[BUGFIX] Fast monsters set wrong under certain circumstances when switching wads

electricbrass:
Currently, `D_UndoDehPatch` always forces `isFast` to false, even if the backup data was saved while `isFast` is true. Simply removing this would still have issues where monsters with modified speeds in a new dehacked patch might be already in their non-fast state while trying to undo the fastness. The solution is to simply always reset everything to be non-fast just before backing up data before loading a new deh patch.


bcahue (approved):
Looks good, we have nightmare demos that would fall over if this were wrong as well

PR #1957, opened 2026-09-04
[BUGFIX] Automap crosshair dot drawn over player arrow

electricbrass:
This has bugged me for years and I always forget about it. Now it's only drawn if not following the player.


electricbrass:
Also added a fix for the arrow keys not working when am_followplayer is enabled. I broke that back in 11.1, oops.

bcahue (approved):

PR #1960, opened 2026-09-06
Replace global MeansOfDeath with function parameters

electricbrass:
As far as I can tell the only reason this was a global variable was because some of the older functions didn't have a parameter where it could be passed. ...so let's add one

### DIFF
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -1696,7 +1696,7 @@ void CL_SpectatePlayer(player_t& player, bool spectate)
 	if (!player.spectator && !wasalive)
 	{
 		if (player.mo)
-			P_KillMobj(NULL, player.mo, NULL, true, MOD_NONE);
+			P_KillMobj(nullptr, player.mo, nullptr, true, MOD_NONE);
 
 		player.playerstate = PST_REBORN;
 	}
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -312,7 +312,7 @@ void M_Ticker(void);
 
 size_t P_NumPlayersInGame();
 void G_PlayerReborn (player_t &player);
-void P_KillMobj (AActor *source, AActor *target, const AActor *inflictor, bool joinkill);
+void P_KillMobj (AActor *source, AActor *target, const AActor *inflictor, bool joinkill, int mod = 0);
 void P_SetPsprite (player_t& player, int position, int32_t stnum);
 void P_ExplodeMissile (AActor* mo);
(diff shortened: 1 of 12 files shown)


common/p_interaction.cpp
```diff
@@ -1536,17 +1538,16 @@ static void ClientObituary(AActor* self, const AActor* inflictor, AActor* attack
...
 
+	bool friendly = false;
 	if (G_IsCoopGame())
-		MeansOfDeath |= MOD_FRIENDLY_FIRE;
+		friendly = true;
 
 	if (G_IsTeamGame() && attacker && attacker->player &&
 	    self->player->userinfo.team == attacker->player->userinfo.team)
-		MeansOfDeath |= MOD_FRIENDLY_FIRE;
+		friendly = true;;
```
bcahue: Double semicolon

common/p_local.h
```diff
@@ -340,38 +340,42 @@ void P_DamageMobj (AActor *target, const AActor *inflictor, AActor *source, int
...
+	MOD_SUICIDE,
+	MOD_BARREL,
+	MOD_EXIT,
+	MOD_SPLASH,
+	MOD_HIT,
+	MOD_RAILGUN,
+	MOD_FIREBALL, // Odamex-specific - monster fireball.
+	MOD_HITSCAN,  // Odamex-specific - monster hitscan.
+	MOD_VILEFIRE, // Odamex-specific - vile fire.
+	NUMMODS,
```
bcahue: NUMMODS being in the enum makes it possible to use but I imagine its temporary for when we make it an enum class.

bcahue (changes requested):

bcahue (approved):

client/src/cl_main.cpp
```diff
@@ -312,7 +312,7 @@ void M_Ticker(void);
 
 size_t P_NumPlayersInGame();
 void G_PlayerReborn (player_t &player);
-void P_KillMobj (AActor *source, AActor *target, const AActor *inflictor, bool joinkill);
+void P_KillMobj (AActor *source, AActor *target, const AActor *inflictor, bool joinkill, int mod);
```
CurrentResident, line 315: Is there a reason we use an `int` here and not the enumeration?

CurrentResident (approved):
Looks good to me - only one question about the use of an int vs enumeration.

client/src/cl_main.cpp
electricbrass: Yes it was the whole circular dependency issue with p_local.h that I decided to tackle later. So for now this is consistent with all other means of death parameters.

PR #1961, opened 2026-09-06
[BUGFIX] `spawninv` command allows clientside modification of `g_spawninv` in multiplayer

electricbrass:
Without this, players can change their spawninv, which will get applied on respawn. Since it's client-only and true inventory state is managed by the server, it can't be used for cheating but it does cause temporary inventory desyncs.


bcahue (approved):
Quick, easy and simple! Approved.

PR #1962, opened 2026-09-06
Various cleanup plus case-insensitive string types

electricbrass:
This is a bit of a hodgepodge of various code cleanup stuff. Most of it is all pretty minor, stuff like replacing a few C arrays with `std::array`, changing some `const std::string&` arguments to `std::string_view`, etc. The main big thing here is there is are new string and string view types for case insensitive comparison: `IString` and `IStringView`. These are specializations of `std::basic_string` and `std::basic_string_view`, so they act just like `std::string` and `std::string_view` except for the comparison behavior. They are case-preserving, rather than normalizing the string on assignment/construction like in `OLumpName`, and functions are provided to freely convert between them and regular strings/string_views (do note that converting *to* the string types allocates a new string while converting to the string view types results in a view pointing to the same memory).

These are utf-8 *safe* but not utf-8 *aware*. Only alphabetic ascii characters are compared case-insensitively.


electricbrass:
I don't know if I feel like cleaning up the ANIMATED loading code more than I already did. All these warnings are ones that were already there.

common/c_dispatch.cpp
```diff
@@ -317,86 +317,89 @@ void C_DoCommand(std::string_view cmd, uint32_t key)
...
-		while (cend > cstart && *cend == ' ')
-			cend--;
-
-		size_t clength = cend - cstart + 1;
-		memcpy(command.get(), cstart, clength);
-		command[clength] = '\0';
+		while (command.starts_with(' '))
+			command.remove_prefix(1);
+		while (command.ends_with(' '))
+			command.remove_prefix(1);
```
bcahue: I believe this should be remove_suffix(1). Seems to only affect AddCommandString when the string has a trailing space.

client/src/c_bind.cpp
```diff
@@ -487,7 +479,7 @@ std::vector<int> OKeyBindings::GetKeysForCommandByLastDevice(const char* cmd)
 
 	for (const auto& [key, binding] : Binds)
 	{
-		if (binding.empty() || stricmp(cmd, binding.c_str()) != 0)
+		if (binding.empty() || binding.c_str() != cmd)
```
bcahue: Address compare here, should be `binding != cmd` like above

bcahue (changes requested):
I'm getting sleepy so I will pick this review up tomorrow, but I found a few things for now

client/src/cl_game.cpp
```diff
@@ -740,24 +740,24 @@ bool G_Responder (const event_t& ev)
...
+				cmd == "sizeup" &&
+				cmd == "sizedown" &&
+				cmd == "togglemap" &&
+				cmd == "spynext" &&
+				cmd == "chase" &&
+				cmd == "+showscores" &&
+				cmd == "bumpgamma" &&
+				cmd == "screenshot" &&
+                cmd == "stepmode" &&
+                cmd == "step"))
```
bcahue, line 760: stricmp/strnicmp for this conditional is checking for cmd having more chars than the text provided, as equivalency is 0 for those functions. Now we are checking for equivalency. I think it should be inverse -- or better yet, maybe a `constexpr std::array` combined with `std::ranges::none_of`

bcahue (changes requested):
I lied, here's one more. I'd still like to give it the fine toothed comb treatment tomorrow

PR #1965, opened 2026-09-07
[BUGFIX] Add am_ovscalelock to keep the minimap square

### LINKED ISSUE #1956: [BUG] Minimap is taller when using fullscreen hud
**Describe the bug**
The minimap is taller than normal when screenblocks is 11 or 12.

5ae02a5c4607c83cc35461ea0948d3575c387a40

**To Reproduce**
Enable `am_ovminimap` and change screen size with - and +

**Expected behavior**
Be a square in both modes

Image

Image

teesquared:
Fixes #1956 

This adds the cvar and menu option for am_ovscalelock. Use it to lock the width or height of the minimap so that it stays square when switching to the fullscreen HUD.

### DIFF
--- a/client/src/am_map.cpp
+++ b/client/src/am_map.cpp
@@ -2452,6 +2452,8 @@ void AM_Drawer()
 		case 2:
 			f_h = std::min(f_w, v_height);
 			break;
+		default:
+			break;
 		}
 
 		switch (loc)
--- a/client/src/am_map.cpp
+++ b/client/src/am_map.cpp
@@ -146,6 +146,7 @@ EXTERN_CVAR(am_ovbackcolor)
 EXTERN_CVAR(am_ovbackalpha)
 EXTERN_CVAR(am_ovscalewidth)
 EXTERN_CVAR(am_ovscaleheight)
+EXTERN_CVAR(am_ovscalelock)
 EXTERN_CVAR(am_ovlocation)
 
 EXTERN_CVAR(netdebug_automap)
@@ -2435,12 +2436,23 @@ void AM_Drawer()
 		const int v_width = R_ViewWidth(surface_width, surface_height);
 		const int v_height = R_ViewHeight(surface_width, surface_height);
 		const int loc = am_ovlocation;
+		const int lock = am_ovscalelock;
 
 		int x_offset = 0;
 		int y_offset = 0;
 
 		f_w = v_width * am_ovscalewidth;
 		f_h = v_height * am_ovscaleheight;
+
+		switch (lock)
+		{
+		case 1:
+			f_w = std::min(f_h, v_width);
+			break;
+		case 2:
+			f_h = std::min(f_w, v_height);
+			break;
+		}
 
 		switch (loc)
 		{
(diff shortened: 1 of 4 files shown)


client/src/m_options.cpp
```diff
@@ -1362,7 +1363,13 @@ std::array<value_t, 6> MinimapLocations = {{
 	{ .value = 5.0, .name = "Right Bottom"},
 }};
 
-std::array<menuitem_t, 22> AutomapItems = {{
+std::array<value_t, 6> MinimapScaleLock = {{
```
bcahue: array sized for 6 but only 3 options exist, would result in menu having 3 blank options

client/src/am_map.cpp
```diff
@@ -2435,13 +2436,26 @@ void AM_Drawer()
...
 		f_h = v_height * am_ovscaleheight;
 
+		switch (lock)
+		{
+		case 1:
+			f_w = std::min(f_h, v_width);
+			break;
+		case 2:
+			f_h = std::min(f_w, v_height);
+			break;
```
bcahue, line 2454: Are these supposed to be f_w = f_h = std::min(...) /  f_h = f_w = std::min(...)? I don't see how it locks square without it.

client/src/am_map.cpp
```diff
@@ -2435,13 +2436,26 @@ void AM_Drawer()
...
+		default:
+			break;
```
bcahue, line 2456: Can the default lock a square as well (maybe keep smallest)? As it currently stands `am_ovscalelock 0` still has the bug listed in #1956

bcahue (changes requested):

client/src/m_options.cpp
teesquared: Good catch! Fixed.

client/src/am_map.cpp
teesquared: You are right. If you have the scale lock at 0 (none) the minimap will not be square. I changed the default to 2 (lock height) so by default now it forces a square minimap but left the 2 scale sliders in and the no lock option (lock none) in case folks want more size options. If you want only square than I'll go with your suggestion and also remove the second slider as we don't need width and height sliders if it's always square. The lock option can also be removed in this case unless we want to give users the ability to scale the square map by view width or height. This does simplify it a lot. Let me know.

client/src/am_map.cpp
teesquared: Answered above. I changed the default for now but we can only allow square instead.

client/src/am_map.cpp
bcahue: I am not sure of the team's opinion but a square mini map makes sense to me, and would collapse 2 sliders and 1 chooser into one "scale" slider as you mentioned. I like that idea a lot since simplifies the scaling.

electricbrass: I do find it pretty neat to be able to do something like have the minimap be the full screen height and a quarter or so of the width, but overall I think the simplicity of just having it always square is probably better.

Issue #1393, opened 2025-10-11
[BUG] Team Colors Aren't Used on Corpses Upon Joining a Game in Progress


Acts19quiz:
**Describe the bug**
Upon joining a team game in progress (e.g., CTF), the corpses of the other players are default green rather than their team colors.

**Build that the bug occurred in**
11.1.1

**To Reproduce**
1. Join a team game (e.g., CTF) in progress.
2. Observes the corpses of players that died in-game before you joined.

**Expected behavior**
The team colours to always be correct.

**Screenshots, NetDemos, & Crash Dumps**
CTF game here (WDL2025D.WAD):
[Odamex_CTF_20251011_182354_WDL2025D.WAD_MAP10.zip](https://github.com/user-attachments/files/22867635/Odamex_CTF_20251011_182354_WDL2025D.WAD_MAP10.zip)
Screenshot 1:
<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/d8dc5903-f5cd-41b8-bbc4-c7d703e37127" />
Screenshot 2 (after all the corpses that were present when I joined despawned):
<img width="1600" height="900" alt="Image" src="https://github.com/user-attachments/assets/150603a2-b6f8-4224-901e-b830da211a7c" />

**Additional context**
This may also happen in team modes besides CTF; it might also happen for skin colors in general. Further testing would be required to see how deep the issue goes (if it's all skin color situations, or just team colors; Team DM/LMS instead of just CTF; etc.).

This is also not a mere netdemo bug; this happens in live play.

PR #1966, opened 2026-09-08
[BUGFIX] Corpse Color

bcahue:
This is a PR that fixes the corpse color issue described in #1393. Corpses never had player information attached to them, so in order for you to see their true colors, you had to see them die for them to get the correct color. Players who leave will also cause their corpses to revert back to green for everyone to see.

That has been fixed -- late joiners get to see all corpse colors, and for players within a game, corpses take the player's colors from the player they came from on death and stay there. Corpses all get their own cached translation that isn't tied to the players translation table -- a given recipe in a std::map will return a translationtable. This is also compatible with `r_forceenemycolor` and `r_forceteamcolor` -- only caveat being that new joiners in DM won't be able to tell what corpses are their own as the playerid is recycled after disconnect.

There's no actor walking -- if `r_forceenemycolor` or `r_forceteamcolor` are flipped, the recipes for those corpses are simply rebuilt and available to be used on the same corpses on the next display tic.

V_RGBtoHSV is now in `common` so the server can calculate the player color "bump" that is applied to team colors and apply it to corpses.


client/src/v_text.h
```diff
@@ -53,6 +53,10 @@ struct OGlobalFont
 	{
 		return m_lineHeight;
 	}
+	[[nodiscard]] const lumpHandle_t* data() const
```
CurrentResident, line 56: Not specifically this line, but in general:

> This is where I could've stopped, but since I made the cached translation table, I figured I'd make it modular enough to support the next things you'd might want to do with a translation table container -- add a lifetime to it, and it'd be simple to create translations to handle sampling and remapping of fonts and blood colors too.

The direction you're taking it in sounds pretty good, but my vote is that this should have been two PRs, even if they had to be stacked on top of each other with this one as a prerequisite.  Speaking as a relatively new reviewer on this project, it makes my life easier, with smaller, more narrowly-focused PRs tending to get merged in quickly due to being easier to digest.

I'm not necessarily asking for any specific change here, just a consideration going forward for future PRs.

common/actor.h
```diff
@@ -454,6 +454,12 @@ class AActor : public DThinker
 
 	team_t friend_teamid = TEAM_NONE; // team of the player who spawned this actor
 
+	uint32_t corpse_color = 0;
```
CurrentResident, line 457: Another general comment / question:

> This is also compatible with r_forceenemycolor and r_forceteamcolor -- only caveat being that new joiners in DM won't be able to tell what corpses are their own as the playerid is recycled after disconnect.

Can you help me understand what exactly this means?  Will late-joining players see existing corpses for previous players as their own colors rather than that player's color, since technically as fresh joiners, they haven't yet left any corpses laying around.  Or in what way could they be confused?

Also, I'm totally up for not recycling player IDs.  Once upon a time, mobj netids were apparently recycled, and I can barely imagine the complexities needed to make all the infrastructure around non-unique IDs actually comfortably lay flat around that wrinkle.  If reworking player ID management helps this (and potentially other features) work right, I'm onboard!

CurrentResident (dismissed):

common/actor.h
rakohus: I think he means if they disconnect and reconnect mid-game

client/src/v_text.h
bcahue: In the interest of making this more reviewable, I'll split off the font translation stuff into another branch.

common/actor.h
bcahue: Sorry, I should've been more clear with this scenario. r_forceenemycolor will change the enemy color and their corpse color to what you specify, but when you die, your corpse color is not translated to the r_forceenemycolor color. When you disconnect and reconnect, all corpses will be that enemy color, because the link of what corpse belonged to you is lost.

We don't do any playerid association of corpses in this, it simply records which corpse is yours, and playerids do not move with corpses now, only colors, teams, and whether or not the corpse belonged to you, which is derived instead of transmitted.

bcahue: Okay I just tested this, and it looks like on reconnect and r_forceenemycolor 1, you DO retain your isself property on your corpses, since it doesn't look like the corpse queue is cleared on disconnect. So it does work for the simple disconnect/reconnect case.

bcahue: Nevermind, no corpse queue (only for horde monsters), its just that the playerid is not being removed on disconnect, so if you reconnect and get the same playerid , CL_GetPlayerColor which runs after setting the corpse colors (to apply r_forceenemycolor) does its own playerid check.

PR #1967, opened 2026-09-08
[BUGFIX] Better CONCHARS handling

bcahue:
A bug exposed by [specdest.wad](https://www.doomworld.com/idgames/levels/doom2/Ports/megawads/specdesc), this WAD when loaded had no glyphs in the console. Turns out was because of a CONCHARS lump that was 256x32 instead of 128x128 like Odamex's own CONCHARS lump.

Now we allocate a surface the same size of the CONCHARS lump, and check to see if glyphs can be used based on their relative size. If they can, the glyph is added to the console font -- now patched one glyph at a time since a PWADs CONCHARs lump usually doesn't have Odamex-specific glyphs, so instead of taking a blank one and using it, we check each glyph for emptiness, and if so, simply use what was there before.

The fallback is to always use Odamex's CONCHARS lump if there's an issue with all loaded CONCHARS lumps from loaded PWADs. This ensures a console can't be blank with no glyphs which is an unfortunate failure mode.

### DIFF
--- a/client/src/c_console.cpp
+++ b/client/src/c_console.cpp
@@ -934,7 +934,10 @@ EXTERN_CVAR(con_scrlock)
 
 static constexpr int CONCHARS_GLYPH_DIM = 8;
 static constexpr int CONCHARS_COUNT = 256;
-static constexpr int CONCHARS_GLYPH_BYTES = CONCHARS_GLYPH_DIM * CONCHARS_GLYPH_DIM * 2;
+
+static constexpr int CONCHARS_ROW_BYTES = 2 * CONCHARS_GLYPH_DIM;
+static constexpr int CONCHARS_GLYPH_BYTES = CONCHARS_GLYPH_DIM * CONCHARS_ROW_BYTES;
+static constexpr size_t CONCHARS_BYTES = static_cast<size_t>(CONCHARS_COUNT) * CONCHARS_GLYPH_BYTES;
 static constexpr palindex_t CONCHARS_TRANSCOLOR = 0xF7;
 
 //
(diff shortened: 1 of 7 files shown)


client/src/c_console.cpp
```diff
@@ -1059,7 +1070,7 @@ void C_InitConCharsFont()
 void C_ShutdownConCharsFont()
 {
 	delete [] ConChars;
```
electricbrass: Could probably change `ConChars` to be a `std::unique_ptr<byte[]>` and get rid of this function

client/src/c_console.cpp
```diff
@@ -1038,9 +1049,9 @@ void C_InitConCharsFont()
 		}
 
 		const int filenum = W_GetLumpFile(lumpnum);
-		const char* filename = filenum >= 0 && filenum < static_cast<int>(wadfiles.size())
-		                           ? wadfiles[filenum].getBasename().c_str()
-		                           : "an unknown file";
+		const char* filename = "an unknown file";
+		if (filenum >= 0 && std::cmp_less(filenum, wadfiles.size()))
```
electricbrass: I foresee wanting this same functionality in the future for other error messages as well (I was actually just thinking about it the other day for parsing ANIMATED lumps). Should we make a function that just returns a string_view of the file name? Also the casting here has me wondering if there's a good reason for W_GetLumpFile to return `int` instead of `size_t`

electricbrass: Oh right, -1. Still might suggest `std::optional<size_t>` if it were changed. Would also prevent anyone from trying `wadfiles[W_GetLumpFile(lump)]` and not accounting for the possibility of `-1`.

bcahue: Good idea, implemented.

client/src/c_console.cpp
bcahue: I changed this to an std::vector<byte> instead because it was triggering a `modernize-avoid-c-arrays` clang-tidy error with std::unique_ptr<byte[]>. We still get RAII/easy lifetime management with vector, so I was still able to get rid of `C_ShutdownConCharsFont()`.

electricbrass: Nice, I wasnt't sure if a vector would be a harder change vs unique_ptr.

electricbrass (approved):

PR #1968, opened 2026-09-08
[FEATURE] All the slopes!

### LINKED ISSUE #1952: [FEATURE] Vertex slopes
We have line slopes

We should also add support for vertex slopes

**Additional context**
Would make for more interesting terrain and structures, it would also support maps created by tools such as [this](https://www.doomworld.com/forum/topic/118563-terrain-generator-dd_terrain-for-gzdoomk8vavoomrad-updated-jan-3-2022/)

electricbrass:
Resolves #1952

This adds support for all of the remaining ways of creating slopes, other than those that require UDMF (which should be super easy to build on top of this once we have support for it). 

I don't feel like describing them here, just go read the ZDoom wiki page: https://zdoom.org/wiki/Slope
It's everything except:
- Plane_Align (we already had this)
- UDMF plane equations
- UDMF vertex heights
- Eternity Engine's Doom format specials for Plane_Copy (these are only supported by UZDoom for it's half-working Eternity compat mode and I don't expect we'll be running EE maps anytime soon)
- Eternity Engine's Doom format specials for Plane_Align (we have the ZDoom version of those, which are completely identical except for the special number, also same reasoning as Plane_Copy)

I did not add them to the UDB config because they are already there, inherited from the ZDoom config

### DIFF
--- a/common/p_setup.cpp
+++ b/common/p_setup.cpp
@@ -73,7 +73,7 @@ void P_SpawnCompatibleSectorSpecial(sector_t* sector);
 namespace {
 void P_SetupLevelFloorPlane(sector_t *sector);
 void P_SetupLevelCeilingPlane(sector_t *sector);
-void P_SetupSlopes();
+void P_SetupLineSlopes();
 }
 
 void P_InvertPlane(plane_t *plane);
@@ -1628,7 +1628,7 @@ void P_SetupPlane(sector_t* sec, line_t* line, bool floor)
 	srcplane->texy = refvert->y;
 }
 
-void P_SetupSlopes()
+void P_SetupLineSlopes()
 {
 	for (line_t& line : R_GetLines())
 	{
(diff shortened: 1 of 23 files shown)


drivingspectrum:
Awesome work! Here is what it looks like with the volcanos.terrain example wad I generated with dd_terrain, the texturing doesn't work but that is probably an entire different pr by itself
image
[volcanos_terrain.zip](https://github.com/user-attachments/files/32005148/volcanos_terrain.zip)

common/p_setup.cpp
```diff
@@ -1658,6 +1658,411 @@ void P_SetupSlopes()
...
+
+		v3double_t p;
+		p.x = FIXED2DOUBLE(line.v1->x);
+		p.y = FIXED2DOUBLE(line.v1->y);
+		p.z = P_PlaneZ(FIXED2DOUBLE(line.v1->x), FIXED2DOUBLE(line.v1->y), &plane);
+
+		v3double_t v1;
+		v1.x = FIXED2DOUBLE(line.dx);
+		v1.y = FIXED2DOUBLE(line.dy);
+		v1.y = P_PlaneZ(FIXED2DOUBLE(line.v2->x), FIXED2DOUBLE(line.v2->y), &plane) - p.y;
```
bcahue: is this supposed be z?

common/p_setup.cpp
```diff
@@ -1658,6 +1658,411 @@ void P_SetupSlopes()
...
+
+		v3double_t v2;
+		v2.x = FIXED2DOUBLE(x - line.v1->x);
+		v2.y = FIXED2DOUBLE(y - line.v1->y);
+		v2.y = FIXED2DOUBLE(z) - p.y;
```
bcahue: same here regarding z

common/p_setup.cpp
```diff
@@ -1658,6 +1658,411 @@ void P_SetupSlopes()
...
+
+		v3double_t cross;
+		M_CrossProductVec3(&cross, &v1, &v2);
+		M_NormalizeVec3(&cross, &cross);
+
+		// Fix backward normals
+		if ((cross.y < 0 and floor) or (cross.y > 0 and not floor))
```
bcahue: same here regarding z

common/p_setup.cpp
```diff
@@ -1890,6 +2293,8 @@ void P_SetupLevel (const char *lumpname, int position)
 	P_ResolveSkyPickers();
 	P_ResolveStackLinks();
 
+	P_CopySlopes();
```
bcahue, line 2365: Just curious, is this zdoom's cadence when it comes to slope copies? Thing (vertex) -> Line -> Copy

bcahue (changes requested):

common/p_setup.cpp
electricbrass: Yes and no
Yes its the same as zdoom. But its line slopes -> slope things -> vertex things -> copy things -> line copy

https://zdoom.org/wiki/Slope#Finally

common/p_setup.cpp
electricbrass: Yep, the zdoom 1.23 code I was referencing used indexing for vectors and I made quite a few of these mistakes when converting it over to odamexs vectors

common/p_setup.cpp
electricbrass: (So if somehow i do actually have it doing Plane_Align after things then thats a bug)

bcahue: Oops read it wrong, I blame github expanding code display. I didn't realize this was already documented, nice.

bcahue (approved):

CurrentResident (approved):
Love to see it!

PR #1971, opened 2026-09-10
Large messages

CurrentResident:
This PR adds a new message queue to the OdaMessenger:  `LargeMessage`, which allows for individual messages up to 64 KB.  Messages added to this queue are broken up into 1 KB fragments (or smaller, based on budgeting and sequencing) that are sent in reliable packets, at a slightly lower priority than natively reliable messages.

It also adds a utility for reassembling the fragments into the complete message.  Currently only the client instantiates this utility, only the server sends large messages, and the server will reject any client that sends any of the lower-level proto messages pertaining to fragmented large messages.

As of this PR, only the `MaplistUpdate` proto message is sent via `LargeMessage`, but there are several other messages that could be good candidates for this as well - potentially `MidPrint`, `Print`, or any other message that has arbitrary-length strings.

Thanks to @electricbrass for testing with exceedingly large maplists!

### DIFF
--- a/client/src/cl_main.cpp
+++ b/client/src/cl_main.cpp
@@ -469,7 +469,7 @@ static void CL_HandleDisconnectCompletionPacket()
 	// confirmation.
 	while (::net_message.BytesLeftToRead() > 0)
 	{
-		const ParseResultType result = CL_ParseCommand();
+		const ParseResultType result = CL_ParseCommand(::net_message);
 
 		switch (result.cmd)
 		{
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -222,12 +222,21 @@ void CL_LargeMessageFragment(const odaproto::LargeMessageFragment* msg)
 	s_receivedLargeMessage.Append(msg->payload().data(), msg->payload().length());
 }
 
+void CL_ParseBuffer(buf_t& buffer);
+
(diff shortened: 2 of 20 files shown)


endless-r0ad:
Theres an increasing number of PascalCase files in the repo... why are there two different naming conventions ?

CurrentResident:
It's a consequence of me trying to steer headers and source files away from monolithic grab-bags to being finer-grained with clearer dependencies, plus naming the files after the classes they implement.  So between those two things, we get filenames that TBH are inconsistent with the convention typically used in doom source code.

I can be talked into renaming the files, but I do think being finer-grained ultimately leads to better dependency management.

endless-r0ad:
> It's a consequence of me trying to steer headers and source files away from monolithic grab-bags to being finer-grained with clearer dependencies, plus naming the files after the classes they implement. So between those two things, we get filenames that TBH are inconsistent with the convention typically used in doom source code.
> 
> I can be talked into renaming the files, but I do think being finer-grained ultimately leads to better dependency management.

Makes sense, Id argue you can keep finer-grained meaning while using snake_case. Just my 2c, I dont really care other than it is kind of annoying to see different conventions. I realize the industry standard is name the file the same as the class

drivingspectrum (approved):
I also tested this with a crazy large maplist, it works fine on my end and cannot see anything glaring out at me

PR #1980, opened 2026-09-12
[FEATURE] Add a predicted waggle

bcahue:
I started this years ago but did not finish because registering the servers as `movingsectors` caused the sector to broadcast every tic, flooding the connection with just waggle moving. Because waggle is deterministic, we are able to predict waggle on the client and keep it in lockstep indefinitely.

They are registered with unlagged, but not sector snapshots. Mid-game joiners get a message describing an already-moving waggle mover, and uses `originatorTic` to "catch up" the waggle to the tic currently being predicted by the player. This way, the server and client always see the same waggle state.

The bandwidth for these waggles are low. Attached is a test wad that used to tip over Odamex because of the sheer amount of waggle things transmitting plane position every tic. Now, the map is comparable to Doom 2 MAP01 in bandwidth in.

Also included is a local sector history manager, only used by waggle for now. This is essentially a snapshot manager but for prediction-based local sector plane heights instead of server sent heights. We can use this when we want to predict sector movement based on dead reckoning things based on their start tic -- I tried to generalize this for all specials that move planes and not just waggle, when we want to move towards that.

Also is included is a small fix from the vertex slopes branch, Hexen format maps were spawning things twice, now they only spawn once.

Here's the test WAD and demo:

[waggle8.zip](https://github.com/user-attachments/files/32140059/waggle8.zip)
[waggledemo.zip](https://github.com/user-attachments/files/32140062/waggledemo.zip)

### DIFF
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -3159,6 +3159,27 @@ void CL_ThinkerUpdate(const odaproto::svc::ThinkerUpdate* msg)
 			new DGlow2(&::sectors[secnum], start, end, tics, oneShot);
 		break;
 	}
+	case odaproto::svc::ThinkerUpdate::kWaggle: {
+		const odaproto::svc::ThinkerUpdate_Waggle& wmsg = msg->waggle();
+		const int secnum = wmsg.sector();
+
+		if (::numsectors <= 0)
+			break;
+
+		sector_t* sector = &::sectors[secnum];
+		const bool ceiling = wmsg.ceiling();
+
+		// A waggle owns its plane for as long as it runs, so don't stack a
+		// second one on a sector that already has a mover on that plane.
+		if (ceiling ? sector->ceilingdata != nullptr : sector->floordata != nullptr)
+			break;
+
+		if (secnum < ::numsectors)
+			new DWaggle(sector, ceiling, wmsg.original_height(), wmsg.accumulator(),
+			           wmsg.acc_delta(), wmsg.target_scale(), wmsg.scale(),
+			           wmsg.scale_delta(), wmsg.ticker(), wmsg.state());
+		break;
+	}
 	case odaproto::svc::ThinkerUpdate::kPhased: {
 		short secnum = msg->phased().sector();
 		int base = msg->phased().base_level();
--- a/client/src/cl_parse.cpp
+++ b/client/src/cl_parse.cpp
@@ -3166,18 +3166,36 @@ void CL_ThinkerUpdate(const odaproto::svc::ThinkerUpdate* msg)
 		if (::numsectors <= 0)
 			break;
 
-		sector_t* sector = &::sectors[secnum];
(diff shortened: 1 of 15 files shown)


client/src/cl_parse.cpp
```diff
@@ -3173,6 +3173,45 @@ void CL_ThinkerUpdate(const odaproto::svc::ThinkerUpdate* msg)
...
+			const int messageTic = ThisMessageServerTic();
+
+			if (::last_svgametic > 0 and messageTic > 0)
+			{
+				const int64_t staleTics =
+				    static_cast<int64_t>(::last_svgametic) - messageTic;
+
+				// Don't even bother to catch up the waggle if we're about to get timed out
+				if (staleTics > 0 and
+				    staleTics <= OdaMessenger::DEFAULT_CRITICAL_SEQUENCE_TIMEOUT_IN_TICS)
```
CurrentResident, line 3207: I am a little concerned about this check for a couple reasons:

- It's possible under really nasty network situations to wind up with a negative staleTics here.  In that case, we should still be able to put the sector into a correct position and animation state.
- I strongly recommend we not worry about critical sequence timeout here, because if we find ourselves in this function, it means that we're catching up and should still do what we're supposed to do with the data we have in hand, until the messenger declares that the gig is up.  In other words, timing out at that level is the messenger's job and we shouldn't second-guess it.

common/p_spec.h
```diff
@@ -1141,36 +1147,63 @@ class DWaggle : public DMover
...
+
+	fixed_t m_OriginalHeight = 0;
+	fixed_t m_Accumulator    = 0;
+	fixed_t m_AccDelta       = 0;
+	fixed_t m_TargetScale    = 0;
+	fixed_t m_Scale          = 0;
+	fixed_t m_ScaleDelta     = 0;
+	int     m_Ticker         = 0;
+	int     m_State          = init;
+	bool    m_Ceiling        = false;
```
CurrentResident, line 1182: This is probably something to consider for the future sector messaging and infrastructure refactor, but if we also record the server tic at time of creation and calculate the expected server tic at the time of end, we should have all the data needed to completely timebox the life of the waggler and the sector's height at any point.  The idea being that we should be able to just feed it `world_index` and if that value falls into the timebox, it sets the correct position without any state updates needed.

common/p_lnspec.cpp
```diff
@@ -60,6 +60,10 @@ bool EV_DoZDoomCeiling(DCeiling::ECeiling type, line_t* line, byte tag, fixed_t
 // Returns true if the special for line will cause a DMovingFloor or
 // DMovingCeiling object to be created.
 //
+// Floor_Waggle and Ceiling_Waggle are intentionally absent -- their movement is
+// never sent to clients, so clients have to run the special themselves even
+// when cl_predictsectors is off.
+//
```
CurrentResident, line 66: Great explanation, thanks!

PR #1985, opened 2026-09-13
Friday Night Fragfest features and fixes

electricbrass:
This PR is all of the changes currently being used on the Friday Night Fragfest server.

1. `addmaprange` command for more easily creating large maplists
2. `rand` command for obtaining a random value for use in server scripts
3. `sv_curpwad` is set earlier, fixing a bug where it would have a stale value when executing sv_startwadscript
4. `unalias` and `unset` commands for deleting aliases and variables
5. Ordered comparison for numeric cvars in `if` command with `gt`, `lt`, `geq`, and `leq` operators
6. When `sv_warmup_autostart` is 0, warmup countdown is not started if there are zero player, instead of starting and resetting the countdown every tic
7. `randcaps` chooses 3 team captains if `sv_teamsinplay` is 3 instead of always choosing 2
8. It is no longer possible for the last 2 odd players out to end up on the same team when running `randpickup` with `sv_teamsinplay 3`
9. CVars are unlatched and sent to clients *before* sending the message to load a new map. This fixes an issue where clients would start a map with the server's old latched values. This fixes a desync where fast monsters would be enabled clientside but not the server.
10. `sv_shuffleteams` cvar that randomizes the teams of all players on map start, and assigns players to the team with fewest players when they join the game


common/c_dispatch.cpp
```diff
@@ -532,19 +540,146 @@ BEGIN_COMMAND (if)
 
 	std::string op = argv[2];
 
-	if(op == "eq")
+	// TODO: cvar-overhaul branch - add a `compare` virtual method to cvars
+	// that takes a string and use that here instead
+
+	// TODO: allow boolean cvars  to be checked directly without any comparison
+	if (var->m_Flags & CVARTYPE_BOOL)
```
bcahue: Does this work? I figured m_Type handled this and not m_Flags.

common/c_dispatch.cpp
```diff
@@ -532,19 +540,146 @@ BEGIN_COMMAND (if)
...
+		}
+		else
+		{
+			PrintFmt(PRINT_HIGH, "if: no operator {}\n", argv[2]);
+			PrintFmt(PRINT_HIGH, "if: integer operators are eq, ne, gt, lt, geq, leq\n");
+			return;
+		}
 	}
-	else
+	else if (var->m_Flags & CVARTYPE_FLOAT)
```
bcahue: Does this work? I figured m_Type handled this and not m_Flags.

server/src/c_dispatch.cpp
```diff
@@ -0,0 +1,121 @@
...
+			max = *maxopt;
+		}
+
+		if (max < min)
+		{
+			PrintFmt("Maximum must be greater than minimum");
+			return;
+		}
+
+    	const auto range = static_cast<uint32_t>(max - min);
```
bcahue: This has a chance to pass 0 into M_RandomInt and divide by 0 (well, modulo by 0)

server/src/sv_pickup.cpp
```diff
@@ -158,7 +156,55 @@ BEGIN_COMMAND (randpickup) {
...
+nonstd::expected<void, std::string> Pickup_DistributeAllPlayers() {
+	// This function shouldn't do anything unless you're in a teamgame.
+	if (!G_IsTeamGame()) {
+		return nonstd::make_unexpected("Server is not in a team game.");
+	}
+
+	// Track all eligible players.
+	std::vector<std::reference_wrapper<player_t>> eligible;
+	for (auto& player : players) {
+		if (validplayer(player)) {
```
bcahue: `validplayer` only gates against players who are null, so spectators are counted here.

server/src/c_dispatch.cpp
bcahue: Also, should be static_cast<uint32_t>(max) - static_cast<uint32_t>(min) or the calculation could overflow before it is cast to `uint32_t` since they're both `int32_t`.

server/src/sv_pickup.cpp
```diff
@@ -158,7 +156,55 @@ BEGIN_COMMAND (randpickup) {
...
+		}
+	}
+
+	if (eligible.empty()) {
+		return nonstd::make_unexpected("No eligible players for distribution.");
+	}
+
+	const int teamCount = sv_teamsinplay.asInt();
+	std::vector<team_t> team_order;
+	team_order.reserve(teamCount);
```
bcahue: This doesn't seem to take into consideration `sv_maxplayersperteam`.

server/src/sv_main.cpp
```diff
@@ -3340,20 +3340,54 @@ void SV_JoinPlayer(player_t& player, bool silent)
...
-			if (newteam == TEAM_NONE)
-				return;
+			// assign player to the team with the fewest players
+			// choose randomly if multiple teams are tied for minimum
+			size_t min = std::numeric_limits<size_t>::max();
+			team_t minteam = TEAM_NONE;
+			int mincount = 0;
+			for (int i = 0; i < sv_teamsinplay.asInt(); i++)
+			{
+				const size_t numplayers = P_NumPlayersOnTeam(static_cast<team_t>(i));
```
bcahue, line 3353: This doesn't seem to take into consideration `sv_maxplayersperteam`.

bcahue (changes requested):
Some small requests for changes.

common/c_dispatch.cpp
electricbrass: > Does this work?

Nope!

server/src/sv_pickup.cpp
electricbrass: Correct. It shuffles the team of *all* players. I did it this way in case we wanted to split the two parts of sv_shuffleteams into separate settings.

server/src/sv_pickup.cpp
electricbrass: The old version didn't either, so I didn't want to change that without knowing if that was intentional

server/src/c_dispatch.cpp
electricbrass: > This has a chance to pass 0 into M_RandomInt and divide by 0 (well, modulo by 0)

Oops, meant to have a <= in the check just above that

server/src/sv_pickup.cpp
electricbrass: I guess if no one leaves or joins though it can result in unbalanced teams when there are spectators. But if the settings were split and it didn't shuffle the spectators, that wouldn't work out either. I guess I'll keep it all one setting but make this part skip spectators.

server/src/sv_main.cpp
electricbrass: Should I maybe just remove the else and have that part always happen?

electricbrass: Nope, that doesn't quite work, and we actually know that if the chosen player has max players, all other teams do too, since it was the one with the fewest, so no need to call SV_GoodTeam or anything, so its a very simple additional check.

server/src/sv_pickup.cpp
electricbrass: And for Pickup_DistributeAllPlayers, it doesn't need that, since it's not adding or removing players from the total amount that are currently ingame.

electricbrass: It probably should be added to Pickup_DistributePlayers though since it brings in readied spectators.

electricbrass: SV_SetPlayerSpec will call SV_JoinPlayer, which handles that, but because of when it happens, it can result in unbalanced teams, so it definitely should be added.

server/src/sv_pickup.cpp
```diff
@@ -77,22 +88,24 @@ bool Pickup_DistributePlayers(size_t num_players, std::string &error) {
...
 	// Rip through our eligible vector, forcing players in the vector
 	// onto alternating teams.
-	team_t dest_team = TEAM_BLUE;
-	size_t i = 0;
-	int teamCount = sv_teamsinplay.asInt();
-	for (std::vector<player_t*>::iterator it = eligible.begin();it != eligible.end();++it,++i) {
-		player_t &player = **it;
+	for (size_t i = 0; i < eligible.size(); i++) {
+		player_t& player = *eligible[i];
+		const team_t dest_team = team_order[i % team_count];
```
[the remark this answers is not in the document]
electricbrass: Interesting, I wonder what change allowed it to do this, since this isn't any more possible now than it was before. (Which, it isn't possible, team_count has no way to be anything but 2 or 3, but it can't really reasonably know this)

bcahue (approved):

PR #1991, opened 2026-09-15
Fix client thinker order

CurrentResident:
This PR fixes a variety of assorted mobj desyncs (some warp-walkers and some "just barely" collision mispredictions) by ensuring that for mobjs that have netids, the client-side thinker order matches the server-side order, which very conveniently correlates to netid ordering.

Before:

https://github.com/user-attachments/assets/c883ddf4-44b6-4735-b3ff-a2d561cfe5d6

After:

https://github.com/user-attachments/assets/bce316d7-f3b1-4a52-8609-782334145f4e


drivingspectrum:
Nice work, I barely notice any warp-walkers now after these changes, especially on larger maps, though I'll let another team member review this one.

common/p_mobj.cpp
```diff
@@ -744,6 +744,34 @@ AActor* P_FindThingById(uint32_t id)
...
+			{
+				if (mobj->SpliceAsThinkerBefore(iter->second))
+				{
+					break;
+				}
+			}
+		}
+		else
+		{
+			for (--iter; iter != actor_by_netid.begin(); --iter)
```
bcahue, line 764: This doesn't seem to ever check `actor_by_netid.begin()`, is that intentional?

bcahue: Also, if there's no forward scan fallback for the backwards scan failing, does this mean netids can be out of order?

common/dthinker.cpp
```diff
@@ -132,6 +128,63 @@ void DThinker::Destroy ()
...
 		m_Prev->m_Next = m_Next;
 
+	// Please note that we do NOT set m_Next or m_Prev to nullptr because in the event that
+	// during a single RunThink(), multiple nodes could be Destroyed / Unlinked, we want to
+	// be able to advance through those harmlessly in RunThinkers by just following m_Next.
+}
+
+bool DThinker::SpliceBefore(DThinker* io_node)
+{
+	if (io_node and io_node != this and not io_node->WasDestroyed())
```
bcahue: Do we need to check if this->WasDestroyed() as well?

common/dthinker.cpp
```diff
@@ -132,6 +128,63 @@ void DThinker::Destroy ()
...
+		}
+		io_node->m_Prev = this;
+		return true;
+	}
+	return false;
+}
+
+bool DThinker::SpliceAfter(DThinker* io_node)
+{
+	if (io_node and io_node != this and not io_node->WasDestroyed())
```
bcahue: Do we need to check if this->WasDestroyed() as well?

bcahue (commented):
I don't have a good grasp on the overall netcode changes as a whole but I had a few things to point out

common/p_mobj.cpp
CurrentResident: > This doesn't seem to ever check actor_by_netid.begin(), is that intentional?

Yes.  In this case we're looking for a thinker to splice it after, and if we somehow get to `begin()`, then it means we've checked everything from its current position to the beginning, and didn't find a valid node to place it after, therefore it's fine where it is.  Whatever's at `begin` must precede it, with only invalid/destroyed things or nothing at all in between.

The only time it doesn't work out that way is when it, itself somehow has the lowest netid, thus already being at `begin()` and we must probe forward, which is covered by the earlier `if` block and ties to the answer of the second question:

> Also, if there's no forward scan fallback for the backwards scan failing, does this mean netids can be out of order?

No, because `actor_by_netid` is a `std::map` keyed on netid, so it will always be ordered correctly.

CurrentResident: > Whatever's at begin must precede it, with only invalid/destroyed things or nothing at all in between.

And the reason that's okay is that when we add a mobj to the thinker list, it always goes at the end, whatever its netid is supposed to be...  All other thinkers are before it in the actual thinker list, so if we don't find any lower-id'd mobjs, other than maybe the very first one, it's actually fine for us to just leave it at the end.

common/dthinker.cpp
CurrentResident: That's a reasonable precaution.  Will do.

common/dthinker.cpp
CurrentResident: In practice we can know that we don't need it because we only get into this function when an existing mobj without a netid is given one, but like the other case, it sure would be a reasonable safety check.  Fixing.

bcahue (approved):

PR #1992, opened 2026-09-16
[FEATURE] Voxel Renderer

rakohus:
**Overview**

This PR has been in the works for the better part of 2026. It adds software voxel rendering to Odamex. The renderer is based on work originally done by Andrew Apted, found at https://gitlab.com/andwj/voxel-rend, with additional integration and behavior specific to Odamex. Fortunately, most of the hard work was already done but there was some balancing to be done between how the other two "voxel ports", ZDoom and Woof, handle things. 

Test wad here: [odavoxel_v1.zip](https://github.com/user-attachments/files/32270846/odavoxel_v1.zip)

Voxels can be loaded from WADs _or_ from an external voxels directory. Voxels contained in server-loaded WADs always take priority over files in the client’s external directory.

https://github.com/user-attachments/assets/6ad2c9ff-7842-4355-ae10-45902ca3926d

**New CVARs**
`r_voxels`: Enables software voxel rendering. Defaults to 1.
`r_voxeldir`: Sets the directory used to load external .kvx files. Defaults to voxels.
`sv_allowvoxels`: Allows clients to load custom voxel models from their external directory. Defaults to 1.

Setting sv_allowvoxels to 0 only disables client-provided voxels from r_voxeldir. It does not disable voxels contained in WADs loaded by the server. This is a safe guard to prevent clients from replacing important models, such as player sprites, with custom voxels that could provide an advantage in competitive play.

https://github.com/user-attachments/assets/75b7668a-13a8-4165-8ee0-488b1bf5c85f

**VOXELDEF**

Woof doesn't support `VOXELDEF`, while ZDoom does. We chose to follow ZDoom’s format where possible so existing voxel packs require fewer Odamex-specific changes. 

The following properties are currently supported:

- `AngleOffset`
- `XOffset`
- `YOffset`
- `ZOffset`
- `Spin`
- `PlacedSpin`
- `DroppedSpin`
- `UseActorPitch`

The following properties are _recognized_, but currently ignored with a non-fatal console warning:

- `Scale`
- `PitchFromMomentum`
- `UseActorRoll`
- `OverridePalette`

These were left out because they either 1) need additional renderer work; or 2) do not currently provide enough benefit to justify complicating the initial implementation. 

Odamex also adds a _new_ property: `FaceViewerPitch`

This allows forward-facing voxels, such as spheres and other powerups, to pitch toward the viewer while limiting the maximum angle. This keeps them readable from above or below without allowing extreme rotations (think spheres, they look horrible if they don't tilt toward the player that is above or below). We intend to propose this property to UZDoom so both ports can remain compatible.

When no `VOXELDEF` behavior is provided, Odamex applies sensible defaults intended to make common voxel replacements look good out of the box. Placed weapon pickups spin, while dropped weapons face the player so they remain easy to identify. Powerups and other special items also face the player, with their rotation limited to avoid extreme angles at wide FOVs. Monsters, projectiles, and other actors retain behavior appropriate to their movement and orientation.

https://github.com/user-attachments/assets/0d1605e3-d4a7-4ab3-bee0-4c086abbc22d


drivingspectrum:
Awesome work Ralph! Though unfortunately, I did come across an issue with the imp voxel clipping through a wall
image

rakohus:
> Awesome work Ralph! Though unfortunately, I did come across an issue with the imp voxel clipping through a wall

Yeah, unfortunately this is just how Doom's sprite clipping treats this sort of thing. I think this is largely the same in the other ports that support voxels and I'm not sure how I could fix it without implementing some sort of voxel-aware depth comparison against wall drawsegs. Going beyond the sprite based clipping is way beyond my pay grade for now

bcahue (approved):
Looks good to me here, appreciate the additional guards.

electricbrass (changes requested):
It looks like right now the default voxel directory is a subdirectory of the CWD? We should have it in a proper location that won't change depending on where you run it from.

client/src/r_voxel.h
```diff
@@ -0,0 +1,45 @@
...
+
+struct r_voxelvis_s
+{
+	const void* model{};
+	angle_t angle{};
+	fixed_t TL_x{};
+	fixed_t TL_y{};
+	fixed_t c{};
+	fixed_t s{};
+	fixed_t pitchSlope{};
```
CurrentResident, line 38: Huh interesting.  I never connected the dots that that syntax would do zero-value initialization.

Looks like we use `type fieldName = 0;` syntax with the `=` operator for the structs defined in the .cpp file.  Consider having them use the same syntax for consistency.  Speaking only for myself, I have a slight preference for the curly-brace syntax, but either is fine really.

client/src/r_voxel.h
```diff
@@ -0,0 +1,45 @@
...
+//-----------------------------------------------------------------------------
+
+#pragma once
+
+#include "tables.h"
+
+class AActor;
+struct vissprite_t;
+
+struct r_voxelvis_s
```
CurrentResident, line 30: Consider using the `_t` suffix for this type.

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
+// Emacs style mode select   -*- C++ -*-
+//-----------------------------------------------------------------------------
+//
+// $Id$
+//
+// Copyright (C) 2026 by The Odamex Team.
+//
+// Software voxel rendering based on the original work by Andrew Apted.
+//
```
CurrentResident, line 9: I haven't read the details of what the GPL fully requires, but we should consider reproducing Andrew's copyright line as well.

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
...
+	F_FRONT = 0x08,
+	F_TOP = 0x10,
+	F_BOTTOM = 0x20,
+};
+
+std::unordered_map<uint64_t, VoxelModel> g_voxels;
+std::unordered_map<uint64_t, VoxelRenderOptions> g_voxelOptions;
+std::deque<r_voxelvis_s> g_visibleVoxels;
+fixed_t g_eye_x = 0;
+fixed_t g_eye_y = 0;
```
CurrentResident, line 121: I definitely appreciate the use of `g_` (or `s_` ) for global / statically-scoped objects!

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
...
+constexpr double kMaxViewerPitchDegrees = 89.0;
+constexpr double kAngleUnitsPerCircle = 4294967296.0;
+constexpr fixed_t VX_MINZ = 1 * FRACUNIT;
+constexpr fixed_t VX_MAX_DIST = 2048 * FRACUNIT;
+constexpr fixed_t VX_NEAR_RADIUS = 512 * FRACUNIT;
+constexpr fixed_t VX_Z_OFFSET = -3 * FRACUNIT;
+constexpr angle_t VX_MAX_VIEWER_YAW = ANG45;
+constexpr fixed_t VX_MAX_PITCH_SLOPE = FRACUNIT / 3;
+constexpr double VX_MIN_PITCH_DZ = static_cast<double>(FRACUNIT) / 8.0;
+constexpr double VX_NEAR_VERTICAL_HORIZ = static_cast<double>(FRACUNIT) / 4.0;
```
CurrentResident, line 76: These consts are a mixture of camelCaseNames and CAPS_SNAKE_CASE_NAMES.  Let's consider having a consistent convention.

Also, do we know what the `k` prefix is for?  Is it shorthand for "this is a const literal?"

client/src/r_voxel.h
```diff
@@ -0,0 +1,45 @@
...
+{
+	const void* model{};
+	angle_t angle{};
+	fixed_t TL_x{};
+	fixed_t TL_y{};
+	fixed_t c{};
+	fixed_t s{};
```
CurrentResident, line 37: Consider avoiding single-letter or aggressively-abbreviated names, especially for interface-level types.

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
...
+
+	const byte* p = bytes;
+	p += 4; // num_bytes
+
+	out.x_size = static_cast<int>(p[0]);
+	p += 4;
+	out.y_size = static_cast<int>(p[0]);
+	p += 4;
+	out.z_size = static_cast<int>(p[0]);
+	p += 4;
```
CurrentResident, line 645: This assumes that the data endianness matches the native endian.  I'm guessing that the voxel format assumes little-endian.  If so, let's use our `LELONG(static_cast<int>(p[0]));` macros so that it can work anywhere.

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
...
+		}
+		else
+		{
+			VX_RecursiveDraw(spr, x, y, w, h / 2);
+			y += h / 2;
+			h = (h + 1) / 2;
+		}
+	}
+
+	goto loop;
```
CurrentResident, line 1154: 😱 

`VX_SpritesInNode` has a `for(;;)`.  Let's consider doing that or a `while(1)` instead of the goto.

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
...
+	}
+}
+} // namespace
+
+void VX_Init()
+{
+	g_voxels.clear();
+	g_voxelOptions.clear();
+	g_visibleVoxels.clear();
+
```
CurrentResident, line 1198: If called during runtime, seems like this would let us drop the existing voxel models and potentially do an on-demand load of an alternative set.

If so, it might be nice to trigger this if we had `r_voxeldir` as a `CVAR_FUNC_DECL` where if the directory actually changes, we do a live reload.

CurrentResident (changes requested):

CurrentResident:
BTW, I am super-stoked about this PR  :)

client/src/r_voxel.cpp
```diff
@@ -0,0 +1,1373 @@
...
+}
+
+void VX_SpritesInNode(unsigned int bspnum)
+{
+	for (;;)
+	{
+		if (bspnum & NF_SUBSECTOR)
+		{
+			const subsector_t* sub = &subsectors[bspnum & ~NF_SUBSECTOR];
+			R_AddSprites(sub->sector, sub->sector->lightlevel, FAKED_Center);
```
CurrentResident, line 1177: This is probably a very naive question...

I'm trying to understand the purpose of `VX_NearbySprites`, and I come across this `R_AddSprites` call.  I'm a little perplexed because I can see that it's also called in `R_Subsector`, which happens in the course of plain old `R_RenderBSPNode`.  The thing is, in the main render function, that happens very shortly before the call to `VX_NearbySprites`.  There might just be an understood, larger, and stateful picture that `R_AddSprites` affects that necessitates doing it twice, but I just don't grok yet.

Can you help me understand this?  Or at least, could we add a brief description / summary / usage type info to document the purpose of the interface-level functions, hopefully explaining why / when we make the calls that we do?

client/src/r_voxel.h
CurrentResident, line 30: Actually, I don't see any references to this type outside of r_voxel.cpp.  We should be fine to relocate this type there and just keep the forward declaration in r_defs.h because it's only treated as a pointer to an opaque type that nothing else touches.

client/src/r_voxel.h
```diff
@@ -0,0 +1,45 @@
...
+	fixed_t pitchSlope{};
+};
+
+void VX_Init();
+void VX_ClearVoxels();
+void VX_NearbySprites();
+bool VX_ProjectVoxel(const AActor* thing, int frame, vissprite_t* vis);
+void VX_DrawVoxel(vissprite_t* spr);
```
CurrentResident, line 45: I referenced it in another comment, but figured I'd leave it here given that it's the affected location:  It would be nice if these functions had description / usage information.

client/src/r_voxel.cpp
electricbrass: Yes, that's what the k is. I would get rid of it since we don't use it elsewhere.

PR #1995, opened 2026-09-17
Defer processing messages "from the future"

CurrentResident:
This PR addresses a variety of miscellaneous minor desyncs by postponing processing messages that are "from the future" - i.e. from a server tic that is newer than the one currently being simulated on the client.

Latency jitter is a common cause of this, and common symptom is a monster that appears to take two steps forward, a step backwards, then another two steps forward.  There are other symptoms as well - all coming down to a sensation of non-uniform motion or reactions to the player.  Please note that it does not affect the player's sensation of their own interpolated positions in the game, but it does affect the reaction that monsters and other gameplay elements have to the player's position.

### Merge order

Please note that this PR also rolls up the [fixed client-side thinker order PR](https://github.com/odamex/odamex/pull/1991) and the [large messages feature PR](https://github.com/odamex/odamex/pull/1971).  ***Thus those PRs should be merged before this one.***


client/src/cl_parse.cpp
```diff
@@ -3645,6 +3655,22 @@ const Protos& CL_GetTicProtos()
...
+	//  1.  Packets can technically arrive well out-of-order, and
+	//      we want ordered handling.
+	//  2.  c++11 and up guarantee that ordering between
+	//      values that share a given key is the order of iteration.
+	//
+	// ParseResultType contains a unique_ptr to a parsed message.
+	//
+	// The key is the server tic that the message originated on.
+	// It is expected that the first value for a given tic is the header.
+	std::multimap<int32_t, ParseResultType> s_deferredMessages;
```
bcahue: I don't see a place where this is cleared. I can see it causing issues if it's not cleared on, say, disconnect (and they reconnect without exiting), map load and maybe even netdemo seeking.

drivingspectrum (approved):

PR #2001, opened 2026-09-24
Fix netdemo playback of music where a music-changer changed the music

CurrentResident:
This fixes a bug that came up a couple times in the past during playtesting, where the wrong music begins playing after loading a netdemo snapshot representing a point in time after a level's musinfo actually changed the music during live play.  The expected result is the changed tune begins to play, but the level's default music plays instead.

This is because we simply never had calls to `P_SerializeMusInfo` for the netdemo snapshots.  The de-serialization must occur after `G_InitNew`, because that function ultimately indirectly resets the musinfo object via `S_ClearMusInfo`.

This was tested using the music-change in NT2F map32 upon entering the slaughter arena.


electricbrass:
Is there a similar issue with music changed by ACS SetMusic?

CurrentResident:
> Is there a similar issue with music changed by ACS SetMusic?

Sure looks like there's an issue here.  It looks like all the ACS-triggered music gets funneled through `DLevelScript::ChangeMusic`, which directly goes to `S_ChangeMusic`.

Since all we're ultimately serializing is just a single string, seems like we should forego SerializeMusInfo (again, just a string of what was previously started), and just serialize the lower-level current music name.  I'll give that a whirl.

electricbrass:
Sounds good to me. Should be good for the id24 music changers then too when we add support for those

CurrentResident:
Okay, I have it working at the s_sound level now, but found a wrinkle that was probably there from the get-go that I just overlooked.  Depending on conditions, it could cause the song being played to restart when jumping snapshots.  Looking into it.

CurrentResident:
Okay, fixed again!

